PR #11, opened 2014-12-28
should be == here

PeterDaveHello:


okias:
in formatting terms yes, otherwise it doesn't matter. More likely, it should be in all cases only "=", but as travis use bash, it doesn't matter.

Thank you anyway :)

Issue #8, opened 2014-07-05
Error of ping compilation for musl 


luckboy:
I tried compile iputils-s20140519 and compiler displayed error:
ing.c: In function ‘send_probe’:
ping.c:697:10: warning: initialization makes integer from pointer without a cast [enabled by default]
ping.c:697:10: warning: (near initialization for ‘m.__pad1’) [enabled by default]
ping.c:697:10: error: initializer element is not computable at load time
ping.c:697:10: error: (near initialization for ‘m.__pad1’)

Problem is at the msghdr structure. I used musl-1.0.3.
I created my patch and compiled iputils but my patch propably work for only musl.

PR #12, opened 2015-04-26
ping: don't mess with internals of struct msghdr

doughdemon:
Fixes issue #8 (compilation with musl libc)

Please test


okias:
merged. Thank you.

PR #13, opened 2015-04-30
More intuitive status output

2mac:
Been bothering me for weeks.

When you send a SIGQUIT to `ping`, it gives a quick status update. However, the output is not labeled. When I see the fraction `x/y packets`, I immediately think that I've sent x packets and received y. However, during bad latency issues or bad input, one sees that this is backwards.

I fixed it so that `x/y packets` is the exact packet success ratio.


okias:
Thank you for fix. Merged :)

PR #16, opened 2015-05-12
Transparent support of IPv6 in ping

nmav:
That patch set merges the two tools, keeping them as separate as possible. The approach is hackish but seems to do the trick.


nmav:
That's a quite hard to understand tool due to several global variables in use. I've tried to limit them, but there are way too many. I'd appreciate a review in case something was missed.

pavlix:
So far I read through the changes and all of them look good. I have some more ideas but let's consider them when the basic patchset is merged.

pavlix:
The changes look good to me.

okias:
Could you please rebase a top of latest changes? (now we're in parity with latest YH-iputils commits).

nmav:
Not really. The new changes introduce yet another global variable.

nmav:
I'll check whether I can make a new pull request.

nmav:
created #19

PR #17, opened 2015-05-13
Allow using nettle instead of libgcrypt for MD5

nmav:
nettle is simpler library, and doesn't require any explicit global initialization as libgcrypt does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_install:
   - sudo apt-get update -qq
 install:
   - sudo apt-get install libcap-dev -qq
+  - sudo apt-get install nettle-dev -qq
   # gcc
   - if [ "$CC" = "gcc-4.8" ]; then sudo apt-get install -qq gcc-4.8; fi
   # clang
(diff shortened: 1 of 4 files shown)


okias:
There is no doubt that nettle is simpler, but MD5 performance is little bit worse and gnutls or openssl is installed on almost every linux distrubution.

So - is there reason, to prefer nettle?

Also, when there will be more than two options (openssl, gnutls and nettle), could you make Makefile look little bit prettier?

nmav:
The currently available options in the makefile are libgcrypt (which is not related to gnutls) and openssl.
Nettle is the backend of gnutls since quite some years now. In all systems there is gnutls there is nettle as well. About the makefile prettyness, I'm not sure how to do that, the syntax of ifeq is the cause of this ugliness.

nmav:
It may even make sense to drop the libgcrypt option completely.

okias:
Thank you, merged with s/ping6/ipv6 ping/ :)

PR #19, opened 2015-05-29
Transparent support of IPv6 in ping (new attempt)

nmav:
That patch set merges the two tools, keeping them as separate as possible. The approach is hackish but seems to do the trick.


nmav:
This reverts the last two patches and merges them in the new internal API. Please review.

okias:
Thank you for new pull request. Maybe I didn't describe it well. I though you adapt your patch to actual master.

In worst case scenarion I'd use this pull request, but I'd prefer to keep ping rootless patches as they are without need of reverting and re-applying.

David

nmav:
I understand but I don't think I can help with that. As I mentioned in the previous pull request a rebase on the current head requires rewriting the patch set I originally sent and that's not something I currently plan to do.

okias:
I did forced push cherry-picked patches without reverting (so history is slightly modified, because I removed previously pushed rootless patches).

Thanks for your work! :)

PR #24, opened 2015-06-18
Install

krichter722:


okias:
I guess INSTALL should be rewritten anyway to something these days more meaningful :)

Thanks for pull request, but this needs more work to look sane :P

PR #25, opened 2015-06-18
Dualstack improvements

pavlix:
Please review. There will probably still be issues especially in the ping/ping6 option merge but most of the patches should be just ready to go!


pavlix:
Interesting, I do not see the same error with local builds.

nmav:
I've made some inline comments, but overall this patch set looks fine.

pavlix:
I'm going to provide a couple of other changes and clean up the patchset.

okias:
great! Thank you.

pavlix:
I refactored the patchset thoroughly into a moderate number of focused patches. I have tested it and fixed a couple of issues. One big improvement is that in some kernel configurations it is now possible to use ping without `CAP_NET_RAW`. Please review.

pavlix:
Please comment on the latest commit that implements falling back to raw socket when using ping sockets is not allowed (via sysctl).

pavlix:
To learn more about what's happening, use for example `sudo strace -e socket ./ping 127.0.0.1`.

pavlix:
Updated the branch and tested that raw socket actually works. Sorry for the inconvenience, previously I forgot to set the socktype properly but now we run the whole `create_socket()` function to ensure that the socket is created the very same way as if the raw socket was requested.

nmav:
Except for the error messages which are always output when raw sockets are in use, it looks good.

pavlix:
Good. I squashed the latest commit, fixed NL in the error message, and made the ping socket warning visible only in verbose mode. Please review and merge if ok.

nmav:
Looks like a very good refactoring to me.

okias:
thank you! merged!

PR #37, opened 2015-11-19
ping: always use POSIX locale when parsing -i

jsynacek:
Interpreting floating-point numbers from the command line based on the
current locale is a bad idea.

Original bugreport: https://bugzilla.redhat.com/show_bug.cgi?id=1283277


okias:
Agree, thank you!

srakitnican:
Thank you for a quick fix!

PR #41, opened 2016-01-26
ping: ICMP error replies while errno < 0 is a hard error

zx2c4:
There are some odd conditions in which a device will return an error for
sendto, while at the same time an ICMP error response is generated. In
this case, with the current code, the packet is retransmitted in a
flood, which is not what anybody wants. In fact, when these two
conditions occur, we want to treat the packet as a "hard local error"
and do the ordinary pause as before. The reasoning is that if we did
receive an ICMP error message, then the packet was transmitted, and so
it should be accounted for. But, since we also received an errno, this
should be reported to the user, as something is wonky with their network
interface.

Signed-off-by: Jason A. Donenfeld <Jason@zx2c4.com>

### DIFF
--- a/ping_common.c
+++ b/ping_common.c
@@ -401,9 +401,14 @@ resend:
 		return MININTERVAL;
 	} else {
 		if ((i=fset->receive_error_msg(sock)) > 0) {
-			/* An ICMP error arrived. */
-			tokens += interval;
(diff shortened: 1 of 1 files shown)


okias:
Sorry for delay! Merged!

PR #42, opened 2016-02-15
ping: link against libm

martinbark:
ping fails to compile with uClibc based toolchains with the error

  undefined reference to `__finite'

This commit links ping against libm to resolve the math library function
finite().

Note, this error is not seem with glibc based toolchains because finite seems to be resolved in both libm and libc.


okias:
Perfect, thank you!

martinbark:
Thanks.

BTW forgot to say this issue was found using [buildroot](https://buildroot.org) which includes an iputils package and will build with glibc, uClibc and musl toolchains.

PR #45, opened 2016-03-07
'admin prohibited' should print !X not !S.

morrowc:
The linux (at least) manpage says:
After the trip time, some additional annotation can be printed: !H, !N, or !P (host, network or protocol unreachable), !S (source route failed), !F (fragmentation needed), !X (communication administratively prohibited), !V (host precedence violation), !C (precedence cutoff in effect), or !<num> (ICMP unreachable code <num>). If almost all the probes result in some kind of unreachable, traceroute will give up and exit.

your code says: "ICMP6_DST_UNREACH_ADMIN" -> !S
you really mean !X


okias:
Thank you!

PR #47, opened 2016-03-23
IPv4 fallback if /proc/net/if_inet6 doesn't exist

adrian-the-git:
Without this, ping fails to run on my ipv6-less system unless I explicitly pass the -4 flag.


pavlix:
Have you tested recently? I don't think we need to check for `/proc/net/if_inet6`.

adrian-the-git:
You're absolutely right. I must've been on autopilot.

PR #49, opened 2016-04-20
ping: do not allow oversized packets to root

jsynacek:
The code later fails anyways (as the original comment suggests), which
results in a weird recvmsg() loop that times out eventually.

Reproducer:

./ping -c 1 -s 65530 127.0.0.1


okias:
I always get error, when using reproducer.. but yeah, fix seems to make sense.

PR #50, opened 2016-05-16
libgcrypt: fix static linking

martinbark:
If iputils is built statically using libgrypt as follows

 make USE_NETTLE=no USE_CRYPTO=no USE_GCRYPT=yes CFLAGS=-static

then the code fails to build with errors such as

 libgcrypt.a(libgcrypt_la-misc.o): In function `_gcry_fatal_error':
 (.text+0x22a): undefined reference to `gpg_strerror'

This is because libgrypt depends on libgpg-error.  Solve this by adding
the missing dependency to the Makefile.

Signed-off-by: Martin Bark <martin@barkynet.com>


okias:
Thank you!

PR #52, opened 2016-05-17
ping: Fix GCC warnings when building with -Wstrict-aliasing

hwoarang:
Fix the following warnings when compiling with -Wstrict-aliasing

ping.c: In function ‘ping4_run’:

ping.c:745:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
*(__u32*)&rspace[4+i*8] = route[i];
^
ping.c:765:4: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
*(__u32*)&rspace[4+i*4] = route[i];
^
ping.c: In function ‘ping4_parse_reply’:

ping.c:1025:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
ttl = *(int *) CMSG_DATA(cmsg);
^

Signed-off-by: Markos Chandras <mchandras@suse.de>

### DIFF
--- a/ping.c
+++ b/ping.c
@@ -499,6 +499,7 @@ int ping4_run(int argc, char **argv, struct addrinfo *ai, socket_st *sock)
 	char *target;
 	char hnamebuf[NI_MAXHOST];
 	char rspace[3 + 4 * NROUTES + 1];	/* record route space */
+	__u32 *tmp_rspace;
 
 	if (argc > 1) {
 		if (options & F_RROUTE)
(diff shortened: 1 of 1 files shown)


okias:
Thank you, have you also looked at Wstrict-aliasing=2 ? What do you think?

hwoarang:
It could be useful I suppose. However this "patch" is just a workaround to be honest as explained here http://stackoverflow.com/questions/21214875/gcc-accuracy-of-strict-aliasing-warnings. I assume that future gcc versions could trigger the warning. Another possible solution would be to modify the Makefile to always append -fno-strict-aliasing using the override directive https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC66. Perhaps something like

diff --git a/Makefile b/Makefile
index e60ed80..58a5f52 100644
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,9 @@ ENABLE_RDISC_SERVER=no
 # What a pity, all new gccs are buggy and -Werror does not work. Sigh.
 # CFLAGS+=-fno-strict-aliasing -Wstrict-prototypes -Wall -Werror -g
 CFLAGS?=-O3 -g
-CFLAGS+=-fno-strict-aliasing -Wstrict-prototypes -Wall
+CFLAGS+=-Wstrict-prototypes -Wall
+# Always append -fno-strict-aliasing
+override CFLAGS+=-fno-strict-aliasing
 CPPFLAGS+=-D_GNU_SOURCE
 LDLIB=

I am happy to submit such a patch if that's preferred.

okias:
well, I'd rather see code which works just fine with strict-aliasing :) If you wanted to dive into it, it will be appreciated :D

Anyway with small indentation changes, I included patch.

hwoarang:
@okias thanks. But like I said, this merely silences the current warnings. It'is not a proper fix ;)

PR #53, opened 2016-05-18
tftpd: Drop supplementary groups for root

hwoarang:
Before dropping our root privileges, we need to make sure that root
does not belong to any other group. That's because setgid() will change
the gid but it will leave the supplementary groups unchanged so we may
still be able to do privilege operations. Use setgroups() before set{u,g}id
to ensure that root does not have any unexpected priviledges.

Link: https://www.securecoding.cert.org/confluence/display/c/POS36-C.+Observe+correct+revocation+order+while+relinquishing+privileges
Signed-off-by: Markos Chandras <mchandras@suse.de>

### DIFF
--- a/tftpd.c
+++ b/tftpd.c
@@ -54,6 +54,7 @@
 #include <ctype.h>
 #include <string.h>
 #include <stdlib.h>
+#include <grp.h>
 
 #include "tftp.h"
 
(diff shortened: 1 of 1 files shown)


pavlix:
LGTM

pavlix:
Merged manually.

PR #58, opened 2016-08-04
correctly initialize first hop

### LINKED ISSUE #57: ping6 changed exit code
```
# ./ping6 2000::1 -c 1
PING 2000::1(2000::1) 56 data bytes
ping: sendmsg: No route to host

--- 2000::1 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms

# echo $?
1
```
This used to be 2. Also, this is also inconsistent with ping4 now, which still correctly returns 2.

jsynacek:
When the sin6_family was set to 0, the initial probe connect()
could have succeeded instead of failing.

Fixes #57.

### DIFF
--- a/ping6_common.c
+++ b/ping6_common.c
@@ -805,6 +805,7 @@ int ping6_run(int argc, char **argv, struct addrinfo *ai, struct socket_st *sock
 			}
 			disable_capability_raw();
 		}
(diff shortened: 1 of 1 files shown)


okias:
nice work! Thank you!

PR #62, opened 2016-08-29
arping: make usage output consistent with man page

jsynacek:


jsynacek:
This patch is actually not correct. The -I option has to be used only when the device isn't decided automatically. The error message given by the program is unfortunate, though. I'll post a better patch.

okias:
Should I close it?

jsynacek:
I'll follow up with a patch and close it then.

jsynacek:
v2: https://github.com/iputils/iputils/pull/64

PR #63, opened 2016-09-09
fix ping -6 -I

jsynacek:
ping "passes" the device variable to ping6_common.c, but it's not visible
there.

Original patch by Xin Long <lucien.xin@gmail.com>.


pavlix:
This is a trivial change that fixes up the merger of IPv4 and IPv6 pings.

PR #64, opened 2016-09-09
arping,doc: fix documentation of -I

jsynacek:
Also, give a less misleading error when a device cannot be automatically
determined and has to be specified by the user using the -I option.


okias:
Thank you!

PR #65, opened 2016-09-12
ping: fix error message when getting EACCES from connect()

jsynacek:
According to `man connect(2)`, EACCES can be set either when pinging a
broadcast address, or when a local firewall rule is in place.

Easily reproduced by running `ip route add prohibit <ipaddr>` and then
trying to ping the address.


okias:
Makes sense, thank you!

PR #67, opened 2016-11-09
Add strict pattern matching on response when pattern was provided

locasity:
A mismatch in timeval format (for eg. between the traceroute icmp packet and ping icmp packet when id & seq are the same) could cause ping to read an arbitrarily large timeval and thus cause ping to wait for response for an arbitrarily large value of time.

In this change, if a pattern was supplied in outgoing echo request, then it is validated on the incoming echo response so that ping's icmp packets are not confused with other applications' icmp packets

### DIFF
--- a/ping.c
+++ b/ping.c
@@ -1103,6 +1103,8 @@ ping4_parse_reply(struct socket_st *sock, struct msghdr *msg, int cc, void *addr
 	if (icp->type == ICMP_ECHOREPLY) {
 		if (!is_ours(sock, icp->un.echo.id))
 			return 1;			/* 'Twas not our ECHO */
+		if (!contains_pattern_in_payload((__u8*)(icp+1)))
+			return 1;			/* 'Twas really not our ECHO */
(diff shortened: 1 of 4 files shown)


okias:
1) Could you please Signed-off-by code?

2) Also, why there is added part about Amazon copyright? Are you affilated with Amazon?

locasity:
1. Is there a signed-off-by template? 
2. Yes, I work at Amazon and the company requires that I include the Amazon copyright on my contribution.

okias:
1. http://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for
2. well, I really considering what to do about it. You add more comments lines with copyright than code. I really appreciate contribution, but I'm kinda afraid what's going to happend when every contributor add 3 lines of comments in every affected file.

locasity:
Thanks, I've signed-off-by 'ed.

I'll check company policy around moving the copyright notice into the commit message.

locasity:
Additional copyright notices dropped

okias:
Thank you! :)

Issue #68, opened 2016-11-10
 traceroute6: sendto: Invalid argument


nmeyerhans:
Forwarded from Debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843264

traceroute6 does not work at all:

tglase@tglase:~ $ sudo traceroute6 -n www.teckids.org
traceroute to  (2a02:a00:1022:9846::74) from 2a01:4f8:150:946c::42:3, 30 hops max, 24 byte packets
sendto: Invalid argument
 1 traceroute: wrote  24 chars, ret=-1
 *sendto: Invalid argument
traceroute: wrote  24 chars, ret=-1
^C
130|tglase@tglase:~ $ ping6 www.teckids.org
PING www.teckids.org(terra.teckids.org (2a02:a00:1022:9846::74)) 56 data bytes
64 bytes from terra.teckids.org (2a02:a00:1022:9846::74): icmp_seq=1 ttl=54 time=24.6 ms
^C
--- www.teckids.org ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 24.670/24.670/24.670/0.000 ms

The other implementation on the system, however, does:

tglase@tglase:~ $ traceroute6.db -6 -n www.teckids.org
traceroute to www.teckids.org (2a02:a00:1022:9846::74), 30 hops max, 80 byte packets
 1  2a01:4f8:150:946c::42:1  12.458 ms  12.460 ms  12.460 ms
 2  * * *
 3  2a01:4f8::a:15:b  12.445 ms  12.614 ms  12.621 ms
 4  2a01:4f8:0:3::1c5  12.624 ms 2a01:4f8:0:3::1c1  12.628 ms 2a01:4f8:0:3::1c5  12.815 ms
 5  2a01:4f8:0:3::a  16.829 ms 2a01:4f8:0:3::da  17.050 ms  17.059 ms
 6  2001:7f8::62cb:0:1  17.286 ms  16.609 ms  16.616 ms
 7  2a00:13c8:1:d::2  22.053 ms  22.064 ms  21.994 ms
 8  2a01:170:0:2:3:1:0:c18  23.436 ms  23.446 ms  23.652 ms
 9  2a01:170:0:2:3::2  21.965 ms  21.933 ms  21.749 ms
10  2a02:a00:1022:9846::74  24.218 ms  24.003 ms  24.150 ms

nmeyerhans:
[...]

PR #69, opened 2016-11-10
traceroute6: Fix udp packet port specification.

nmeyerhans:
Addresses #68


okias:
@pavlix could you please review this commit?

pavlix:
Sorry for the inconvenience. The port is indeed lost and your patch fixes that by moving the assignment at the right place.

okias:
@pavlix np, thanks for review :)

PR #71, opened 2017-01-06
ping: prevent possible double free after cap_free()

jsynacek:
Original bugreport: https://bugzilla.redhat.com/show_bug.cgi?id=1410114


pavlix:
Trivial change, checked only with my eyes. Thanks!

PR #77, opened 2017-02-18
tracepath,doc: Added -4, -6 Parameter in doc and void usage(void)

seishinryohosha:
Just saw, that the source has [-4,-6] Parameter, but the doc didn't mention it at all.
So I just added it.


pavlix:
Thanks! Docs improvement, merged.

PR #78, opened 2017-02-22
fix multicast setsockopt calls on big endian

jsynacek:
Originally reported at https://bugzilla.redhat.com/show_bug.cgi?id=1373333.


jsynacek:
Hmm, I can still see that the CI hasn't completed yet, but it actually has. Weird.

pavlix:
Size is indeed integer. Trivial fix. Merged.

PR #79, opened 2017-02-23
arping: exit if network disappears while running

jsynacek:
Originally reported at https://bugzilla.redhat.com/show_bug.cgi?id=1387542


pavlix:
Merged. The change was rather trivial.

PR #80, opened 2017-02-28
traceroute6: Fix use-after-free for hostname

natexornate:
The memory pointed to by hostname is freed by freeaddrinfo so we must
copy the hostname to use it later. You can observe this when invoking
traceroute6 like 'traceroute6 www.google.com' and seeing the garbage
printed after "traceroute to ", or using ASAN.

### DIFF
--- a/traceroute6.c
+++ b/traceroute6.c
@@ -341,6 +341,7 @@ int main(int argc, char *argv[])
 	struct sockaddr_in6 from, *to;
 	int ch, i, on, probe, seq, tos, ttl;
 	int socket_errno;
+	char *resolved_hostname = NULL;
 
 	icmp_sock = socket(AF_INET6, SOCK_RAW, IPPROTO_ICMPV6);
 	socket_errno = errno;
@@ -462,7 +463,13 @@ int main(int argc, char *argv[])
 
 		memcpy(to, result->ai_addr, sizeof *to);
 		to->sin6_port = htons(port);
-		hostname = result->ai_canonname;
+		resolved_hostname = strdup(result->ai_canonname);
+		if (resolved_hostname == NULL) {
+			(void)fprintf(stderr,
+			    "traceroute: cannot allocate memory\n");
+			exit(1);
+		}
+		hostname = resolved_hostname;
 		freeaddrinfo(result);
 	}
 	firsthop = *to;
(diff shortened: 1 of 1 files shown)


pavlix:
I can't reproduce the garbage in traceroute6 output.

pavlix:
On the other hand the fix is pretty much obvious, merging.

pavlix:
The fix seems to introduce a memory leak. I know this is just one variable for the whole file but we probably want a clean fix.

natexornate:
We can free the strdup'd memory, but it's used in each each iteration of the main loop (in send_probe), so we can't free it until after the loop, or the exit(0) in the success or tries exceeded case just before the end of the loop. Given that we can't free the memory up until just before exiting the program, do we want to want to free it at all? Should we free it in every error exit case?

I'm looking at the pattern set by sendbuff (used in send_probe) which is also not free'd in either the success case or the many error cases.

Thoughts?

natexornate:
We could switch to using a buffer with size NI_MAXHOST, like we use to print hostnames from returned packets. Or we could just keep hostname = *argv so that we print whatever the command line argument was as opposed to the canonical name. Not sure exactly what behavior we have to have here. It seems that some traceroute implementations print canonical names and some print the name on the command line.

Looking at ways to avoid leaking memory and avoid using memory after it is freed.

pavlix:
@natexornate I don't care about freeing memory in `exit(1)`, `abort()` and similar cases. But I'm generally trying to keep the success path memory-leak-free in order to be able to check it with valgrind and similar tools. It is also good in case some of the code would become a library used by other tools.

My opinion is that we are hit by bad code in this case. So if you have any improvements for the code structure, that might also help. In this case the same variable can point to static or dynamic data which is certainly bad style. On the other hand the bunch of public variables used in the project are also not great. We should decide whether we have a nice way to improve the code while fixing the problem or we want to go the least problematic path which might be creating a separate variable that would be freed and keeping `hostname` just as a pointer to either of the strings.

natexornate:
Thanks for the direction, I agree with your strategy for this code. I'll see what I can do.

natexornate:
Okay, updated the pull request to be cleaner. Frees the hostname memory in the case we resolve a hostname. I think it's safe to change the exit(0) to a break, it leaves us with fewer exit cases.

natexornate:
Sorry, goofed up the pull request there for a minute. It should be back to correct now.

pavlix:
Thank you very much!

PR #84, opened 2017-04-27
Honor setting source to 0.0.0.0. Which is useful for using arping wit…

jelaas:
…hout having a local IP-address.

Cheers,
Jens Låås

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -1116,10 +1116,10 @@ main(int argc, char **argv)
 		exit(2);
 	}
 
-	if (!dad && unsolicited && src.s_addr == 0)
+	if (!dad && unsolicited && source == 0)
 		src = dst;
 
-	if (!dad || src.s_addr) {
+	if (!dad || source) {
 		struct sockaddr_in saddr;
 		int probe_fd = socket(AF_INET, SOCK_DGRAM, 0);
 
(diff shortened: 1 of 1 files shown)


jelaas:
Looks about equivalent to the original patch. It was a while ago and the code is no longer fresh in my head. I cannot remember how this interacts with dad mode. Can you confirm this is correct with respects to the dad mode checks? (I verified that the original patch worked for my use case. That is arping without a local ip address.)

jelaas:
Looks about equivalent to the original patch. It was a while ago and the code is no longer fresh in my head. I cannot remember how this interacts with dad mode. Can you confirm this is correct with respects to the dad mode checks? (I verified that the original patch worked for my use case. That is arping without a local ip address.)

jelaas:
Looks about equivalent to the original patch. It was a while ago and the code is no longer fresh in my head. I cannot remember how this interacts with dad mode. Can you confirm this is correct with respects to the dad mode checks? (I verified that the original patch worked for my use case. That is arping without a local ip address.)

jelaas:
Looks about equivalent to the original patch. It was a while ago and the code is no longer fresh in my head. I cannot remember how this interacts with dad mode. Can you confirm this is correct with respects to the dad mode checks? (I verified that the original patch worked for my use case. That is arping without a local ip address.)

kerolasa:
> Can you confirm this is correct with respects to the dad mode checks? (I verified that the original patch worked for my use case. That is arping without a local ip address.)

Hi @jelaas great to see you still active with iputils. While I kept wpa_supplicant running, and at the same time systemd.networkd down && ip addr flush dev; I reproduced the ethernet up without address condition. During that test I noticed one of my changes was incorrect, and it is now corrected. I'll merge this tested version - and if it does not work keep in touch (preferably in form of another pull request :-).

PR #86, opened 2017-05-11
arping: support -i interval option

booxter:
This option will allow to set time interval to use between packets sent.

It may become handy to work around timing issues in the network. For
example, Linux kernel as of the time of writing by default ignores all
ARP updates that arrive in 1-second interval since previous update for
the same table entry *irrespective of whether the previous update was
ignored*. It means that calling apring with -c 3 may still not affect
Linux network peers because every consequent update is ignored because
of the previous ARP update (that was also ignored).

In this scenario, we would now be able to call arping -i 2 to work that
kernel behaviour around.

Linux fix can be found in: https://patchwork.ozlabs.org/patch/760372/

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -79,6 +79,7 @@ int dad, unsolicited, advert;
 int quiet;
 int count=-1;
 int timeout;
+int interval=1;
 int unicasting;
 int s;
 int broadcast_only;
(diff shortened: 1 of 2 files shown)


jsynacek:
LGTM.

okias:
Thank you for your pull request.

Since I just now fixed documentation generation (man,html),
may I ask you to incorporate your original SGML man entry into new XML structure? Should be easy and quick. Also for testing man pages you can use xsltproc, which is piece of cake! :)

After it, nothing prevents for accepting pull request.

booxter:
Updated for xml transition.

kerolasa:
Conflict resolution is done, and the change in this pull request is pushed to iputils upstream. See 97926373401e4e794fa90f87b42c6cac9c35daf7

Issue #40, opened 2015-12-18
-M pmtudisc_opt is broken in 1c59920


revellion:
I'm probing some network with the -M do (do not fragment) flag for pmtu set with ping.

I ran the ping from one host with an older iputils from before the ping/ping6 merge.

And one on a host that i have 1c59920 on. (Latest iputils in ArchLinux testing repository)

It does seem that the newer iputils doesn't respect the -M flag at all.

When i check with tcpdump i see that it does put fragments out. Rather than the expected behaviour that it doesn't do it. and use the DF bit to indicate that fragmentation is not desired.

Could this be some regression?

okias:
@pavlix Could you look at it? If you can spot regression?

pavlix:
@revellion Pinging IPv4 or IPv6?

n3ph:
Friendly bump. Debian Stretch needs a new release...

jsynacek:
The same has been reported on RHEL7. It **is** a regression

PR #87, opened 2017-05-15
ping: fix pmtu discovery for ipv6

jsynacek:
Fixes: #40


vincentbernat:
I ran into the same problem. Please, merge this fix.

PR #89, opened 2017-06-01
ping_common.c: POLLERRR is ignored in events

metan-ucw:
See 'man poll' or fs/select.c in kernel sources, POLLERR is ignored in
events and if error happens it will be returned regardless.

Signed-off-by: Cyril Hrubis <metan@ucw.cz>


okias:
yes, it seems it's from original code, maybe BSD somehow honored it in time when ping has been included in iputils...

Thank you!

PR #95, opened 2017-08-03
Makefile: Initialize LASTTAG variable only for target which actually uses the variable

Whissi:
Before this commit, we always executed

  `git describe HEAD`

when *LASTTAG* variable was not set and the Makefile was called which added a dependency on git (or you would see an error if *git* command isn't available on your system).

This commit moves the declaration of the *LASTTAG* variable into the target's recipe which is actually using this variable.

Bug: https://bugs.gentoo.org/626994

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,6 @@ TARGETS=ping tracepath traceroute6 clockdiff rdisc arping tftpd rarpd
 LDLIBS=$(LDLIB) $(ADDLIB)
 
 UNAME_N:=$(shell uname -n)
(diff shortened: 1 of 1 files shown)


okias:
Well be moving forward to Mezon, but for now no problem with merging this fix :)

okias:
I decided to fix it differently, but thank you for your pull request which motivated me to do it anyway :)
Fixed by commit ef1c71104f9eb7bbcc5f7977779b9f75359074e4

PR #97, opened 2017-08-14
Simplify IDNA usage and switch to libidn2 when needed

nmav:
That is, use libc's ```AI_IDN``` consistently. In the case IDNA conversions are needed that cannot be delegated to libc, use libidn2 which implements the IDNA2008 standard (libidn implements the currently obsolete IDNA2003).


nmav:
Note that WIP flag in the ping6 change. I'm not sure what is intention of this function so my change may not be entirely correct. I'd appreciate some review here. 

Note also that my goal was to switch to IDNA2008 completely, however, since most programs used libc's IDN functions, I changed the rest to do the same (except ping6).

okias:
I'm on the phone, but I recall some security issues with libidn2, is everything OK?

Do you have some "tests" to run to verify correct behaviour?

Also, these days we're already able yo build iputils with meson.build, so please also include changes to build with meson :)

okias:
```
$ make
...
ping6_common.c: In function ‘niquery_option_subject_name_handler’:
ping6_common.c:460:7: warning: implicit declaration of function ‘idna_to_ascii_lz’ [-Wimplicit-function-declaration]
  rc = idna_to_ascii_lz(name, &idn, 0);
       ^
ping6_common.c:463:4: warning: implicit declaration of function ‘idna_strerror’ [-Wimplicit-function-declaration]
    idna_strerror(rc));
    ^
ping6_common.c:462:19: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘int’ [-Wformat=]
   fprintf(stderr, "ping6: IDN encoding error: %s\n",
                   ^
cc   -O3 -g -fno-strict-aliasing -Wstrict-prototypes -Wall ping.o ping_common.o ping6_common.o -lcap -lidn2 -lnettle -lresolv -lm   -o ping
ping6_common.o: In function `niquery_option_subject_name_handler':
/home/okias/projects/iputils/ping6_common.c:460: undefined reference to `idna_to_ascii_lz'
/home/okias/projects/iputils/ping6_common.c:462: undefined reference to `idna_strerror'

```

nmav:
@okias It seems that by using libidn2 compatibility functions we can compile it even in ubuntu trusty. The current version of the patch does that.

okias:
Just checking - final version? Build seems to be passing, should I check it and merge?

nmav:
Yes, that's my final version.

okias:
Thank you!

PR #101, opened 2017-09-14
ping: Remove workaround for bug in IP_RECVERR on raw sockets

pevik:
IP_RECVERR for raw socket was broken until 2.4.9.

Removed code was added as extern int working_recverr in
iputils-ss010805 (searching in old tarballs).

See also somehow related bug in various distros:
https://bugs.debian.org/709052
https://bugzilla.redhat.com/73660
https://bugzilla.opensuse.org/927831

### DIFF
--- a/ping.c
+++ b/ping.c
@@ -947,17 +947,6 @@ int ping4_receive_error_msg(socket_st *sock)
 
 		acknowledge(ntohs(icmph.un.echo.sequence));
 
-		if (sock->socktype == SOCK_RAW && !sock->working_recverr) {
-			struct icmp_filter filt;
-			sock->working_recverr = 1;
-			/* OK, it works. Add stronger filter. */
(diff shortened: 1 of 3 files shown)


pevik:
@pavlix Could you please review it?

okias:
I have feeling I tried remove this check. but somehow it sometimes caused issue even on modern kernels. I have stronger feeling that description of impact (only on old kernels) not matching actual situation.

okias:
I searched little bit. Let's give a shot to this code removal and let's hope nothing gets broken. Anyway, in case it does, we need to fix it correctly with corresponding comment, not about 2.4 kernels issue.

pevik:
@okias Thanks for merging! I tested it on various kernels (on various distros) : 3.16 (Debian), 4.4 (SUSE SLE 12 SP3), 4.9 (Debian), 4.12.14-1.5 (SUSE SLE15), 4.13.4 (openSUSE). Working on all.

PR #103, opened 2017-10-08
arping: do not use libsysfs to read from /sys

kerolasa:
Kernel documentation gives pretty clear recommendation: do not use libsysfs.
This commit removes use of that library, and replaces functionality with
opening files oneself.

Reference: https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/sysfs-rules.rst


okias:
great, thank you!

PR #107, opened 2017-10-16
ping: remove spurious error message

jsynacek:
The message has no value for the user, even if verbose mode has been requested.

Originally reported at https://bugzilla.redhat.com/show_bug.cgi?id=1501110.


okias:
thank you!

PR #108, opened 2017-10-19
tracepath: Support calling `tracepath` as `tracepath4` or `tracepath6`

Whissi:
While updating Gentoo's ebuild I noticed that it was just luck that *tracepath6* symlink caused IPv6 resolution because IPv6 is the default on most systems. But if you switch the defaults a symlinked *tracepath6* -> *tracepath* will not do what you expect.

Looks like it is only Gentoo installing *tracepath6* symlink. Not sure if we should get rid of the *tracepath6* symlink at all. However, this addition shouldn't hurt.

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -394,6 +394,12 @@ int main(int argc, char **argv)
 	setlocale(LC_ALL, "");
 #endif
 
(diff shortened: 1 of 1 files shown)


okias:
Thank you, merged!

PR #111, opened 2017-10-29
Fix build on MUSL

Whissi:
Add missing AI_IDN and NI_IDN declarations.

Bug: https://bugs.gentoo.org/503914

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -44,6 +44,13 @@
 
 #ifdef USE_IDN
 #include <locale.h>
+
+#ifndef AI_IDN
+#define AI_IDN 0x0040
+#endif
+#ifndef AI_CANONIDN
+#define AI_CANONIDN 0x0080
+#endif
 #endif
 
 #include "SNAPSHOT.h"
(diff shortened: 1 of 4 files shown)


okias:
Please rebase a top of latest IDN2 changes! Thank you 👍

Whissi:
Rebased like requested, but see issue #117 first. Looks like there's a general problem on musl with idn (not just with idn2, previous version shows same problem).

prometheanfire:
ya, this patch is required if using IDN on musl, but also causes runtime failures.

okias:
@prometheanfire can you be more specific about runtime failures on MUSL? Thanks

pevik:
@yvs2014 Thanks for a bug report. Could you please create a separate issue? (ideally with a simple reproducer)

pevik:
> <s>@yvs2014 Thanks for a bug report. Could you please create a separate issue? (ideally with a simple reproducer)</s>

I see you have already done it. Thank you!

PR #113, opened 2017-11-01
doc/ping: Replace deprecated commands in "SEE ALSO" section

pevik:
ifconfig and netstat are still widely used, but it's a time
to propagate their replacements ip and ss.


okias:
Thank you

Issue #88, opened 2017-05-31
Bug in ipv6 ping -I


Bog999:
In ping6_run() there is an if statement which is supposed to set the IPV6_PKTINFO and SO_BINDTODEVICE options, but only the first setsockopt will be executed because there is && instead of || (ping6_common.c line 798)

okias:
@Bog999 can you provide some test-case?

Thank you

Bog999:
connect eth1 and eth2 to two networks (or to the same network). There should be a host with the address 2001::2/64 to perform the test on.

ip -6 addr add 2001::1/64 dev eth1
ip -6 addr add 2001::222/64 dev eth2

"ping 2001::2" - Works as expected
One of these tests will fail because ping ignores the -I flag - The packets will go out of the wrong interface.
ping 2001::2 -I eth1 - Expected: Ping goes out of eth1
ping 2001::2 -I eth2 - Expected: Ping goes out of eth2

pevik:
Confirmed

pevik:
@okias @pavlix I fixed this issue in pull request, but the question it is actually better than the previous behavior, when IP is accessible, but not from the iface we specified.

Bog999:
I believe that this is the functionality for IPv4

PR #115, opened 2017-11-03
ping/IPv6: Enforce using specified interface

pevik:
Previous version just selected another interface if the one required
with -I didn't work and warn about it:
$ ping -I eth2 2001::2
ping6: Warning: source address might be selected on device other than eth2.
PING 2001::2(2001::2) from 2001::1 eth2: 56 data bytes
64 bytes from 2001::2: icmp_seq=1 ttl=64 time=0.784 ms

$ ping -I eth2 2001::2
This enforces it even if it means to fail:
PING 2001::2(2001::2) from 2001::3 eth2: 56 data bytes
From 2001::3: icmp_seq=1 Destination unreachable: Address unreachable

NOTE: I kept checking interface and warning about it as it can be caused
by kernel (added by 63579a1 "ping: Warn if kernel has selected source
address from other interface.")

Fixes: #88

### DIFF
--- a/ping6_common.c
+++ b/ping6_common.c
@@ -763,13 +763,16 @@ int ping6_run(int argc, char **argv, struct addrinfo *ai, struct socket_st *sock
 			    IN6_IS_ADDR_MC_LINKLOCAL(&firsthop.sin6_addr))
 				firsthop.sin6_scope_id = iface;
 			enable_capability_raw();
-			if (
+
+			int err = 0;
 #ifdef IPV6_RECVPKTINFO
-				setsockopt(probe_fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1 &&
-				setsockopt(sock->fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1 &&
+			err |= setsockopt(probe_fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1;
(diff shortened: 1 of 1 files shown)


ping6_common.c
```diff
@@ -763,13 +763,16 @@ int ping6_run(int argc, char **argv, struct addrinfo *ai, struct socket_st *sock
...
-				setsockopt(sock->fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1 &&
+			err |= setsockopt(probe_fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1;
+			err |= setsockopt(sock->fd, IPPROTO_IPV6, IPV6_PKTINFO, &ipi, sizeof ipi) == -1;
 #endif
-				setsockopt(probe_fd, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device)+1) == -1 &&
-				setsockopt(sock->fd, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device)+1) == -1) {
+			err |= setsockopt(probe_fd, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device)+1) == -1;
+			err |= setsockopt(sock->fd, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device)+1) == -1;
+
+			if (err) {
```
Bog999: Please note that this will print SO_BINDTODEVICE in the error message even if only the IPV6_PKTINFO option fails

PR #118, opened 2017-12-10
Update autotools

kerolasa:
Hello,

I see that there is an effort to make meson build system to work. Unfortunately I do not have any experience how to make it work, but I do know how to use autotools so I decided to make that tooling to do what is needed. Even if this is not the final version project needs I am sure that clarifying what build-sys is expected to do will help meson work.

Usual autotools conventions should now work. These include DESTDIR installation, out-of-tree builds, and so on. Libraries are find with pkg-config when possible. I tried to build on a bit older system, and found one will need to --disable-idn2 when on such system. The ninfod is not build by default. If the default should change lets do that as a separate change. I copied ssl library preference hierarcy from meson build - if it is not right I propose to change  both in separate change. Notice that manuals will not be built where xsltproc is not available.


okias:
Thank you for your effort, but in Linux world projects slowly dropping autotools, because of its slowness and complexity. At this moment, I'll only accept meson improvements (since having two build systems is pain).

At some point, you'll have to move to better build system anyway, so if you want I'll be happy if you want improve our Meson build.

kerolasa:
Lets see if / when I have time to learn Meson. Meanwhile doing this really helped me to see what iputils build-sys needs to do. I will leave the branch alive as reference so that Meson gaps are 'easy' to see, depending on how familia one is with autotools.

PR #119, opened 2017-12-11
Provide example of systemd unit file for rarpd

plusky:


pevik:
@okias LGTM.

kerolasa:
This commit was rebased and included in https://github.com/iputils/iputils/pull/149 that integrated the rarpd with meson build.

PR #126, opened 2018-04-24
ping6: Fix device binding

### LINKED ISSUE #88: Bug in ipv6 ping -I

deastoe:
The use of the && operator instead of || within ping6_run() results
in only the first setsockopt() call for IPV6_PKTINFO/SO_BINDTODEVICE
getting executed if it returns success, rather than all setsockopt()
calls as intended.

Fixes #88

Signed-off-by: Duncan Eastoe <deastoe@vyatta.att-mail.com>


deastoe:
This fix addresses the same issue as https://github.com/iputils/iputils/pull/115 and fixes the comment from that PR regarding the error message.

okias:
Thank you!

PR #138, opened 2018-07-21
ping: Fix AI_CANONIDN usage on some systems

pevik:
Commit 99f67db used AI_CANONIDN in a way, which broke compilation on
systems where AI_CANONIDN is not defined in netdb.h (e.g. glibc < 2.3.4,
alternative libcs that don't support IDN: e.g. current musl 1.1.19 and
uClibc-ng 1.0.30) when not using the system libidn2.

Fixes: 99f67db ping: Fix ping name encoded using ACE on C locale

Reported-by: Nicholas Fish

### DIFF
--- a/ping.c
+++ b/ping.c
@@ -207,9 +207,9 @@ main(int argc, char **argv)
 
 #ifdef USE_IDN
 	setlocale(LC_ALL, "");
-#endif
 	if (!strcmp(setlocale(LC_ALL, NULL), "C"))
 		hints.ai_flags &= ~ AI_CANONIDN;
+#endif
 
 	/* Support being called using `ping4` or `ping6` symlinks */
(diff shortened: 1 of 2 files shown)


pevik:
@okias Mentioned in https://github.com/iputils/iputils/commit/99f67db03d12b60f37b3af91f809f9f72bb1d29b#commitcomment-29788897

okias:
not happy about it (just use recent glibc or libidn2...)

I'd like to drop non-IDN paths, because there is already lot of IDN and ping without full support of DNS is not much useful tool + it just complicate things.

Also libidn2 on x64 system is 107K (O3, perf. optimized - NOT SIZE), which makes it on some 32 bit archs probably about half of it's size. Even for small OpenWRT like machines it's not excuse when you need to work with DNS..

PR #140, opened 2018-07-30
fix ping duration time error

kanehuange:
for ping duration time, there are a mistake when convert tv to ms .


kerolasa:
@kanehuange thank you for contribution.

Issue #141, opened 2018-08-02
Time rounding breaks time display


abrasive:
Hi,

the rounding correction merged in commit 5f5a34c1b4d418ea68c27c00d3a0c173dd5560d6 causes erroneous output when the fractional part rounds to 1.
For example, the value 4.996ms should be displayed as 5.00ms, but is displayed as 4.100ms:
```
64 bytes from 10.42.4.4: icmp_seq=110 ttl=64 time=2.56 ms
64 bytes from 10.42.4.4: icmp_seq=111 ttl=64 time=4.100 ms
64 bytes from 10.42.4.4: icmp_seq=112 ttl=64 time=4.76 ms
```

dhxgit:
Should be fixed with PR #142

PR #123, opened 2018-02-18
Ping: Correct rounding of timing displays

bbczeuz:
Previously, ping rounded individual times down to the next lower 10us.
Doing any statistics on these values lead to wrong results. The summary
display is correct, but due to this bug a different value than using the
min/max/avg of the individual results. This commit fixes the rounding.

PR #142, opened 2018-08-03
Fix Rounding Issue #141

dhxgit:
Rounding of triptimes like 17950 usec were printed as 17.10 ms instead of the correct 18.0 ms.
Was: (17950 % 1000 -> 950 + 50 -> 1000 / 100 -> 10)
Should be: (17950 + 50 -> 18000 % 1000 -> 0 / 100 -> 0)

### DIFF
--- a/ping_common.c
+++ b/ping_common.c
@@ -856,11 +856,11 @@ restamp:
 			if (triptime >= 100000)
 				printf(" time=%ld ms", (triptime+500)/1000);
 			else if (triptime >= 10000)
-				printf(" time=%ld.%01ld ms", triptime/1000,
-				       ((triptime%1000)+50)/100);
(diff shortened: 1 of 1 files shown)


dhxgit:
The problem was introduced with PR #123 which intended to fix a displaying error of about 0.5% when rounding is needed.

But also made it so that 17950 usec triptime would have been printed as 17.10 ms, instead of the correctly rounded 18.0 ms, which is an error of 5%.

Or 10950 usec with 10.10 ms current and 11.0 fixed, wrong by about 8.9%.

dhxgit:
Does no maintainer care about the output being wrongly calculated and formatted?

It annoys me every time I have a long running ping to monitor a system.

okias:
thanks, sorry for delay, having lot of work in real life..

PR #144, opened 2018-08-26
Fix mdev calculation and description

kosekmi:
Implicit conversion of int-type divided by int-type results in int-type, but result should be of float-type. Fixed by changing int-type variables to float-type for correct calculation of mdev parameter, which is now in line with freebsd ping implementation.

Additionally, the description of the mdev parameter was fixed: there is no such thing as a "Median deviation". The mdev parameter states the population standard deviation, commonly refered to as "mean deviation". I would also argue to change the parameter name to "stddev" for clarification and to be in line with freebsd ping.

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -475,7 +475,7 @@ loss calculation, although the round trip time of these packets is used
 in calculating the minimum/average/maximum/mdev round-trip time numbers.
 </para>
 <para>
-Median deviation (mdev), essentially an average of how far each ping RTT is from the mean RTT. The higher mdev is, the more variable the RTT is (over time).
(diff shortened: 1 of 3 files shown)


kerolasa:
Times and types are a bit of a mess in ping source files. That said it is not obvious to me why changing a long to double only to cast it back to long makes code better. It would be far better to change all time measurements to use struct timespec and clock_gettime() monotonic time. But that sort of approach requires quite a bit more work than demonstrated in this change, so I'm not going to hold you accountable delivering that.

Could you consider 1) dropping type change 3552a0465694bd3e367125595d15361a9608dfe7 from this pull request, 2) rebase the manual update on top of most recent origin/master 3) change commit message to 'ping: fixed description of mdev parameter'?

kerolasa:
Merged, 266dc1a03a454616ae239d610d0e46987c3e823a & e15b82bb41600c5ac9fa86278000ffcf127abecd.

PR #145, opened 2018-09-13
Fixed comment of pinger functions

53ningen:
"timeval" struct is 8 bytes in x86 but it is 16 bytes in x86_64.
So I think these comments are a little bit confusing.

> The first 8 bytes	and the sequence number is an ascending integer.  The first several bytes of the data portion are used to hold a UNIX "timeval" struct in VAX of the data portion are used to hold a UNIX "timeval" struct in VAX byte-order, to compute the round-trip time.


okias:
Thank you :+1:

PR #147, opened 2018-10-01
doc/ninfod: Fix the debug mode option (-a -> -d)

primeos:
It's a very trivial fix for a typo in the man page of `ninfod`. I noticed this recently and suppose it helps nevertheless :)


kerolasa:
Good catch, merging.

PR #150, opened 2018-10-16
build-sys: Check if optional idn2 was found

victortoso:
In systems without the headers for idn2, the build is failing as we
always set USE_IDN to 1.

Signed-off-by: Victor Toso <victortoso@redhat.com>


okias:
thank you! :)

Issue #112, opened 2017-10-31
ping6 crash while getifaddrs return entries with if name NULL


gpb76:
ping6 segfault 0 while getifaddrs return entries with ifa->ifa_name == NULL. 
https://sourceware.org/bugzilla/show_bug.cgi?id=21812 --> this bug hit, causing ping6 to crash

crashed here ==> if (!strncmp(ifa->ifa_name, device, sizeof(device) - 1)

pevik:
It looks like a corner case. It would be nice to have a way how to reproduce it.

gpb76:
I don't see any easy way to reproduce it.  the system has feature of adding /removing the interfaces continuously and seen once.

kerolasa:
@gpb76 if you are still around it would be nice to get pull request I just created tested where you have the issue.

kerolasa:
Lets hope commit d52fa39545bfa8dd6656a5bfc6cf119c1d57381e makes code robust enough that this issue will not happen in future.

PR #152, opened 2018-10-19
ping: check getifaddrs(3) ifa_name data before use

kerolasa:
The getifaddrs(3) can return invalid data when system call is interrupted.
Issue was fixed in glibc 2.28 (relesed 2018-08-01). It is fair to assume
there are systems with older libc so make the ping more robust and check
ifa->ifa_name is not NULL before using it.

Reference: https://sourceware.org/bugzilla/show_bug.cgi?id=21812
Addresses: https://github.com/iputils/iputils/issues/112


pevik:
@kerolasa There is number of failures in commits https://github.com/iputils/iputils/commits/master. Could you please restart jobs in travis. I suppose, it's just travis connection failure.

PR #159, opened 2018-11-19
Refactor clockdiff

kerolasa:
I think these changes work, but it would be great to get extra pairs of eyes to have look. Two first changes are pretty easy. Even after trying to use clockdiff in all possible ways two last commits (merge & split) remain somewhat suspicious. If a bug did creep in as part of this pull request it's going to be in these two changes.


kerolasa:
29a3536..acb7fbb merged.

PR #160, opened 2018-11-21
ninfod: fix couple warnings

kerolasa:
arithmetic on a pointer to void is a GNU extension [-Wpointer-arith]

comparison of integers of different signs: '__u32' (aka 'unsigned int') and
'int' [-Wsign-compare]

After these it's ok to use warning level 3 in meson build.

p.s. I could not push to origin/master. No idea what happen there..

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('iputils', 'c',
 	default_options : [
 		'c_std=c99',
-		'warning_level=2',
+		'warning_level=3',
 	],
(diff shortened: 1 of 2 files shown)


kerolasa:
Above patch added to 7737480 with small change to abs(read_size); that I think should be abs(result);

Meson warning level change is now in separate commit 29a3536.

kerolasa:
Merged

PR #164, opened 2019-01-29
clockdiff: remove unused localhost info.

rumvadim:
I had some issues with getaddrinfo on one of the aws machines. Couldn't figure out what was wrong with this call, but since ctl.myname is never used, I figured it's easier just to delete it.

I offer this patch, because others may experience the same issue.

### DIFF
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -473,7 +473,6 @@ int main(int argc, char **argv)
 	};
 	struct addrinfo *result;
 	int status;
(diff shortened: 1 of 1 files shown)


kerolasa:
Thank you for contribution. Because the two commits were very much related to each other I squashed them to be a single change, and updated commit message a little bit. See commit 20992d255eca5f6795ca2c5884e91cb0d68c00e5.

PR #170, opened 2019-03-25
doc: Use namespace correctly

jtojnar:
The files declared xmlns:db but did not use the db namespace at all. They did not define the default namespace at all, which coincidentally worked with Docbook 4 stylesheets, making them think the files 
were written in Docbook 4.

I fixed the namespaces of the documents and switched to the correct Docbook 5 stylesheets.

cc @dtzWill


kerolasa:
Thank you @jtojnar for contribution, pull request is rebased and merged: commit c503834519d21973323980850431101f90e663ef.

PR #174, opened 2019-03-28
arping: Fix packets count

sem-gh:
On my system values of CLOCK_MONOTONIC and CLOCK_MONOTONIC_RAW differed by about 18 seconds. Therefore arping -c3 exited immediately after start.
Also I got implicit declaration of function 'offsetof' warning when compiling with USE_CRYPTO=kernel.


kerolasa:
Thank you @sem-gh  for sending these changes, all merged in commits 13c24bc, f65b38b, and c3c04a9.

PR #177, opened 2019-04-15
Use the value of sbindir to install rinfod, rarpd & ninfod

heirecka:
...instead of hard-coding 'sbin'.


kerolasa:
Thank you @heirecka for contribution, change is merged.

PR #178, opened 2019-05-16
Various fixes and improvements for the build system

primeos:
The first commit improves the error messages when building with older versions of Meson (+ updates the project version), the second commit makes `setcap` really optional, and the third commit removes the old dependency on `docbook-xsl`.

So none of these changes are critical fixes, but they should improve the build experience in certain cases.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -221,10 +221,13 @@ config_h = configure_file(
 setcap = find_program('setcap', '/usr/sbin/setcap', '/sbin/setcap', required : false)
 if get_option('NO_SETCAP_OR_SUID')
 	perm_type = 'none'
+	setcap_path = '/dev/null'
 elif cap_dep.found() and setcap.found()
 	perm_type = 'caps'
+	setcap_path = setcap.path()
 else
 	perm_type = 'setuid'
+	setcap_path = '/dev/null'
 endif
 
 ############################################################
@@ -243,7 +246,7 @@ if build_ping == true
 		join_paths(get_option('prefix'), get_option('bindir')),
 		'ping',
 		perm_type,
-		setcap.path()
+		setcap_path
 	)
 endif
 
@@ -263,7 +266,7 @@ if build_traceroute6 == true
 		join_paths(get_option('prefix'), get_option('bindir')),
 		'traceroute6',
 		perm_type,
-		setcap.path()
+		setcap_path
 	)
 endif
 
@@ -276,7 +279,7 @@ if build_clockdiff == true
 		join_paths(get_option('prefix'), get_option('bindir')),
 		'clockdiff',
 		perm_type,
-		setcap.path()
+		setcap_path
 	)
 endif
 
@@ -306,7 +309,7 @@ if build_arping == true
 		join_paths(get_option('prefix'), get_option('bindir')),
 		'arping',
 		perm_type,
-		setcap.path()
+		setcap_path
 	)
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -3,6 +3,7 @@ project('iputils', 'c',
 		'c_std=c99',
 		'warning_level=3',
 	],
+	meson_version : '>=0.39',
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -221,10 +222,13 @@ config_h = configure_file(
 setcap = find_program('setcap', '/usr/sbin/setcap', '/sbin/setcap', required : false)
 if get_option('NO_SETCAP_OR_SUID')
 	perm_type = 'none'
+	setcap_path = '/dev/null'
```
primeos: BTW I'm not sure if that's the best idea, but I didn't have a better one and it should at least result in a failure if the script would ever use it:
```bash
$ /dev/null
-bash: /dev/null: Permission denied
```

And `build-aux/setcap-setuid.sh` only requires `setcap_path` if `perm_type == 'caps'`.
Alternatively we could e.g. use a path that doesn't exist.

pevik:
@kerolasa I guess we want continue to use `vcs_tag` so requiring meson 0.44 is necessary.
But I'd try avoiding to require recent meson releases. Although meson has rapid development so distros update it quite often, changing build system to meson we already put off some old distros. It'd be bad to make the situation worse just because build system (we're a quite basic system utils, not a random software, so we should have very common build system).

primeos:
> I guess we want continue to use vcs_tag so requiring meson 0.44 is necessary.
But in the future I'd try avoid using features which require recent meson releases.

Regarding that: We could theoretically still use `vcs_tag` with older versions if we add something like `fallback : meson.project_version()` but that only requires 0.41 anyway.
But after depending on 0.41 I got the following warning:
```
WARNING: Project targetting '>=0.41' but tried to use feature introduced
in '0.44.0': warning
```
but I haven't checked which feature that was and how many things we'd have to change.

**Update**: I just wanted to check what that feature was but I don't get the second warning anymore (at least with the default configuration). Maybe it was something that I removed. I could provide the additional argument for `vcs_tag` then and require an older version of Meson. The worst-case scenario should be only an additional warning (e.g. if a different build configuration uses an additional feature).

> But in the future I'd try avoid using features which require recent meson releases.

I agree, this sounds like a good idea. By explicitly requiring a minimum version this should also be easier due to the warnings.

primeos:
I managed to get the required version down to 0.39, at least with the default configuration.
Version 0.38.1 fails with the following error message:
```
Meson encountered an error in file doc/meson.build, line 75, column 2:
Array as argument 4 contains a non-string.
```

Which is due to `custom_man_xsl` in:
```
command : [
                                xsltproc,
                                '-o', '@OUTPUT@',
                                xsltproc_args,
                                custom_man_xsl,
                                '@INPUT@',
                        ],
```

Which is a file object/array defined as `custom_man_xsl = files('custom-man.xsl')`.

kerolasa:
Quickly looking the changes in this pull request appear to be fine. I'll do proper read through Saturday, and sort out conflict resolution.

primeos:
Hm, that was actually a bit strange, GitHub showed me the latest release (on the web interface) but didn't let me fetch the latest commit and the tag (`s20190515`). Therefore my latest commit included the change from 37facc7e70139bf5e05448b4412ac71ab3333f62 which caused the conflict. I did a rebase as I was now able to fetch the commit (not sure what happend before, I actually fetched twice as I was confused about the missing tag but din't notice that I was also missing a commit - maybe I hit another server that had an older version cached or something like that, but that shouldn't normally happen I guess...).

kerolasa:
Yesterday I accidentally only pushed tag but not master branch, so situation was a bit confusing. Sorry about that.

primeos:
Ah, that explains it - no problem and thanks for the information :) Then I probably didn't get the tag either as Git couldn't resolve it / retrieve the objects.

kerolasa:
Got around looking to this a bit earlier. I did not find any reasons not to merge, so merged.

PR #185, opened 2019-06-07
meson.build: fix static build with gcrypt

ffontaine:
gcrypt depends on gpg-error so save the result of
cc.find_library('gpg-error') in gpg_error_dep and use it when needed to
fix static build

Fixes:
 - http://autobuild.buildroot.net/results/fb698e3e903869978bd5e69d791ec362317b7981

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -119,6 +119,9 @@ elif crypto == 'none'
 		description : 'If set RFC6744 random does not use any CRYPTO lib.')
 endif
 
+# gpg-error is a dependency of gcrypt
+gpg_error_dep = dependency('gpg-error', required : false)
+
 systemd = dependency('systemd', required : false)
 systemdunitdir = get_option('systemdunitdir')
 if systemdunitdir == '' and systemd.found()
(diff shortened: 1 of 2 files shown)


ffontaine:
I'll update the PR shortly to fix build failure

ffontaine:
I didn't test it when gcrypt was disabled :-/

kerolasa:
I think first version was close what is needed, but without new variable. If I understood the issue correctly gcrypt, and gcrypt only, will require gpg-error. Linking with it in other cases would pull in unnecessary library reference when not needed. So how about changing crypto_dep variable to be array that can hold multiple dependencies, and requiring two when that's the case and one at rest of the time. See below how I think that could be done. BTW it would great to hear if that version will fix all known issues, so please give a try if possible.

https://github.com/kerolasa/iputils/commit/eabe21164bbb29dead1a902b6d6e910f6892e3a7

ffontaine:
I tested your version, it fixes the buildroot's build failure and is indeed a better solution than my proposal.

kerolasa:
Great to hear @ffontaine, eabe21164bbb29dead1a902b6d6e910f6892e3a7 is merged.

PR #186, opened 2019-06-09
meson.build: fix build with NLS

ffontaine:
With some toolchains, intl is needed for NLS support so search for this
library and use if needed

Fixes:
 - http://autobuild.buildroot.org/results/0a8a3efe734ac7fb3a68ba505277681857dc0a3d

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -88,6 +88,9 @@ if opt == true
 	subdir ('po')
 endif
 
+# Some systems needs intl for gettext
+intl_dep = cc.find_library('intl', required : false)
+
 opt = get_option('USE_IDN')
 if opt == true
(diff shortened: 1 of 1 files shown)


ffontaine:
We should probably link with intl_dep only if USE_GETTEXT is enabled however I'll let you enhance this pull request as you're better than me with meson.

kerolasa:
I think we can use intl_dep as array little bit like with gcrypt fix. I modified commit a little bit and pushed it to master as 418d8bea5424d8a14a21fd7928747a233557618c. Let me know if that does not work.

kerolasa:
See also: https://github.com/iputils/iputils/pull/188/

PR #196, opened 2019-07-06
Update shell command in line with comment.

amouat:
The previous change only updated the comment, not the actual code.

This commit updates the code in-line with the comment (removes the effective bit).


amouat:
Oh, I should mention that this was spotted by @tamalsaha

PR #200, opened 2019-07-09
ping: Remove redundant space in output

pevik:
When run ping -I IP_ADDR (i.e. not -I DEVICE) ping prints
redundant space:

$ ping -c1 -I 127.0.0.1 192.168.0.109
PING 192.168.0.109 (192.168.0.109) from 127.0.0.1 : 56(84) bytes of data.

Check for device to avoid it.

+ create print_header() to reduce duplicity.
+ translate only "flow" and "from" (not with surrounding formatting data)

### DIFF
--- a/ping.c
+++ b/ping.c
@@ -804,7 +804,7 @@ int ping4_run(int argc, char **argv, struct addrinfo *ai, socket_st *sock)
 	if (!(packet = (unsigned char *)malloc((unsigned int)packlen)))
 		error(2, errno, _("memory allocation failed"));
 
-	printf(_("PING %s (%s) "), hostname, inet_ntoa(whereto.sin_addr));
+	printf("PING %s (%s) ", hostname, inet_ntoa(whereto.sin_addr));
 	if (device || (options & F_STRICTSOURCE))
 		printf(_("from %s %s: "), inet_ntoa(source.sin_addr), device ? device : "");
 	printf(_("%d(%d) bytes of data.\n"), datalen, datalen + 8 + optlen + 20);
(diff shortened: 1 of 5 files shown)


pevik:
@kerolasa again, given you're busy so I can do the release without this PR.

ping.c
```diff
@@ -1618,3 +1616,28 @@ void ping4_install_filter(socket_st *sock)
 	if (setsockopt(sock->fd, SOL_SOCKET, SO_ATTACH_FILTER, &filter, sizeof(filter)))
 		error(0, errno, _("WARNING: failed to install socket filter"));
 }
+
+void print_header(int family, char *hostname, char *source, char *whereto, uint32_t flowlabel)
+{
+	printf("PING %s(%s) ", hostname, whereto);
```
kerolasa: Like it or not people parse command outputs. Can you explain very carefully why ABI breakage is needed so that the future complaints can be directed to that text, and they convince people whose scripts, monitoring, and who knows what is broken that this was indeed great idea.

pevik: Good point about output being parsed. I wonder if people noticed this behavior. But agree, whether it was deliberate to keep space for parsing or not, maybe some people depend on this formatting.

I might prepare another PR for other changes in this PR (Don't translate PING and reduce duplicity with `print_header()`, which will keep current formatting).

Ad parsing - weird idea: have also json output format (`ip` from [iproute2](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git)) supports it: `ip --json addr`)

PR #201, opened 2019-07-09
Allow user to specify VRF and source IP

jonharrisonms:
ping currently has some limitations for sending a ping in the context of a VRF.
We can send a ping with a specific source IP address.  For example:
   ping 10.1.1.3 -I 10.1.1.2
We can send a ping in the context of a Linux VRF.  For example:
   ping 10.1.1.3 -I vrf_red
However, when pinging in the context of a VRF, Linux does not always choose a sensible source IP address – the source IP might not belong to the VRF.  As a result, the ping won’t get a response.   As a result, we want to be able to specify both a VRF and a source IP address when initiating a ping.  For example:
   ping 10.1.1.3 -I vrf_red -I 10.1.1.2

Ping reads in the command line parameters fine and sets up the 'source' and 'device' variables, but currently ignores the device if the source IP address is non-zero.  I've added a branch to ping.c that does the socket bind to the device even in the case where the source IP is non-zero.  This branch is based on the existing case where source IP is zero, but simplified a bit because we've already got a source IP address to use.  
I've tested this fix on a box with a VRF configured and the above syntax now works fine.

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -269,11 +269,15 @@ xml:id="man.ping">
         <listitem>
           <para>
           <emphasis remap="I">interface</emphasis> is either an
-          address, or an interface name. If
+          address, an interface name or a VRF name. If
           <emphasis remap="I">interface</emphasis> is an address, it
           sets source address to specified interface address. If
-          <emphasis remap="I">interface</emphasis> in an interface
+          <emphasis remap="I">interface</emphasis> is an interface
           name, it sets source interface to specified interface.
+          If <emphasis remap="I">interface</emphasis> is a VRF
+          name, each packet is routed using the corresponding
+          routing table; in this case, the <option>-I</option> option
+          can be repeated to specify a source address.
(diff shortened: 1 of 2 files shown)


kerolasa:
This pull request has pretty good explanation why this is needed, but git commit message is essentially empty. I would much rather that text in git commit, and less effort in transient pull request message. Well now that the pull request is good no need to delete the text, but could you copy that to commit message.

Secondly. User interface changes really need explaining and example in manual page. Again the commit message text is good start to get that done.

pevik:
\+1, you can add my:
Acked-by: Petr Vorel <petr.vorel@gmail.com>

jonharrisonms:
Thanks for the helpful comments.  I've updated the commit message and added some man page text.  Please let me know if there's anything else I should do.

kerolasa:
Sorry @pevik I missed your ack when pushing this to origin/master, that said 9e08707d743b29e853df81bd7def1729e3afe55d is the upstream commit that  contains your change @jonharrisonms.

Issue #203, opened 2019-07-10
Disable setcap and setuid for arping (prevent ARP Poisoning)


pevik:
Disable setcap and setuid for arping as it can be used for ARP Poisoning
https://en.wikipedia.org/wiki/ARP_spoofing

pevik:
@nmeyerhans setuid is also enabled for Debian
https://salsa.debian.org/debian/iputils/blob/master/debian/iputils-arping.postinst

nmeyerhans:
Yes, we do install arping in Debian such that unprivileged users can run it. This has been the case since the beginning of time. Since arping isn't installed by default, and Debian provides an easy mechanism by which an admin can change the permissions of a file like the arping binary in such a way that the change is preserved across package upgrades, this hasn't been considered a major concern. It's worth reconsidering that decision, though. Thanks for raising the issue.

pevik:
This has been fixed by 054670a.

PR #204, opened 2019-07-10
security: Disable setcap and setuid for arping

pevik:
arping can be used for ARP Poisoning
https://en.wikipedia.org/wiki/ARP_spoofing

Fixes: #203

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -226,10 +226,7 @@ config_h = configure_file(
 	configuration : conf)
 
 setcap = find_program('setcap', '/usr/sbin/setcap', '/sbin/setcap', required : false)
-if get_option('NO_SETCAP_OR_SUID')
-	perm_type = 'none'
-	setcap_path = '/dev/null'
-elif cap_dep.found() and setcap.found()
+if cap_dep.found() and setcap.found()
 	perm_type = 'caps'
 	setcap_path = setcap.path()
 else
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -312,12 +312,6 @@ if build_arping == true
 		dependencies : [rt_dep, cap_dep, idn_dep, intl_dep],
 		link_with : [libcommon],
 		install: true)
-	meson.add_install_script('build-aux/setcap-setuid.sh',
```
kerolasa: That's  too harsh. I would rather see NO_SETCAP_OR_SUID default to true.

```
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -47,7 +47,7 @@ option('BUILD_MANS', type : 'boolean', value : true,
 option('BUILD_HTML_MANS', type : 'boolean', value : false,
        description : 'Build html manuals')
 
-option('NO_SETCAP_OR_SUID', type : 'boolean', value : false,
+option('NO_SETCAP_OR_SUID', type : 'boolean', value : true,
        description : 'Allow disabling build-aux/setcap-setuid.sh')
```

pevik: But you definitely need it for `ping` and probably good idea to have it for `clockdiff` (both shouldn't have security constraints). Therefore `NO_SETCAP_OR_SUID` is good for distros (packaging tools usually do it with their own tools), but not for users compiling from sources. I was thinking to have also switch for each of these 3 binaries, maybe that'd be the way.

kerolasa: Good point. Perhaps setcaps / setuid should be configurable per application, and arping & clockdiff should default to no extra bits.

While fixing that it would be great to see better variable name. The NO_... leads to weird disable enabling. Just dropping the NO_ prefix, and flipping meaning of true and false will make that much better.

pevik: @kerolasa how about now?

kerolasa:
First impression; very promising. I will try to to have a closer look and use disable / enable mechanism after work.

pevik:
@kerolasa sure, take your time. I'll push it only after adding your ack (if you don't have any comments / requirements to change something).

meson.build
pevik: @kerolasa `traceroute6` permissions does not work. I need to look into it more (even `cap_net_raw+pe` does not work)

kerolasa:
Tested-by: Sami Kerola <kerolasa@iki.fi>

Thank you @pevik for sorting this out, remember to close the bug after merge.

pevik:
@kerolasa thanks for testing, pushed.

PR #206, opened 2019-07-18
tftpd maintenance  commits

kerolasa:
Any objections to move tftpd files to subdirectory, add example xinetd configuration file, and stop building tftpd by default as long it is inetd driven. Sometime earlier I said this should be rewrote, but time is my enemy. Perhaps before  end of year (not telling which one) that'll be done. Meanwhile I do not think there is much sense shipping this pretty useless daemon.

Assuming no negative feedback by 2019-07-21 08:00 UTC-0 I will merge this.


pevik:
Good idea.
Acked-by: Petr Vorel <petr.vorel@gmail.com>

Only one note: could you add get `BUILD_TFTPD=1` to .travis.yml
(other binaries not being build by default I've added in PR https://github.com/iputils/iputils/pull/207).

kerolasa:
Great point @pevik I'll add a commit that will enable everything in travis.

pevik:
If you don't mind, please ack or push https://github.com/iputils/iputils/pull/207 first (if it's ok for you, changes are trivial) to avoid conflict in it with adding `BUILD_TFTPD=1`. I'll be offline most of the weekend.

PR #208, opened 2019-07-26
Makefile: Make all targets phony

fblaese:
None of the Makefile targets create a file called like the target.
Therefore all targets are declared to be phony.


kerolasa:
Thank you @fblaese

PR #212, opened 2019-09-07
Fixes of recent changes in ping and arping

pevik:


kerolasa:
Feel free to merge, although the revert now makes checking what happen with memory with valgrind annoying again.

pevik:
> Feel free to merge, although the revert now makes checking what happen with memory with valgrind annoying again.

Hm, not happy to keep code just to make `valgrind` happy (`ping` isn't a long running daemon and it's definitely not eating memory), but understand your concern.

Why simple free wasn't enough (why return was needed)? I'd use simple
```
diff --git ping.c ping.c
index 508e3d2..4edc163 100644
--- ping.c
+++ ping.c
@@ -890,6 +890,7 @@ int ping4_run(int argc, char **argv, struct addrinfo *ai, socket_st *sock)
        setup(sock);
 
        main_loop(&ping4_func_set, sock, packet, packlen);
+       free(packet);
 }
 
 static void pr_options(unsigned char *cp, int hlen)
diff --git ping6_common.c ping6_common.c
index a38218c..d14f2e3 100644
--- ping6_common.c
+++ ping6_common.c
@@ -844,6 +844,7 @@ int ping6_run(int argc, char **argv, struct addrinfo *ai, struct socket_st *sock
        drop_capabilities();
 
        main_loop(&ping6_func_set, sock, packet, packlen);
+       free(packet);
 }

```
What am I missing?

arping.c
```diff
@@ -765,7 +765,7 @@ static int event_loop(struct run_state *ctl)
 				}
 				total_expires += exp;
 				if ((0 < ctl->count && (uint64_t)ctl->count < total_expires) ||
-				    (ctl->quit_on_reply && ctl->timeout < total_expires)) {
+				    (ctl->quit_on_reply && ctl->timeout < (int)total_expires)) {
```
pevik: This should cast to `(long)` I guess.

kerolasa:
> Why simple free wasn't enough (why return was needed)? I'd use simple

Some of the allocations are done in main, that are not reachable deeper in call stack. Namely these.

    freeaddrinfo(result);
    free(outpack);

Then again I don't care that much. Obvious improvement to many ping issues is to start janitorial fixes  to bring the source code to this decade, that should address among many other problems proper resource management.

pevik:
> Some of the allocations are done in main, that are not reachable deeper in call stack. Namely these.
> 
> ```
> freeaddrinfo(result);
> free(outpack);
> ```

Thanks for info.
> 
> Then again I don't care that much. Obvious improvement to many ping issues is to start janitorial fixes to bring the source code to this decade, that should address among many other problems proper resource management.

+1

Merged only comparison fix ec821e5 and kept valgrind happy.

PR #213, opened 2019-09-08
tracepath: use clock_gettime() rather gettimeofday()

kerolasa:
The gettimeofday() is obsolete.  Secondly, ensure time jumps do not effect
measurements by using monotonic time.

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -268,7 +268,7 @@ static int send_pack(struct run_state *ctl)
 	memcpy(p, &ctl->gdst, 4);
 	p += 4;
 
-	clock_gettime(CLOCK_MONOTONIC_RAW, &now);
+	clock_gettime(CLOCK_MONOTONIC, &now);
 	err = sendto(ctl->socketfd, buf, p - buf, 0, (struct sockaddr *)HE, sll_len(ah->ar_hln));
 	if (err == p - buf) {
 		ctl->last = now;
@@ -323,7 +323,7 @@ static int recv_pack(struct run_state *ctl, unsigned char *buf, ssize_t len,
 	unsigned char *p = (unsigned char *)(ah + 1);
 	struct in_addr src_ip, dst_ip;
 
-	clock_gettime(CLOCK_MONOTONIC_RAW, &ts);
+	clock_gettime(CLOCK_MONOTONIC, &ts);
 
 	/* Filter out wild packets */
 	if (FROM->sll_pkttype != PACKET_HOST &&
--- a/tracepath.c
+++ b/tracepath.c
@@ -21,6 +21,7 @@
 #include <sys/socket.h>
 #include <sys/time.h>
 #include <sys/uio.h>
+#include <time.h>
 #include <unistd.h>
 
 /*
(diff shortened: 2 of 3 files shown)


pevik:
Acked from my side.

Few notes (feel free to ignore)
* `CLOCK_MONOTONIC_RAW` has been added long time ago (in kernel in 2.6.28), so fallback to `CLOCK_MONOTONIC` is probably too careful.
* just a note at commit message`ensure time jumps do not effect measurements by using monotonic time.` time jumps don't affect both `CLOCK_MONOTONIC` and `CLOCK_MONOTONIC_RAW`, the advantage of `CLOCK_MONOTONIC_RAW` is that it's not affected by NTP either.

BTW we still use `time_t` in clockdiff.c and ninfod/ni_ifaddrs.c, do you plan to fix it (if not, I'll do it).

pevik:
\+ `timespecsub()` looks useful enough to be in some header. But that can be done later, when it's actually needed by more source files.

pevik:
I've found some reports that `CLOCK_MONOTONIC_RAW` can be quite slow (depending on HW). I guess it should be ok for tracepath usage, if not we can always change it to `CLOCK_MONOTONIC`.

kerolasa:
> I've found some reports that CLOCK_MONOTONIC_RAW can be quite slow (depending on HW).

I guess you are referring to this:
http://btorpey.github.io/blog/2014/02/18/clock-sources-in-linux/

The CLOCK_MONOTONIC_RAW seems to have standard deviation around 50 nanoseconds. Considering tracepath claims accuracy of no more than microsecond that is unlikely a problem.

Bare in mind when measuring time difference it is irrelevant how the time stamp relates to external time. If reading time has predictable 100 nanosecond latency it's the same latency both times around when taking measurement. 

> I guess it should be ok for tracepath usage, if not we can always change it to CLOCK_MONOTONIC.

The CLOCK_MONOTONIC is subject to adjtime(3) jumps, while CLOCK_MONOTONIC_RAW is not, see manual page below.

http://man7.org/linux/man-pages/man3/clock_gettime.3.html

But now when I think this a bit more perhaps adjtime() speed up / slowdown of time ought to be taken account when measuring time differences. After all the whole point of ntp adjustment is to inform local hardware clock is not ticking right speed. Obviously wrongly measured second will effect to how  long does it take time to send and receive packages.

TL;DR yes, changing to CLOCK_MONOTONIC is the right thing to do.

kerolasa:
@pevik a commit da9a6105b2cab2f92135a282d00cc939760e64c0 added to use CLOCK_MONOTONIC in all files.

kerolasa:
> clockdiff.c and ninfod/ni_ifaddrs.c do you plan to fix

I did not (yet). clockdiff might be useful to fix. Quick read of code in ninfod makes suspect thinking is wrong more than just a call to obsolete function. @pevik if you like feel free to take clockdiff fix, I'll try to sort ninfod - but these should be separate pulls.

> timespecsub() looks useful

If / when that is in use elsewhere in project moving the definition to iputils_common.h is fine. That's possibly when clockdiff is sorted out.

kerolasa:
Branch below has ninfod and pseudo-random number generator initialisation fixes. I'll make review request when two currently open pulls are merged.

https://github.com/kerolasa/iputils/commits/more_gettimeofday_removals

pevik:
> > clockdiff.c and ninfod/ni_ifaddrs.c do you plan to fix
> 
> I did not (yet). clockdiff might be useful to fix. Quick read of code in ninfod makes suspect thinking is wrong more than just a call to obsolete function. @pevik if you like feel free to take clockdiff fix, I'll try to sort ninfod - but these should be separate pulls.

OK, I'll do `clockdiff`. Sure, both should be separate PRs. This one is acked from my side.

pevik:
FYI: I also removed more_gettimeofday_removals branch (https://github.com/iputils/iputils/commits/more_gettimeofday_removals) which you put (probably by accident) to the project instead to your fork.

PR #214, opened 2019-09-09
More gettimeofday removals

kerolasa:
Hi @pevik sorry if you already spent dev time with clockdiff(8) gettimeofday(2) removal. I had a look of that change, and it decided see if I could get the change done. See the 72e5aae53d471fc3a94cf30504d413d32a4772ca for results, that also makes use of ppoll(2) rather than select(2) to avoid unnecessary time unit conversions. When asking if you could have a look I did not realise there was also need to fix file descriptor wait code.

Anyway, here's some changes. Would be great if you could have a quick look.

### DIFF
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -61,6 +61,7 @@
 #include <netinet/in.h>
 #include <netinet/ip.h>
 #include <netinet/ip_icmp.h>
+#include <poll.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -116,9 +117,8 @@ struct run_state {
 };
 
 struct measure_vars {
-	fd_set ready;
-	struct timeval tv1;
-	struct timeval tout;
+	struct timespec ts1;
+	struct timespec tout;
 	int count;
 	int cc;
 	unsigned char packet[PACKET_IN];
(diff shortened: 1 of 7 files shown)


pevik:
@kerolasa  I'm sorry to delete more_gettimeofday_removals (mentioned at https://github.com/iputils/iputils/pull/213#issuecomment-529669032). Fortunately commit is still there (7622a8c )
BTW proper way would be to have it in your fork.

pevik:
Just a detail `As discussed in pull request` in commit message: referencing to git commit hash is more friendly to people (sometimes) working offline.

ninfod/ninfod_core.c
```diff
@@ -396,26 +382,31 @@ static int ni_send_fork(struct packetcontext *p)
...
 		return 0;
 	}
 
-	timersub(&tv, &last, &sub);
+	sub.tv_sec = now.tv_sec - last.tv_sec;
+	sub.tv_nsec = now.tv_nsec - last.tv_nsec;
+	if (sub.tv_nsec < 0) {
+		sub.tv_sec--;
+		sub.tv_nsec += 1000000000L;
+	}
```
pevik: This is a duplicity with `timespecsub` from `tracepath.c` (9ada40b). No big deal, but how about moving `timespecsub` into  `iputils_common.h` and use it here?

ninfod/ninfod_core.c
```diff
@@ -396,26 +382,31 @@ static int ni_send_fork(struct packetcontext *p)
...
```
pevik: You also replaced `select()` with `ppoll()`. No problem with that (actually agree with that), but it'd be nice to mention it (+ ideally your reasons for the change).

clockdiff.c
```diff
@@ -198,20 +198,19 @@ static int measure_inner_loop(struct run_state *ctl, struct measure_vars *mv)
...
-		mv->tout.tv_usec = (tmo - (tmo / 1000) * 1000) * 1000;
+		mv->tout.tv_nsec = (tmo - (tmo / 1000) * 1000) * 1000000;
 	}
 
-	if ((mv->count = select(FD_SETSIZE, &mv->ready, NULL, NULL, &mv->tout)) <= 0)
+	if ((mv->count = ppoll(&p, 1, &mv->tout, NULL)) <= 0)
 		return BREAK;
 
-	gettimeofday(&mv->tv1, NULL);
+	clock_gettime(CLOCK_REALTIME, &mv->ts1);
```
pevik: `CLOCK_REALTIME` requires `CAP_SYS_TIME`. But I guess all places which use `iputil_srand_fallback()` also require privileges (at least ninfod and `niquery_init_nonce()` in `ping`, I suppose `rdisc` needs privileges as well).

kerolasa: CAP_SYS_TIME is required when setting the time. Reading is unprivileged operation.

http://man7.org/linux/man-pages/man7/capabilities.7.html see CAP_SYS_TIME

pevik: ok. Acked from my side anyway.

PR #215, opened 2019-09-10
More gettimeofday removals

kerolasa:
Follow up to pull requests #213 & #214. In this revision feedback mostly from #214 is added to commits, and clockdiff(8) commit message has couple words why ppoll(2) replaced select(2) calls.

If I don't hear anything by Sunday 2019-09-15 I will just merge these change. I am optimistic most of the review already happen in couple earlier pulls.


pevik:
Thanks for addressing issues in #214 :)

`Acked-by: Petr Vorel <petr.vorel@gmail.com>`

PR #216, opened 2019-09-15
ping -- big changes

kerolasa:
Commit 1.  Move ping sources to sub-directory.

Commit 2. Remove global variables and move them to runtime structure. See commit message for  further explanation why this is preferable.


pevik:
@kerolasa LGTM, but give me a day or two to find time to look on it more deeply.

kerolasa:
No problem. This is pretty big change, so long review is highly advisable.

pevik:
I wonder if it would be too bad to have `struct ping_rts` as static instead of passing it.

ping/ping.h
```diff
@@ -0,0 +1,402 @@
...
+		opt_so_dontroute:1,
+		opt_sourceroute:1,
+		opt_strictsource:1,
+		opt_tclass:1,
+		opt_timestamp:1,
+		opt_ttl:1,
+		opt_verbose:1;
+};
+/* FIXME: global_rts will be removed in future */
+extern struct ping_rts *global_rts;
```
pevik: Also, what is a TODO (what should be a replacement)? It looks to me a good idea to have struct for options.

ping/ping.c
```diff
@@ -57,6 +57,9 @@
 #include <ifaddrs.h>
 #include <math.h>
 
+/* FIXME: global_rts will be removed in future */
+struct ping_rts *global_rts;
+
```
metan-ucw: What is actually wrong with having it as a global variable? Once the parameters are packed in a structure it solves the problems you describe with the commit description nicely and passing this pointer to all functions complicates things for no good reason. So unless you plan to pass a different pointer to different functions at runtime it makes more sense to keep it as a global variable.

ping/ping.h
kerolasa: The signalfd() and event loop. That global is needed only in signal handler.

ping/ping.c
kerolasa: Moving toward better scoping is best to by having some scoping, how ever little it might be at the beginning.

pevik: I would really use static struct until it's needed.

kerolasa: Passing the struct conditionally happens already. Meanwhile since it was required to make this change bigger to make a point I really need this runtime structure to make overall program structure better here is next commit 99aa4091b8cd003f259a668aa302106594085470

p.s. There were couple bugs in global runtime struct commit that should be fixed.

PR #217, opened 2019-10-15
ping: use random value for the identifier field

teknoraver:
ping historycally used to fill the Identifier field (see RFC 792) with
its PID:

    # tcpdump -tnni lo icmp &

    # ping -q 127.0.0.1 &
    IP 127.0.0.1 > 127.0.0.1: ICMP echo request, id 347, seq 1, length 64
    IP 127.0.0.1 > 127.0.0.1: ICMP echo reply, id 347, seq 1, length 64
    IP 127.0.0.1 > 127.0.0.1: ICMP echo request, id 347, seq 2, length 64
    IP 127.0.0.1 > 127.0.0.1: ICMP echo reply, id 347, seq 2, length 64

    # ps 347
      PID TTY      STAT   TIME COMMAND
      347 hvc0     S      0:00 ping -q 127.0.0.1

This can be seen as an information leak, let's use a random number as id.

Signed-off-by: Matteo Croce <mcroce@redhat.com>

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -296,7 +296,7 @@ void print_timestamp(struct ping_rts *rts)
 /*
  * pinger --
  * 	Compose and transmit an ICMP ECHO REQUEST packet.  The IP packet
- * will be added on by the kernel.  The ID field is our UNIX process ID,
+ * will be added on by the kernel.  The ID field is a random number,
  * and the sequence number is an ascending integer.  The first several bytes
  * of the data portion are used to hold a UNIX "timeval" struct in VAX
  * byte-order, to compute the round-trip time.
@@ -521,7 +521,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 	}
 
 	if (sock->socktype == SOCK_RAW)
-		rts->ident = htons(getpid() & 0xFFFF);
+		rts->ident = rand() & 0xFFFF;
 
 	set_signal(SIGINT, sigexit);
 	set_signal(SIGALRM, sigexit);
--- a/ping/node_info.c
+++ b/ping/node_info.c
@@ -93,7 +93,6 @@ int niquery_is_enabled(struct ping_ni *ni)
 void niquery_init_nonce(struct ping_ni *ni)
 {
 #if PING6_NONCE_MEMORY
-	iputils_srand();
 	ni->nonce_ptr = calloc(NI_NONCE_SIZE, MAX_DUP_CHK);
 	if (!ni->nonce_ptr)
 		error(2, errno, "calloc");
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -501,6 +501,8 @@ main(int argc, char **argv)
 	if (!argc)
 		error(1, EDESTADDRREQ, "usage error");
 
+	iputils_srand();
+
 	target = argv[argc - 1];
 
 	rts.outpack = malloc(rts.datalen + 28);
@@ -1394,7 +1396,7 @@ in_cksum(const unsigned short *addr, int len, unsigned short csum)
 /*
  * pinger --
  * 	Compose and transmit an ICMP ECHO REQUEST packet.  The IP packet
- * will be added on by the kernel.  The ID field is our UNIX process ID,
+ * will be added on by the kernel.  The ID field is a random number,
  * and the sequence number is an ascending integer.  The first several bytes
  * of the data portion are used to hold a UNIX "timeval" struct in VAX
  * byte-order, to compute the round-trip time.
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -145,7 +145,7 @@ struct ping_rts {
 	size_t datalen;
 	char *hostname;
 	uid_t uid;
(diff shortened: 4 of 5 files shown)


ping/ping_common.c
```diff
@@ -521,7 +521,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 	}
 
 	if (sock->socktype == SOCK_RAW)
-		rts->ident = htons(getpid() & 0xFFFF);
+		rts->ident = rand() & 0xFFFF;
```
pevik: It'd be worth to mention in commit message, that it's only for `SOCK_RAW`. For `SOCK_DGRAM` is id sequential (probably handled by protocol, i.e. kernel icmp implementation).

"historycally" => historically :)

I guess `rand()` is ok for us

teknoraver: You mean the one enabled by `net.ipv4.ping_group_range` right?
unfortunately all distro disables it due to past security issues

kerolasa:
I'm not sure about this change. The ident value is used in ping/ping_common.c function is_ours() to identify if a reply package is originated from the process running or some other process. Sure enough a pid value can overflow 16 bit int, but it is still a little bit better way to distinct one process form another than random number that collide much more easily.

teknoraver:
Hi @kerolasa,
It tooks me 20 seconds to get a pid with the same lowest 16 bit as a running ping with the following program:

```C
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(int argc, char *argv[])
{
	int pid;

	if (argc < 3)
		return 1;

	pid = atoi(argv[1]);

	while ((getpid() & 0xffff) != pid)
		if (fork() > 0)
			return 0;

	printf("Running with PID %d\n", getpid());

	execvp(argv[2], argv + 2);

	return 0;
}
```

This is the output in the previously running ping:

```
$ ping 192.168.85.1
PING 192.168.85.1 (192.168.85.1) 56(84) bytes of data.
64 bytes from 192.168.85.1: icmp_seq=1 ttl=64 time=0.523 ms
64 bytes from 192.168.85.1: icmp_seq=2 ttl=64 time=0.497 ms
64 bytes from 192.168.85.1: icmp_seq=3 ttl=64 time=0.487 ms
64 bytes from 192.168.85.1: icmp_seq=4 ttl=64 time=0.519 ms
64 bytes from 192.168.85.1: icmp_seq=5 ttl=64 time=0.492 ms
64 bytes from 192.168.85.1: icmp_seq=6 ttl=64 time=0.491 ms
64 bytes from 192.168.85.1: icmp_seq=7 ttl=64 time=0.488 ms
64 bytes from 192.168.85.1: icmp_seq=8 ttl=64 time=0.429 ms
64 bytes from 192.168.85.1: icmp_seq=9 ttl=64 time=0.450 ms
64 bytes from 192.168.85.1: icmp_seq=1 ttl=64 time=0.472 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=10 ttl=64 time=0.449 ms
64 bytes from 192.168.85.1: icmp_seq=2 ttl=64 time=0.484 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=11 ttl=64 time=0.495 ms
64 bytes from 192.168.85.1: icmp_seq=3 ttl=64 time=0.499 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=12 ttl=64 time=0.494 ms
64 bytes from 192.168.85.1: icmp_seq=4 ttl=64 time=0.499 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=13 ttl=64 time=0.460 ms
64 bytes from 192.168.85.1: icmp_seq=5 ttl=64 time=0.481 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=14 ttl=64 time=0.486 ms
64 bytes from 192.168.85.1: icmp_seq=6 ttl=64 time=0.481 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=15 ttl=64 time=0.483 ms
64 bytes from 192.168.85.1: icmp_seq=7 ttl=64 time=0.487 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=16 ttl=64 time=0.487 ms
64 bytes from 192.168.85.1: icmp_seq=8 ttl=64 time=0.500 ms (DUP!)
64 bytes from 192.168.85.1: icmp_seq=17 ttl=64 time=0.489 ms
64 bytes from 192.168.85.1: icmp_seq=9 ttl=64 time=0.485 ms (DUP!)
```

Another issue is: containers. I can start two identical docker instances running a PID, and they both will have a very low PID for the ping executable.
If they share the same IP address, most likely they will clash.

kerolasa:
@teknoraver your point about containers is good one. That in mind I am leaning towards merging the changes. But before merge I think three comments need update.

`sed -i 's/our UNIX process ID/a random number/' ping/ping.c ping/ping6_common.c ping/ping_common.c`

kerolasa:
Thank you @teknoraver update is now merged, ref: 5026c2221a15bf13e601eade015c971bf07a27e9

PR #218, opened 2019-10-17
Fixes

zenczykowski:


kerolasa:
In absence of other maintainers commenting anything I assume everyone is fine with these changes. If that turned out to be bad decision for some reasons revert(s) will follow.

PR #225, opened 2019-11-02
Various nit picks

kerolasa:
Couple commits I wrote while back that have sit in my local branch. I though to combine these with ping timing work, but that change has become so big that it's better to get small commits reviewed separately. Also, the ping time stuff is taking lots of time to get done (pun intended), and being busy with that I could concurrently wait feedback.

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -689,7 +689,8 @@ static int event_loop(struct run_state *ctl)
 	uint64_t exp, total_expires = 1;
 
 	unsigned char packet[4096];
-	struct sockaddr_storage from = {};
+	struct sockaddr_storage from;
+	memset(&from, 0, sizeof(from));
 	socklen_t addr_len = sizeof(from);
 
 	/* signalfd */
--- a/ninfod/ninfod.h
+++ b/ninfod/ninfod.h
@@ -31,6 +31,8 @@
  * Author:
  * 	YOSHIFUJI Hideaki <yoshfuji@linux-ipv6.org>
  */
+#ifndef IPUTILS_NINFOD_H
+#define IPUTILS_NINFOD_H
 
 /* definitions */
 #define NINFOD			"ninfod"
@@ -118,3 +120,4 @@ int pr_nodeinfo_ipv4addr(CHECKANDFILL_ARGS);
 void init_nodeinfo_nodename(INIT_ARGS);
 int pr_nodeinfo_nodename(CHECKANDFILL_ARGS);
 
+#endif /* IPUTILS_NINFOD_H */
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -517,7 +517,7 @@ main(int argc, char **argv)
 		error(2, errno, _("memory allocation failed"));
 
 	/* Create sockets */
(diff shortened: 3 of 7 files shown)


pevik:
`Reviewed-by: Petr Vorel <pvorel@suse.cz>`. LGTM (first 3 commits are obviously correct and could be merged straight away), @metan-ucw @nmeyerhans @jsynacek @pavlix, could you please have a look?

nmeyerhans:
Agreed with @pevik, the first 3 changes are no problem. I'd like to review the next 3 in a little more depth. I'm a bit skeptical that they'll improve things in the way they're intended to, so I want a little more time to either convince myself that they're good or come up with convincing evidence that they don't...

nmeyerhans:
Hi @kerolasa. I had hoped to provide something more definitive to express my skeptecism about the removal of the inline keywords and the use of heap space for memory, but I haven't been able to devote the time to do so. So please feel free to merge.

A couple of thoughts on the topic: When running with any optimization flags, as is done in the iputils build configuration, gcc doesn't actually inline any functions marked with the inline keyword anyway. When any `-O` flag is passed to gcc, then inlined functions may be generated inline. So in some sense, you haven't been inlining these functions for a while. (In Debian, we compile with `-O2` and have thus been generating inline code.)  This is detailed [in the gcc documentation](https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Optimize-Options.html#Optimize-Options)

For very small functions, inlining is almost certainly going to result in more efficient code. That's even more the case when the function is only called once within a given C file, as is the case, for example with the `rcvd_set()` function in ping.c. In fact, inlining such functions is so often beneficial that gcc has a `inline-functions-called-once` optimization setting, enabled at all optimization levels, that considers _any_ function, whether explicitly marked inline or not, to be a candidate for inlining. But that hasn't caused any problems yet, given my comments above about `-O` so I think it's not a big deal.

Regarding the use of `malloc()` for memory allocation, it certainly doesn't hurt to avoid allocating memory that you never use. I'm not sure that dynamic memory allocation is a win otherwise, but in this context I don't think it's an issue.

kerolasa:
Hello @nmeyerhans and @pevik. Let me merge the first three changes, and abandon the rest.

What comes inlining sometimes you win, other times not. http://www.cs.technion.ac.il/users/yechiel/c++-faq/inline-and-perf.html is pretty good short explanation how complicated this stuff is. My understanding benefit of having functions in cache is often better than inlining them. Anyway this is not so important that I want to push removal hard.

Reason why I want to  malloc() rather than use stack is pretty simple. I want to fit the runtime configuration to a single memory page so that it is very effectively cached. As a minor thing catching allocation failure is possible with malloc(), while stack allocations are more nasty but since we are talking about couple hundred bytes this is pretty theoretical note.

pevik:
@kerolasa IMHO commit "arping / ping: fix couple warnings" is also obviously correct (backwards compatibility). Feel free to merge it with my ack.

pevik:
I dared to push your fix `arping / ping: fix couple warnings`.

PR #226, opened 2019-11-02
[WIP] Meson path fixes + setup messages

pevik:
The most important of this PR is fix in the last commit and debug messages.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -44,6 +44,7 @@ systemdunitdir = get_option('systemdunitdir')
 if systemdunitdir == '' and systemd.found()
 	systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
 endif
+systemdunitdir = prefix + systemdunitdir
 
 setcap_arping = false
 setcap_clockdiff = false


meson.build
```diff
@@ -44,6 +44,7 @@ systemdunitdir = get_option('systemdunitdir')
 if systemdunitdir == '' and systemd.found()
 	systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
 endif
+systemdunitdir = prefix + systemdunitdir
```
pevik: Not sure, about this, I might miss something.
NetworkManager (which uses both autotools and meson) is ok with not using prefix
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/blob/master/meson.build#L329
`systemd_systemdsystemunitdir = systemd_dep.get_pkgconfig_variable('systemdsystemunitdir')`

But running autotools adds prefix:
`systemdunitdir: $(prefix)/lib/systemd/system`
running meson doesn't:
`systemdunitdir: /lib/systemd/system`
I need to check it further.

kerolasa: When I added `systemdunitdir` I wanted to have something to overwrite pkgconfig value. When prefix is provided should build tooling then ignore pkg config's value by appending the provided to to prefix? That can be argued both ways around.

I would say that the suggestion in commit _meson: Use prefix for systemd directory_ `ignores the prefix, maybe meson bug` is a decision  one way rather than other. Seeing this change from that point of view then if we start to manipulate data provided by pkg the least we that should be done is to add rationality why iputils thinks this is better.

BTW I would restrict comparison to meson builds. Starting to compare apples to pears, or meson to autotools, cmake, scons, bazel, and what ever else there might be is a path to long inconclusive debate. I think it's best to stick as close to meson standards / defaults as possible, as it is the build tool this project has decided to use.

kerolasa: These three look ok.

      meson: Rename rinfod to rdisc
      meson: Print configuration + reuse variables
      meson: Print system paths + use always use prefix

pevik: @kerolasa I'm also still not convinced about the last commit "meson: Use prefix for systemd directory", probably it's not needed. I'll merge only 3 previous commits with your `Reviewed-by:` tag.
BTW here is patch and discussion in buildroot, which inspired me for changing this: https://patchwork.ozlabs.org/patch/1188505/

pevik:
I merged these 3 commits, and left the last one. I'll investigate it further, but I'll probably close it as "meson: Use prefix for systemd directory" is probably invalid.

PR #229, opened 2019-11-10
ping: use struct timespec to measure time

kerolasa:
Use the struct timespec to measure how much is the time, and various related
intervals.  Aim of this change is to be exact how time calculation happen,
so that it as easy as possible to understand what happens with them.

The exponentially weighted moving average changed slightly while writing
this change.  Earlier implementation was so hard to follow that after this
change maths may have changed, but even if they did at least now one should
be able to understand the equation - and fix if needed.

Notice that there is a need for follow up to remove 'next' millisecond int
variable in pinger() and schedule_exit().  That change will be complicated
enough to be separate commit.

p.s. This commit is build on top of  https://github.com/iputils/iputils/pull/225 Depending on what happens in that
pull request rebase, and rework may be needed.  That in mind look only the
last commit within context of this pull request.


ValZapod:
I think we need to have tests smh. Can I propose we use wirehsark/tshark/tcpdump as reference? I suppose those are more precise, right?

ping/ping.c
```diff
@@ -1372,7 +1377,7 @@ int ping4_receive_error_msg(struct ping_rts *rts, socket_st *sock)
 		if (rts->opt_flood) {
 			write_stdout("\bE", 2);
 		} else {
-			print_timestamp(rts);
+			print_timestamp("[", rts, "]");
```
pevik, line 1380: Does not this change output? Not aware of `[ ... ]`.

pevik: OK, looking you moved `[` from `print_timestamp()`. Could you keep it there? Because than the formatting is duplicated on both places (ping IPv4 and IPv6), i.e. wrong direction.

pevik (changes requested):
appart from detail with `[` `]` the rest LGTM on first look.

PR #231, opened 2019-11-19
2019wk46 couple warnings fixes and better clockdiff time format

kerolasa:
These are hopefully none controversial changes.

### DIFF
--- a/ninfod/ninfod.c
+++ b/ninfod/ninfod.c
@@ -151,7 +151,8 @@ __inline__ static const char * log_level(int priority) {
 	}
 }
 
-void DEBUG(int pri, char *fmt, ...)
+__iputils_attribute_format(__printf__, 2, 3)
+void DEBUG(int const pri, char const *const fmt, ...)
 {
 	int saved_errno = errno;
 	va_list ap;
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -55,6 +55,7 @@
 #include <errno.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <getopt.h>
 #include <linux/types.h>
 #include <math.h>
 #include <netdb.h>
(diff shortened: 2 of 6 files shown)


ninfod/ninfod.c
```diff
@@ -151,7 +151,10 @@ __inline__ static const char * log_level(int priority) {
 	}
 }
 
-void DEBUG(int pri, char *fmt, ...)
+#ifdef __GNUC__
+__attribute__((__format__ (__printf__, 2, 3)))
+#endif
```
pevik: @kerolasa Could you please move this as a some definition into iputils_common.h (undef on `#else /* __GNUC__ */`) and use this definition, instead of copy paste whole thing?

pevik:
Timestamp change "clockdiff: use unambiguous timestamp format": LGTM, but do we care about backwards compatibility? Should we also keep the old format? Or even support custom format?

kerolasa:
You are right @pevik. The clockdiff needs `--time-format=iso` time formatting option similar to dmesg.

ninfod/ninfod.c
kerolasa: Okay: 59838bf4b39542abbdd93f58f046c8e9c2b1a43b

ninfod/ninfod.c
```diff
@@ -151,7 +151,8 @@ __inline__ static const char * log_level(int priority) {
 	}
 }
 
-void DEBUG(int pri, char *fmt, ...)
+__iputils_attribute_format(__printf__, 2, 3)
```
pevik: Sorry to tell you now. I'd use `iputils_attribute_format`. Double underscore at the beginning is reserved for libc internals. Using `__iputils_*` prefix is obviously safe, but also `iputils_*` is safe.

kerolasa:
The short options `-T` and `-I` are added in 26032762a7ee576367559a6967bdf68c1d27f64e. I will merge this change at Sunday if tests pass, and there is no further feedback.

pevik:
Very minor thing: I'd also add
`#IPUTILS_ATTRIBUTE_PRINTF __iputils_attribute_format(__printf__, 2, 3)`
in `iputils_common.h` and use this in ninfod. Or is there change, that `__attribute__` with `printf` will ever have other params?

You could also add my Suggested-by tag to third commit :).
Anyway, LGTM, feel free to merge. Thanks for your work.

kerolasa:
I don't see how adding IPUTILS_ATTRIBUTE_PRINTF that will hide the format and their position would make code easier to read, so I will not add that. The leading underscores are removed from final version. Thank you for review @pevik

pevik:
I learned from LTP to hide implementation details. But never mind, it's not important.
BTW: how about creating new release? If you agree, we can create ticket and ping other people to test (kind of release candidate).

kerolasa:
I have couple minor things that could / should / would be nice to have before release. The ping timespec review does not appear to get much attention, so that probably has to be left out this time.

Maybe we should formalize release procedures a little bit more. How about starting the release by announcing incoming release using bug ticket that tells when the release will happen, and stays open for a bit afterwards to capture fallout. The rc releases are a bit problematic. They do not tend to get more testers than the usual community. Or that's at least my former experience, and I remain doubtful if they are worthwhile doing at all.

pevik:
Moving a discussion to #234.

PR #235, opened 2019-11-27
2019wk47 few little changes before next release

kerolasa:
Assuming lgtm scan starts to work after 9e44d98f821a17e69cb204edd5cb6c92eb032ec5 I might find reasons to submit few more changes before next release, but lets first see if that does the trick. Other two changes should be pretty straightforward to review.

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -435,7 +435,7 @@ static int outgoing_device(struct run_state *const ctl, struct nlmsghdr *nh)
 	struct rtattr *ra;
 
 	if (nh->nlmsg_type != RTM_NEWROUTE) {
-		error(0, 0, "NETLINK new route message type");
+		error(0, 0, _("NETLINK new route message type"));
 		return 1;
 	}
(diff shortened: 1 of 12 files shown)


pevik:
Thanks for your work.

Feel free to merge these 2 commits: `build-sys: disable manual builds when xsltproc does not work` and `rdisc: tell in --version output if ENABLE_RDISC_SERVER was defined` (nit: here I'd omit the dot at the end):
```
Tested-by: Petr Vorel <pvorel@suse.cz>
```

Ad localization: I'd consider to completely remove localization, at least from some apps (`arping`, `ping`). Some time ago we discussed that we shouldn't change ping output format, because it might be parsed. I guess more support for translation might just lead people to be forced to run `LC_ALL=C arping` instead of `arping`. Besides we have just Japanese translator. But maybe people will like full localization support

kerolasa:
Sounds good, thank you for review @pevik. I dropped the translation commit, and merged the rest.

PR #238, opened 2019-11-30
Lgtm fixes [release preparations]

kerolasa:
These changes should be pretty easy to review. The ping fix is something it would nice to get @pavlix to have look.


kerolasa:
I think these fixes are fine. Assuming I got that wrong reverts can be done.

Issue #239, opened 2019-11-30
ping icmpv6 nodeinfo queries use non-random nonce


nmeyerhans:
While working on openwrt/packages#10633, I observed some unexpected behavior regarding nonce generation for ICMPv6 node info queries in some scenarios.

If ping is built with `USE_CRYPTO=kernel`, but the kernel does not have support for the md5 algorithm, then we never actually write anything to the buffer allocated for random data, and instead leave it initialized to all zero values. When this happens, the nonce ends up being essentially equivalent to the sequence number, when it should be random. There's no indication given to the user that this has occurred.

Practically speaking, I'm not sure that the risk of spoofed nodeinfo replies is all that great, but [the spec](https://tools.ietf.org/html/rfc4620) does require the nonce to be unpredictable enough to protect against them. I think ping should at the very least print a warning when it's unable to generate a random nonce, and it may be reasonable to abort altogether.

kerolasa:
Great catch @nmeyerhans. I wrote change that will cause execution to stop if / when kernel AF_ALG socket does not work as expected.

But I'm not sure if that is great. Maybe we should copy md5 source code to iputils and get rid of dependency to library or kernel api altogether.  That is what util-linux does, https://github.com/karelzak/util-linux/blob/master/lib/md5.c

nmeyerhans:
When building for a general purpose Linux distro, the openssl dependency isn't a big deal, since it's probably installed already. On the other hand, we have such a limited and specific need for any crypto code, that importing just the md5 routines might make sense. I'll be curious to see what this does to the size of the ping binary.

pevik:
Yep, I wouldn't mind to import md5 code either, if binary size does not dramatically increase.

kerolasa:
Lets measure the size change. After compiling & stripping binary I saw 4096 bytes size difference from 77352 to 81448 bytes. If acceptable earlier 'common: fail execution when USE_CRYPTO=kernel does' change will be removed before merge.

https://github.com/kerolasa/iputils/commit/7cf7bb7a1dc9fe3e55309f2e45c04c2c877367a2

pevik:
Size looks ok for me (~5% increase).
Change not related to this commit: you changed `travis/ubuntu.sh` to be copy of `travis/debian.sh` and `travis/centos.sh` copy of `travis/fedora.sh`. They both are supposed to be a symlinks.

kerolasa:
[...]

PR #240, opened 2019-11-30
nodeinfo: warn if random nonce generation fails

nmeyerhans:
If using the kernel crypto API, but MD5 algorithm support is unavailable,
nodeinfo queries may contain a predictable nonce value. Warn the user via
stderr if this happens.

This is a proposed fix for #239. Open questions:

 * Do we want to warn only once when starting?
 * Should we exit, instead of continuing?


nmeyerhans:
This was more an illustration of the issue than a proper proposal for a fix. Closing it in favor of #241.

Issue #252, opened 2020-01-28
ping -6 fails for hosts with link local addresses


RPigott:
Sorry if this is a duplicate, but I didn't see anything like it. I did see #100, but I'm not sure under what circumstance it is intended to allow scope-less link local ping to work.

I see this behavior in the `ping` from the Arch repos and in a build from master.

The issue I see is the following:
```
$ ping -n -c1 -6 _gateway
PING _gateway(fe80::42b0:76ff:feaf:1578) 56 data bytes
ping: sendmsg: Invalid argument
[...]
```

I tracked down why this happens [here](https://bbs.archlinux.org/viewtopic.php?pid=1884995). I concluded that this failure is caused by a missing scope ID after name resolution and is a limitation of getaddrinfo if we want to only query for only the AAAA record. However, it's worth noting that the dual-stack version works fine, and even selects ipv6 over ipv4:

```
$ ping -n -c1 _gateway
PING _gateway(fe80::42b0:76ff:feaf:1578%wlp2s0) 56 data bytes
64 bytes from fe80::42b0:76ff:feaf:1578%wlp2s0: icmp_seq=1 ttl=64 time=6.28 ms
[...]
```

As I described in the post, this is because it is able to use nss_gethostbyname4_r over nss_gethostbyname3_r, so nss-resolve is able to indicate the scope ID to the caller.

So this definitely feels like a bug to me, and I'd consider writing a patch, but I'm not actually certain this is actionable. I have the following questions:

0. Is this really a limitation of getaddrinfo, or is there something more nss-resolve could be doing to indicate the scope ID to the caller with the provided addrinfo hints?

If it is a limitation of getaddrinfo,

1. Is there an alternative method that could be used to discover the correct scope ID? I cannot think of one.
2. Would ping consider using dual-stack name resolution in getaddrinfo even for ping6 to fix this case? This would cause an irrelevant A record request which sucks, but otherwise sounds OK.
3. Would ping consider using an alternative name resolution interface over getaddrinfo to fix this case? I think probably not. Even if it was agreeable and worth the effort it would cause an inconsistency with applications that resolve with the NSS which is probably undesirable.

ValZapod:
You understand you need to use %iface, right?😝

RPigott:
@ValZapod Yes. That's called a scope ID. Please read my description again and the linked post.

[...]

PR #253, opened 2020-01-30
ping: fix ping6 name resolution for hostnames with link-local addresses

RPigott:
Fixes: #252 

This implements my second suggestion from the issue, which I believe is the least disruptive.
Commit message follows. PTAL.

---

getaddrinfo has several potential interfaces it can use to fulfill hostname
requests with the nss, namely nss_gethostbyname{,2,3,4}_r. It selects
which interface to choose based on the passed hints. Unfortunately, the
only interface with the potential to indicate an appropriate scope id
for link-local IPv6 addresses is nss_gethostbyname4_r, which is
necessarily dual-stack. As a result, getaddrinfo avoids that interface
when the requested address family is not AF_UNSPEC causing ping6
requests to fail for hostnames that resolve to link-local addresses
because of an unspecified scope id, even when they could succeed in the
dual-stack case.

This is a problem when using a local resolver such as systemd-resolved
that synthesizes resource records in some cases containing link-local
addresses. Although `ping -6 _gateway` seems like an appropriate test
for IPv6 connectivity, it fails for the above reason.

This commit causes ping to always request the dual-stack name resolution
behavior, even when it intends only to ping IPv6 hosts. If name
resolution returns an addrinfo list that does not contain an address
from the appropriate family, it intentionally fails with EAI_ADDRFAMILY.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -544,11 +544,29 @@ main(int argc, char **argv)
 	if (rts.tclass)
 		set_socket_option(&sock6, IPPROTO_IPV6, IPV6_TCLASS, &rts.tclass, sizeof rts.tclass);
 
+	/* getaddrinfo fails to indicate a scopeid when not used in dual-stack mode.
+	 * Work around by always using dual-stack name resolution.
+	 *
+	 * https://github.com/iputils/iputils/issues/252
+	 */
+	int target_ai_family = hints.ai_family;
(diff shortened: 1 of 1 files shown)


kerolasa:
Thank you for contribution and very good commit message @RPigott This change was merged as commit d217170ec9478a1983127428ad94db33c8bc5649

PR #255, opened 2020-02-15
traceroute6: add handling for source route header

alexaring:
This patch adds handling for source header. After adding in the nexthdr
evaluation chain it begun to work in my case to run traceroute6 in a
RPL IPv6 source route network.

Note there existing generic extension header which have all mostly the
same header setup as type and length in 8 byte octects... sadly it
doesn't fit for every case. May not the best soltution but it fix my case.
@mcr might give here more input because he tried to explain it to me.

### DIFF
--- a/traceroute6.c
+++ b/traceroute6.c
@@ -517,6 +517,7 @@ static int packet_ok(struct run_state *ctl, int cc, struct sockaddr_in6 *from,
 {
 	struct icmp6_hdr *icp = (struct icmp6_hdr *)ctl->packet;
 	uint8_t type, code;
+	unsigned char *p;
 
 	type = icp->icmp6_type;
 	code = icp->icmp6_code;
@@ -535,6 +536,13 @@ static int packet_ok(struct run_state *ctl, int cc, struct sockaddr_in6 *from,
 			nexthdr = *(unsigned char *)up;
 			up++;
 		}
+		if (nexthdr == 43) {
+			nexthdr = *(unsigned char*)up;
+			p = (unsigned char *)up;
+			p++;
+			p += (*p << 3) + 7;
+			up = (struct udphdr *)p;
+		}
 		if (nexthdr == IPPROTO_UDP) {
 			struct pkt_format *pkt;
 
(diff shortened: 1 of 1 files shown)


traceroute6.c
```diff
@@ -535,6 +536,13 @@ static int packet_ok(struct run_state *ctl, int cc, struct sockaddr_in6 *from,
 			nexthdr = *(unsigned char *)up;
 			up++;
 		}
+		if (nexthdr == 43) {
```
mcr: These headers can occur in any order, so this whole skip process needs to be a loop.

alexaring:
but on which nexthdr values I looking on then? I handle every the same? But UDP works different... okay then I make a switch case and the default is the generic exthdr handling?

mcr:
Alexander Aring <notifications@github.com> wrote:
    > but on which nexthdr values I looking on then? I handle every the same?
    > But UDP works different... okay then I make a switch case and the
    > default is the generic exthdr handling?

yes, switch in a for loop constrained by length of packet, switch case, take
UDP as the break, and ICMP, TCP and SCTP as errors.  There is an RFC on this.
Let me find it.

alexaring:
I know the code has these issues... but was already so before (okay that's not an excuse).
The RFC could be good information for me to really fix it without thinking about xxx possible nexthdr cases. :-)

mcr:
See https://tools.ietf.org/html/rfc7045

kerolasa:
Is anyone working on making the necessary changes to this pull request so that all headers are examined?

alexaring:
No, I still working on to fix bugs in the feature what I was working on which had a source routing header involved.

"so that all headers are examined?" - Hard goal to reach, I think you can do the best you can and provide a simple implementation to add new "exceptions" how to deal with nexthdr foobar.

I mentioned this bug in a paper of mine and would like to see a solution which I can mention as well.

Of course I can test the upcoming solution with my use case of source routing nexthdr, just mention me and I will test it.

pevik:
I'm sorry, I also planned to look into RFC 7045, but haven't done it yet.

pevik:
Closing as we removed traceroute6 in a1394212fd4b3e3259104467d9861909961b219e.

PR #256, opened 2020-02-16
Put IPv6 header NextHeader field definitions into header

pevik:
Instead of hardwire 43 and 44 constants, it's better to copy the
definitions from include/net/ipv6.h from kernel.

### DIFF
--- a/ipv6.h
+++ b/ipv6.h
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ * Copyright (c) 2020 Petr Vorel <petr.vorel@gmail.com>
+ */
+
+#ifndef IPV6_H
+#define IPV6_H
+
+/* Definitions from kernel include/net/ipv6.h */
+
+/*
+ *	NextHeader field of IPv6 header
+ */
+
+#define NEXTHDR_HOP		0	/* Hop-by-hop option header. */
+#define NEXTHDR_TCP		6	/* TCP segment. */
+#define NEXTHDR_UDP		17	/* UDP message. */
+#define NEXTHDR_IPV6		41	/* IPv6 in IPv6 */
+#define NEXTHDR_ROUTING		43	/* Routing header. */
+#define NEXTHDR_FRAGMENT	44	/* Fragmentation/reassembly header. */
+#define NEXTHDR_GRE		47	/* GRE header. */
+#define NEXTHDR_ESP		50	/* Encapsulating security payload. */
+#define NEXTHDR_AUTH		51	/* Authentication header. */
+#define NEXTHDR_ICMP		58	/* ICMP for IPv6. */
+#define NEXTHDR_NONE		59	/* No next header */
+#define NEXTHDR_DEST		60	/* Destination options header. */
+#define NEXTHDR_SCTP		132	/* SCTP message. */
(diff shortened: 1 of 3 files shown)


pevik:
@alexaring FYI

pevik:
@kerolasa maybe we can put only the 2 definitions we actually need (but I'd put all).
And we might also want to bring enum definitions (`DEFAULT_PROBES`, ...) from `<linux/icmpv6.h>` to `icmpv6.h` it's own header (although they're used just here, it's a bit cleaner solution). We could also check for `<linux/icmpv6.h>` and include only when not defined.

alexaring:
Sure using defines are better but can we not first talk about how to fix the issue in general, the nexthdr chain is broken in many cases, then cleanup? I don't know why this is NOW necessary because it obviously conflicts with a bugfix.

pevik:
Sure, this is just a cleanup. Don't worry, rebase will be easy.

kerolasa:
@pevik moving the ICMPV6_DEST_UNREACH and other linux/icmpv6.h enums to this new ipv6.h is fine. Notice that I used enum rather than define because enums are included in gdb and other debuggers with symbolic names making them nicer to look when trying to understand crashes where constants are involved.

Whether this or https://github.com/iputils/iputils/pull/255 should be merged first is a question of which of these work streams becomes ready earlier. Assuming conflict resolution is needed after one or another is merged first I can take care of it which ever comes second.

pevik:
@kerolasa I liked just to copy paste parts from kernel header, but changing it to enum is trivial. I'll try to read RFC https://tools.ietf.org/html/rfc7045 to understand a problem in #255. Cleanup is trivial, it can be done easily.
BTW: After solving #255 and this we could start working on the release.

pevik:
> @pevik moving the ICMPV6_DEST_UNREACH and other linux/icmpv6.h enums to this new ipv6.h is fine. Notice that I used enum rather than define because enums are included in gdb and other debuggers with symbolic names making them nicer to look when trying to understand crashes where constants are involved.

I see 15f24990b5d1430818f346488a5f5e466e89f983. I understand having numeric names is nice, but we mix our application values with values from `<linux/icmpv6.h>` (public header, and some of it's values are used also in tracepath.c, which loads this header) and few `NDISC_` related kernel private definitions from `include/net/ndisc.h`. I regret `<linux/icmpv6.h>` didn't define enum, like `<linux/in.h>`:

```
enum {
  IPPROTO_IP = 0,       /* Dummy protocol for TCP       */
#define IPPROTO_IP      IPPROTO_IP
  IPPROTO_ICMP = 1,     /* Internet Control Message Protocol    */
#define IPPROTO_ICMP        IPPROTO_ICMP
...
```
If we want to keep enum, we'd need to rename our constants (add them some prefix), which is ugly. So I'd be for removing these linux related constants from enum.

Other question is how to handle these linux definitions. We have `iputils_common.h`, where we load  header first and use definition as a fallback. The same approach is used in LTP in so called lapi headers (https://github.com/linux-test-project/ltp/tree/master/include/lapi). But we could take iproute2 approach, they just copy paste kernel headers and keep them updated https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/include, but I guess current approach is more suitable for us. But IMHO we shouldn't add everything into single header, that's why I added definitions into separate file. I'll try to prepare this cleanup soon and then I'll look into issue #255).

PR #263, opened 2020-03-09
Add Ukrainian translation

yurchor:
Tested with msgfmt -vc for validity.

Many thanks in advance for merging.


kerolasa:
Great to see there is interest to get translations, thank you for that @yurchor. That said I just pushed 5094405c7a900722818687c9cf0e3fa3574a6661 that updated iputils.pot file. The old one was rather old, and I think I should make a habit ensuring any changes to translations become immediately part of iputils.pot file. Meanwhile it would be great if you could check the new pot and update the po accordingly. Sorry that you probably wasted a bit of effort with translations.

kerolasa:
Thank you @yurchor  for quick update. I merged the translations, f406ac9272885016d43214c415b9a9972cd3dc55

PR #264, opened 2020-04-11
clang-tidy fixes

neheb:


tracepath.c
```diff
@@ -321,7 +321,8 @@ static int recverr(struct run_state *const ctl)
 			if (rethops >= 0) {
 				if (sndhops >= 0 && rethops != sndhops)
 					printf(_("asymm %2d "), rethops);
-				else if (sndhops < 0 && rethops != ctl->ttl)
+
+				if (sndhops < 0 && rethops != ctl->ttl)
 					printf(_("asymm %2d "), rethops);
 			}
```
pevik, line 327: IMHO this should be:
```
if (rethops >= 0 &&
	   ((sndhops >= 0 && rethops != sndhops) ||
		(sndhops < 0 && rethops != ctl->ttl)))
	   printf(_("asymm %2d "), rethops);
```

ninfod/ninfod_core.c
```diff
@@ -158,11 +158,11 @@ static __inline__ struct subjinfo *subjinfo_lookup(size_t code)
...
 struct qtypeinfo {
-	uint16_t qtype;
 	char	*name;
 	int	(*getreply)(CHECKANDFILL_ARGS);
 	void	(*init)(INIT_ARGS);
 	int	flags;
+	uint16_t qtype;
 };
 
 static struct qtypeinfo qtypeinfo_table[] = {
```
pevik, line 168: I don't understand compilation enough to ack this. https://reviews.llvm.org/D14779 shows this warning is noisy. If really improvement, I'd also move `.qtype =` to the end to not to be confusing.

pevik:
Not sure about commit: "[clang-tidy] use != 0 for compare functions", I don't see it as a big improvement.
Commit "[clang-tidy] fix misplaced cast": is probably correct, but can't it lead this way to overflow?
I've put some comments in the text.
Other commits I'd happily merge. Maybe we merge these non-controversal and talk about the rest.

ninfod/ninfod_core.c
neheb: It's an exercise in struct packing.

pevik: Right, acked this one :).
http://www.catb.org/esr/structure-packing/

neheb:
The != 0 thing is proper usage of those functions since they can return negative. 0 is the only way to return an error.

pevik:
But result for `if (foo)` and `if (foo !=0)` is the same. Or, different `-1` and `1` are both `!= 0`. Or what am I missing?
Anyway, going to merge all, but this commit. @kerolasa do you want to include also this commit?

pevik:
Merged all, but commit "[clang-tidy] use != 0 for compare functions". Thanks for your contributions.
Keeping open to resolve whether take this commit.

neheb:
Added two extra commits.

pevik:
@kerolasa could you please have look on `get rid of ctime_r` as it's related to your changes?
`fix unwanted loss of precision` is IMHO ok. And I still don't see much value in `[clang-tidy] use != 0 for compare functions`, but of course it does not harm.

kerolasa:
Remaining two commits were merged as 93dfb95d48977d151dbe94983e4998959e748aee and d6e1d77dd27f8beba0f35f71427da6df33cb38f9. The `[clang-tidy] use != 0 for compare functions` commit makes code worse so I do not want to merge it. Thank you for contributions @neheb and @pevik

PR #265, opened 2020-05-06
ping: Use C locale for parsing -i/-W without IDN

MofX:
There is no reason to only use C local, if IDN is used.
The parameters have nothing to do with IDN and iputils might
be compiled without IDN support (e.g. on ubuntu), this should not
affect -i and -W

Signed-off-by: Joerg Vehlow <joerg.vehlow@aox-tech.de>

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -177,13 +177,11 @@ static double ping_strtod(const char *str, const char *err_msg)
 	if (str == NULL || *str == '\0')
 		goto err;
 	errno = 0;
-#ifdef USE_IDN
+
 	setlocale(LC_ALL, "C");
-#endif
 	num = strtod(str, &end);
-#ifdef USE_IDN
 	setlocale(LC_ALL, "");
-#endif
+
 	if (errno || str == end || (end && *end)) {
 		error(0, 0, _("option argument contains garbage: %s"), end);
 		error(0, 0, _("this will become fatal error in future"));


pevik:
@MofX thanks for caring of this.

`Signed-off-by: Petr Vorel <pvorel@suse.cz>`

pevik:
BTW I wonder if `setlocale(LC_ALL, "C")` in https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L202
and then reset back https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L231 is needed (originated from cab83b1844f3eb0934cd24fc775a6eceda12193e).
IMHO working hex numbers is perfectly safe in all locals, so I'd be for removing it. If not safe, then it should be also done unconditionally, like your fix.

The same question is for counting the loss in https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L889 floating number computation works regardless of locale setup, IMHO it's not needed.

Not sure if nls (localization in po files) is anyhow affected by `setlocale()`.

MofX:
> BTW I wonder if `setlocale(LC_ALL, "C")` in https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L202
> and then reset back https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L231 is needed (originated from [cab83b1](https://github.com/iputils/iputils/commit/cab83b1844f3eb0934cd24fc775a6eceda12193e)).
> IMHO working hex numbers is perfectly safe in all locals, so I'd be for removing it. If not safe, then it should be also done unconditionally, like your fix.

I guess you are right here. I only fixed what I found to be a real problem for me...

> The same question is for counting the loss in https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L889 floating number computation works regardless of locale setup, IMHO it's not needed.

This is different. It uses %g in printf, that is affected by locale. Question is does ping want to print out numbers in local format or in posix format. Personally I hate when tools like this are localized...

> Not sure if nls (localization in po files) is anyhow affected by `setlocale()`.
Correct me if I am wrong, but gettext - which should be used here - is fully influenced by setlocale (especially with LC_ALL) . So it is probably wrong to setlocale at all, if _ is called.

pevik:
> > IMHO working hex numbers is perfectly safe in all locals, so I'd be for removing it. If not safe, then it should be also done unconditionally, like your fix.
> 
> I guess you are right here. I only fixed what I found to be a real problem for me...

Understand :). I might remove it.

> 
> > The same question is for counting the loss in https://github.com/iputils/iputils/blob/master/ping/ping_common.c#L889 floating number computation works regardless of locale setup, IMHO it's not needed.
> 
> This is different. It uses %g in printf, that is affected by locale. 

Good point, thanks!

> Question is does ping want to print out numbers in local format or in posix format. Personally I hate when tools like this are localized...

+1. I'll probably open a ticket for that to investigate, whether localization is required for all tools. I'd prefer at least `ping` and `arping` were not localized. ATM we have Japanese and Ukrainian translation (the latter ones added recently).

> 
> > Not sure if nls (localization in po files) is anyhow affected by `setlocale()`.
> 
> Correct me if I am wrong, but gettext - which should be used here - is fully influenced by setlocale (especially with LC_ALL) . So it is probably wrong to setlocale() at all, if _ is called.

+1.

Issue #268, opened 2020-05-17
ping6 does not support -F flowlabel option


amitcohen1:
I use Fedora31 and ping6 failed with the following error:
$ ping6 2001:db8:2::2 -F 0 -c 1 -q
ping6: flow labels are not supported

The same command, just without -F option works well.
$ ping6 2001:db8:2::2 -c 1 -q
--- 2001:db8:2::2 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.116/0.116/0.116/0.000 ms
PING 2001:db8:2::2(2001:db8:2::2) 56 data bytes

flowable option works well in iputils-20180629-4 and it's broken in iputils-20190515-5.

kerolasa:
Hi @amitcohen1 I just made a pull request  https://github.com/iputils/iputils/pull/270 that should address the issue. Let see what other people who watch this project have to say about the change, as it has open question of why the headers are behaving so strange way.

kerolasa:
The issue should be fixed, see 60c28a5865b7dcc41946bff7e8ed173bf712c0d1. Thank you for bug report @amitcohen1

amitcohen1:
Thanks!

amitcohen1:
Hi, trying to check the fix and it seems that the issue is still occurs.
I installed iputils from master using - "./configure && make && make install".
$ which arping
/usr/local/bin/arping

The same error is shown:
$ ping6 2001:db8:2::2 -F 0 -c 1 -q
ping6: flow labels are not supported

Did I miss something?

pevik:
I tested it only on `::1`, works for me (and verify that without fix it's not working). Can you check that (+ check version)?
```
$ ./builddir/ping/ping -6 ::1 -F 0 -c 1 -q
PING ::1(::1) , flow 0xf78bb, 56 data bytes

--- ::1 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.066/0.066/0.066/0.000 ms

$ ./builddir/ping/ping -V
ping from iputils s20190709-96-g60c28a5
```

amitcohen1:
Can I declare the default "ping6" to be ./builddir/ping/ping -6?
I want to use it as part of kernel tests without changing the tests.

pevik:
Either use `ping -6` or symlink ping to be `ping6` (equal functionality).

BTW are you using LTP network tests?
[...]

PR #270, opened 2020-05-25
ping: include linux/in6.h and fix related compile errors

kerolasa:
Various #ifdef did not take effect although the linux/in6.h is included in
ping.h header.

Addresses: https://github.com/iputils/iputils/issues/268

### DIFF
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -58,9 +58,25 @@
  *	with net_cap_raw enabled.
  */
 #include <stddef.h>
+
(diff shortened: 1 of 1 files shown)


pevik:
I'd prefer to move the explanation from source to the commit message.
And it would be nice to bisect and put commit which caused regression into commit message as `Fixes`.
I also planned to move these definitions into some common header, but that can wait.

pevik:
`Tested-by: Petr Vorel <petr.vorel@gmail.com>`

PR #272, opened 2020-05-29
ping: strictsource option not set when using -B

deastoe:
Since b3a41a6ed616176dba4b556fc0d90355513447e9 presence of the -B
option has caused the strictroute flag to be set instead of
strictsource.

Signed-off-by: Duncan Eastoe <duncan.eastoe@att.com>


kerolasa:
Than you @deastoe merged as 24e7431f05f4589f2ed22d455adcdc2fd5b47ffc

Issue #276, opened 2020-06-16
ping would be stuck if modify the current time to the previous time


lac-0073:
It would be stucked during ping if i modify the current time to the previous time ,and will not exit normally:

Login 2 windows at the same time
1st:
[root@localhost ~]#  ping  9.82.156.184  -c 20
PING 9.82.156.184 (9.82.156.184) 56(84) bytes of data.
64 bytes from 9.82.156.184: icmp_seq=1 ttl=64 time=0.420 ms
64 bytes from 9.82.156.184: icmp_seq=2 ttl=64 time=0.480 ms
64 bytes from 9.82.156.184: icmp_seq=3 ttl=64 time=0.497 ms
64 bytes from 9.82.156.184: icmp_seq=4 ttl=64 time=0.434 ms
64 bytes from 9.82.156.184: icmp_seq=5 ttl=64 time=0.467 ms
64 bytes from 9.82.156.184: icmp_seq=6 ttl=64 time=0.413 ms
64 bytes from 9.82.156.184: icmp_seq=7 ttl=64 time=0.450 ms
64 bytes from 9.82.156.184: icmp_seq=8 ttl=64 time=0.449 ms

2st:
[root@localhost rpm]# date
Tue Jun 16 19:54:42 CST 2020
[root@localhost rpm]# date -s "2020-06-16 19:52:50"
Tue Jun 16 19:52:50 CST 2020
[root@localhost rpm]# date
Tue Jun 16 19:52:52 CST 2020

and then 1st window will be stucked  until the current time is greater than the time of the initial ping

kerolasa:
This is not a great fix. It would be far better to use `clock_gettime(2)` CLOCK_MONOTONIC.

lac-0073:
> This is not a great fix. It would be far better to use `clock_gettime(2)` CLOCK_MONOTONIC.

I rewrote it like you said. could you please review it again? Thank you.  #278

lac-0073:
> > This is not a great fix. It would be far better to use `clock_gettime(2)` CLOCK_MONOTONIC.
> 
> I rewrote it like you said. could you please review it again? Thank you. #278

And I have verified this modification.

pevik:
[...]

PR #278, opened 2020-06-16
ping:fix ping dead loop problem

lac-0073:
It would be stuck during ping if i modify the current time to the previous time ,and will not exit normally
fix the bug #276


pevik:
Thank you for trying to fix the problem. But why do have 5 commits instead of 1?

PR #280, opened 2020-06-29
Update POTFILES and pot

rffontenelle:
Changes in this PR include:
- ping* files are in ping/ directory, so changed po/POTFILES to reflect that
- ping/node_info.c has 2 translatable strings, but was not include in po/POTFILES
- POT file updated using `meson . builddir && ninja iputils-pot -C builddir`

PR #281, opened 2020-06-29
WIP: Use plural form in arping

rffontenelle:
*arping* has strings with counts which plural form code applies to. It could be presented in single form (probably not very often), but also in plural forms. Many languages have more then one plural form (not English case), so the use of ngettext is required.

Having that said, I tried patching it but it didn't work correctly. POT files receives the plural form strings, but building *arping* will now output the following warning:
```
../arping.c: In function ‘finish’:
../arping.c:286:14: warning: format ‘%d’ expects a matching ‘int’ argument [-Wformat=]
  286 |       "Sent %d probes",
      |             ~^
      |              |
      |              int
```
Indeed, *arping* is printing incorrectly. I'm not sure how to solve this. Help is appreciated.

By the way, this patch created on top of #280

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -282,16 +282,28 @@ static int send_pack(struct run_state *ctl)
 static int finish(struct run_state *ctl)
 {
 	if (!ctl->quiet) {
-		printf(_("Sent %d probes (%d broadcast(s))\n"), ctl->sent, ctl->brd_sent);
-		printf(_("Received %d response(s)"), ctl->received);
+		printf(ngettext("Sent %d probe",
+						"Sent %d probes",
+						ctl->sent));
+		printf(" ");
+		printf(ngettext("(%d broadcast)\n",
+						"(%d broadcasts)\n",
+						ctl->brd_sent));
+		printf(ngettext("Received %d response",
+						"Received %d responses",
+						ctl->received));
 		if (ctl->brd_recv || ctl->req_recv) {
 			printf(" (");
 			if (ctl->req_recv)
-				printf(_("%d request(s)"), ctl->req_recv);
-			if (ctl->brd_recv)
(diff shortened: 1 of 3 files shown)


pevik:
@yurchor Looking at this very old PR. Would you care about plural?

yurchor:
Sure. That would be nice to have it.

pevik:
@rffontenelle you need to use the variable also in `printf()`:

```
printf(ngettext("%d request",
"%d requests",
-ctl->req_recv));
+ctl->req_recv), ctl->req_recv);
```
I rebased and fixed that in https://github.com/pevik/iputils/commits/use-plural-form, but it looks like it does not work in Ubuntu Bionic clang:
https://github.com/pevik/iputils/actions/runs/3405790812/jobs/5664019588

```
/usr/include/stdio.h:318:43: note: passing argument to parameter '__format' here
extern int printf (const char *__restrict __format, ...);
                                          ^
../arping.c:318:12: warning: incompatible integer to pointer conversion passing 'int' to parameter of type 'const char *' [-Wint-conversion]
                                printf(ngettext("%d request",
                                       ^~~~~~~~~~~~~~~~~~~~~~
/usr/include/stdio.h:318:43: note: passing argument to parameter '__format' here
extern int printf (const char *__restrict __format, ...);
                                          ^
../arping.c:323:12: warning: incompatible integer to pointer conversion passing 'int' to parameter of type 'const char *' [-Wint-conversion]
                                printf(ngettext("%d broadcast",
                                       ^~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/stdio.h:318:43: note: passing argument to parameter '__format' here
extern int printf (const char *__restrict __format, ...);
                                          ^
5 warnings and 1 error generated.
```
Not sure what's wrong now.

PR #292, opened 2020-08-24
common: fix infinite loop when getrandom fails

### LINKED ISSUE #291: common: infinite loop when getrandom() fails
A call to getrandom was added in https://github.com/iputils/iputils/commit/ad0256f3543a69ff0d44bca838bade6aac04bcbd .

There is a fallback for when getrandom returns an error; however, in this case, the `break` statement does not exit the while loop:
https://github.com/iputils/iputils/blob/ad0256f3543a69ff0d44bca838bade6aac04bcbd/iputils_common.c#L102-L110

This results in an infinite loop whenever getrandom fails for a reason other than `EINTR`.

---

### more info

The manpage of getrandom on my system says:

> getrandom() was introduced in version 3.17 of the Linux kernel.  Support was added to glibc in version 2.25.

I reproduced the issue on my Banana Pi running sunxi Linux 3.4.104, using iputils-20200821.
Ping hangs using 100% CPU. Using strace, I can see it's looping trying to call `getrandom()`:

```
clock_gettime(CLOCK_REALTIME, {tv_sec=1598291358, tv_nsec=964378872}) = 0
getpid()                                = 1148
getuid32()                              = 1000
getrandom(0xbee08c00, 4, GRND_NONBLOCK) = -1 ENOSYS (Function not implemented)
clock_gettime(CLOCK_REALTIME, {tv_sec=1598291358, tv_nsec=964758305}) = 0
getpid()                                = 1148
getuid32()                              = 1000
getrandom(0xbee08c00, 4, GRND_NONBLOCK) = -1 ENOSYS (Function not implemented)
```

I will provide a patch in a bit.

nuno-silva:
fixes #291

### DIFF
--- a/iputils_common.c
+++ b/iputils_common.c
@@ -101,15 +101,17 @@ void iputils_srand(void)
 #if HAVE_GETRANDOM
 	ssize_t ret;
 
-	while ((ret = getrandom(&i, sizeof(i), GRND_NONBLOCK)) != sizeof(i)) {
+	do {
+		ret = getrandom(&i, sizeof(i), GRND_NONBLOCK);
 		switch(errno) {
 		case EINTR:
 			continue;
 		default:
 			i = iputil_srand_fallback();
+			ret = sizeof(i); // exit the while
 			break;
 		}
-	}
+	} while (ret != sizeof(i));
 #else
 	i = iputil_srand_fallback();
 #endif


kerolasa:
Reviewed and merged with small modification, see 469b41ac89b9f6772ea31df8379669d205be95f8. Thank you for contribution @nuno-silva

nuno-silva:
@kerolasa you're welcome!
I must confess I specifically tried to avoid using `goto`, but it also works ;)

pevik:
Although I also try avoid `goto` here looks less confusing that reusing `ret` for something else.
BTW `goto` is mainly used for error handling like here (you can see in [Linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/) that `goto` very much used like this).
Please next time add your `Signed-off-by:` tag.
Anyway, thanks both for these fixed.

kerolasa:
The one, and the only, good reason to use goto is to break out from nested code blocks. The siglongjmp is the one that cannot be ever justified, or at least I have never found a single example that could not be wrote cleaner way using functions.

Sure, I should have added signed-off-by.

kerolasa:
BTW if someone wants to get rid of the goto the following snippet should do the job.

        while (1) {
                ssize_t ret;

                errno = 0;
                ret = getrandom(&i, sizeof(i), GRND_NONBLOCK);
                if (errno == EINTR)
                        continue;
                if (errno == 0 && ret == sizeof(i))
                        break;
                i = iputil_srand_fallback();
                break;
        }

pevik:
Feel free just to merge it. I plan to work on other fixes I found during the release and build at least some testing.

PR #296, opened 2020-10-15
Gateway address is displayed in reverse order

DominikNeubauer:
On ICMP redirects, the proposed gateway address is displayed in reverse order, while correctly transmitted in the ICMP packet. See output below:

dominik@host:~$ ping 10.240.1.1
PING 10.240.1.1 (10.240.1.1) 56(84) bytes of data.
From 192.168.1.1 icmp_seq=75 Redirect Host(New nexthop: 201.1.168.192)

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1224,7 +1224,7 @@ static void pr_icmph(struct ping_rts *rts, uint8_t type, uint8_t code,
 			struct sockaddr_in sin = {
 				.sin_family = AF_INET,
 				.sin_addr =  {
-					icp ? icp->un.gateway : htonl(info)
+					icp ? icp->un.gateway : ntohl(info)
 				}
 			};
 


pevik:
Are you sure? This was changed in e2e9a2dd4639924614bdbee43907a49134e8da19 and it looks to me correct that for presentation `htonl()` needs to be used.
@teknoraver @kerolasa FYI.

DominikNeubauer:
I would assume that here,  the conversion needed to be from network to host byte order - hence `ntohl()` rather. So this would be a cosmetic change only, since both `htonl` and `ntohl` would equal a byte swap on x86.

Just checked with Ubuntu repositories, apparently the package for 20.04 was built on a code base before the commit you mentioned.

teknoraver:
Please check with both RAW and ICMP sockets, they have different behaviour, this is why I made the change.
You can switch between the two with:
```
sysctl -w net.ipv4.ping_group_range='0 2147483647'
sysctl -w net.ipv4.ping_group_range='1 0'
```

DominikNeubauer:
I'll do when I'm back home on the test machine; but for the time being, please consider this pull request as only a cosmetic change and not an actual bug.

DominikNeubauer:
root@host:~# sysctl -w net.ipv4.ping_group_range='1 0' ; ping 10.240.1.1
net.ipv4.ping_group_range = 1 0
PING 10.240.1.1 (10.240.1.1) 56(84) bytes of data.
From 192.168.1.1: icmp_seq=1 Redirect Host(New nexthop: 192.168.1.201)

root@host:~# sysctl -w net.ipv4.ping_group_range='0 2147483647' ; ping 10.240.1.1
net.ipv4.ping_group_range = 0 2147483647
PING 10.240.1.1 (10.240.1.1) 56(84) bytes of data.
From 192.168.1.1 icmp_seq=1 Redirect Host(New nexthop: 201.1.168.192)

teknoraver:
So, it's broken only for ICMP sockets. They have a different code path than raw ones

pevik:
Thanks both for taking care. BTW any idea what is required to setup Redirect Host? Ideally something virtualized (e.g. network namespaces and iptables)?

teknoraver:
Hi @pevik, it's pretty simple.
If you have a LAN with a gateway, and by mistake use a host as gateway instead of the real gateway, the host will send you a redirect with the IP of the real gateway.

```sh
# create two netns with a veth each
ip netns add c1
ip netns add c2
ip link add name c1 type veth peer netns c1
ip link add name c2 type veth peer netns c2

# link the veth by a bridge
ip link add name lanbr type bridge
ip addr add 192.168.0.254/24 dev lanbr
ip link set lanbr up
ip link set c1 master lanbr
ip link set c2 master lanbr

# set addresses in the same subnet
ip -n c1 addr add 192.168.0.1/24 dev veth0
ip -n c2 addr add 192.168.0.2/24 dev veth0
ip -n c1 link set veth0 up
ip -n c2 link set veth0 up
ip link set c1 up
ip link set c2 up

# add an exit route to c2, and a wrong route to c1
ip -n c2 route add 192.168.1.0/24 via 192.168.0.254
ip -n c1 route add 192.168.1.0/24 via 192.168.0.2

# enable forward and send_redirects in c2
ip netns exec c2 sysctl net.ipv4.conf.all.send_redirects=1 net.ipv4.ip_forward=1
```

You should see a redirect, e.g.:
```
# ip netns exec c1 ping 192.168.1.1
PING 192.168.1.1 (192.168.1.1) 56(84) bytes of data.
From 192.168.0.2: icmp_seq=1 Redirect Host(New nexthop: 192.168.0.254)
From 192.168.0.2: icmp_seq=2 Redirect Host(New nexthop: 192.168.0.254)
--- 192.168.1.1 ping statistics ---
2 packets transmitted, 0 received, 100% packet loss, time 1010ms
```

pevik:
@teknoraver Great, thanks a lot! I'm busy with other projects, but hoping to find some time for iputils in 1-2 weeks.

PR #297, opened 2020-10-19
meson: allow installing systemd units without systemd being installed

andir:
In some environments it is desired to generate the systemd units even
thought systemd doesn't exist there. One of those use cases are
sandboxed builds that only have the build dependencies. This commit adds
support for setting the option INSTALL_SYSTEMD_UNITS to true (in
combination with systemdunitdir) to still install the files without
systemd being available.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,8 +40,10 @@ localedir = join_paths(prefix, get_option('localedir'))
 sbindir = join_paths(prefix, get_option('sbindir'))
 
 systemd = dependency('systemd', required : false)
+install_systemd_units = systemd.found() or get_option('INSTALL_SYSTEMD_UNITS')
+
 systemdunitdir = get_option('systemdunitdir')
-if systemdunitdir == '' and systemd.found()
+if systemdunitdir == '' and install_systemd_units
 	systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
 endif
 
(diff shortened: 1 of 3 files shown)


andir:
CI looks good besides that one transient(?) Debian unstable issue.

kerolasa:
Nice build feature, thank you for contribution @andir the commit is rebased and merged: 950d36f8ba5a669cbc34a7972db611b675725fb5

pevik:
Thank you both for handling this!

pevik:
Thinking about it twice, variable name `INSTALL_SYSTEMD_UNITS` might be a bit misleading, because with systemd installed even `INSTALL_SYSTEMD_UNITS=false` does not stop installing it (Although there is probably no reason for not installing them when systemd installed.). Because the true meaning is `FORCE_INSTALL_SYSTEMD_UNITS` (but not voting for this name), maybe variable description should mention that.

Other option would be to have `systemd.found()` as a default value (if that's possible in meson). This would allow to respect user's decision for false.

andir:
I am really not obsessed by the variable name we pick. `FORCE_INSTALL_SYSTEM_UNITS` is okay with me if you think that is better.

pevik:
@kerolasa WDYT? Leave as it is or improve docs or change variable name?

kerolasa:
I don't mind what is the name of this variable. But the defaulting not install was something I did not think through when merging. The earlier default was install. New default should be to install, with option not to  install. That would be nicer to people who are packaging iputils, I assume most do not want to change build options to disable systemd units as no-one has ever before asked for that.

pevik:
Thx! Let's leave it until anybody complain.

PR #301, opened 2020-11-17
ping: process interrupts in ping*_receive_error_msg

matshch:
If `recvmsg(..., MSG_ERRQUEUE)` returned error, `ping` can misinterpret `EHOSTUNREACH` from `recvmsg` in the main loop as error that happened from his own packets and print the following message:

    ping: recvmsg: No route to host

However, this also could be an error from other ping, that was received by this process because of RAW socket. So it's better to skip this error and process it properly in the next cycle.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1314,8 +1314,11 @@ int ping4_receive_error_msg(struct ping_rts *rts, socket_st *sock)
 	msg.msg_controllen = sizeof(cbuf);
 
 	res = recvmsg(sock->fd, &msg, MSG_ERRQUEUE | MSG_DONTWAIT);
-	if (res < 0)
+	if (res < 0) {
+		if (errno == EAGAIN || errno == EINTR)
+			local_errors++;
 		goto out;
+	}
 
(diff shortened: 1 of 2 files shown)


kerolasa:
Seems reasonable, merged. Thank you for contribution @matshch

pevik:
@kerolasa thanks for handling this. I wasn't sure about the patch.
@matshch any idea how to reproduce the problem? I'm collecting tips reproducers for fixes, something like this: https://github.com/iputils/iputils/pull/296#issuecomment-710044824 (using setup created by network namespaces).
Ping is quite complicated, I plan sooner or later to cover it with tests.

matshch:
I'm not quite sure how to reproduce this with unit test. We were observing this issue on ARM64 platform with high CPU load and many ICMP echoes going around, including high count of unreachable targets. In this case, ping was showing dozens of "No route to host" messages every ~ten seconds, even if no packets were lost by this instance of ping.

pevik:
OK, at least this is a hint, thanks!

Issue #74, opened 2017-02-16
main_loop() requests POLLERR but ignores it when returned


vcaputo:
I believe I've encountered a kernel bug in linux 4.10-rc8 which caused this pathological case to occur:

```
1487242826.199486 poll([{fd=3, events=POLLIN|POLLERR}], 1, 970) = 1 ([{fd=3, revents=POLLERR}]) <0.000026>
1487242826.199578 recvmsg(3, 0x7ffc8e05e260, MSG_DONTWAIT) = -1 EAGAIN (Resource temporarily unavailable) <0.000024>
1487242826.199657 gettimeofday({1487242826, 199681}, NULL) = 0 <0.000021>
1487242826.199733 poll([{fd=3, events=POLLIN|POLLERR}], 1, 970) = 1 ([{fd=3, revents=POLLERR}]) <0.000027>
1487242826.199824 recvmsg(3, 0x7ffc8e05e260, MSG_DONTWAIT) = -1 EAGAIN (Resource temporarily unavailable) <0.000022>
1487242826.199899 gettimeofday({1487242826, 199923}, NULL) = 0 <0.000022>
1487242826.199991 poll([{fd=3, events=POLLIN|POLLERR}], 1, 970) = 1 ([{fd=3, revents=POLLERR}]) <0.000026>
1487242826.200085 recvmsg(3, 0x7ffc8e05e260, MSG_DONTWAIT) = -1 EAGAIN (Resource temporarily unavailable) <0.000022>
1487242826.200159 gettimeofday({1487242826, 200183}, NULL) = 0 <0.000022>
1487242826.200236 poll([{fd=3, events=POLLIN|POLLERR}], 1, 970) = 1 ([{fd=3, revents=POLLERR}]) <0.000028>
1487242826.200350 recvmsg(3, 0x7ffc8e05e260, MSG_DONTWAIT) = -1 EAGAIN (Resource temporarily unavailable) <0.000023>
1487242826.200427 gettimeofday({1487242826, 200451}, NULL) = 0 <0.000022>
1487242826.200503 poll([{fd=3, events=POLLIN|POLLERR}], 1, 969) = 1 ([{fd=3, revents=POLLERR}]) <0.000026>
```

I've already reported it to lkml, but this caused `ping` to spin and burn CPU when arguably it should have exited with an error.  Perhaps the assumption is that a subsequent sendmsg would return error when POLLERR caused poll to return, but that's not happening and sendmsg and recvmsg are both succeeding:
```
[...]

PR #304, opened 2020-12-30
Ping: Resolved packet loss issue.

apriya-git:
Due to EINTR & EAGAIN error found 1 packet loss during flood ping test.
Add- read-write for "EINTR" & "EAGAIN" error in "ping_common.c" file.

Signed-off-by: Anu Priya <apriya@nvidia.com>

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -673,9 +673,9 @@ int main_loop(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock,
 				 * on the socket, try to read the error queue.
 				 * Otherwise, give up.
 				 */
-				if ((errno == EAGAIN && !recv_error) ||
-				    errno == EINTR)
-					break;
+				if ((errno == EINTR && !recv_error) ||
+					(errno == EAGAIN && !recv_error))
+					continue;
 				recv_error = 0;
 				if (!fset->receive_error_msg(rts, sock)) {
 					if (errno) {


apriya-git:
Why merging is blocked?
unable to understand about "base branch requires all commits to be signed"

pevik:
`git commit --amend -S`

pevik:
You now quit on `EINTR`(interrupt) only when no `POLLERR`. Why? That's IMHO wrong. And you should explain the reason in the commit message.

Changing `break` to `continue` is most likely wrong. It does not quit on error, which leads to CPU burn. It's easily reproducible on my machine with `ping ::1`
And again, not explained why you're requesting this change.
Also you haven't updated the comment added in 4aefa8a8988b1439a75e59f227e3de504321e7ab.

Ping is really hard to read and understand, explanation *why* is really important. Also ideally explain, how to reproduce your problem.

Mention file `ping_common.c` is useless (it's obvious from git status).
Not sure what you mean by "Add- read-write".

4aefa8a8988b1439a75e59f227e3de504321e7ab is a good example of commit message, although some valuable info was hidden in the relevant pull request https://github.com/iputils/iputils/pull/75 and issue https://github.com/iputils/iputils/issues/74 (e.g. kernel fix: http://patchwork.ozlabs.org/patch/770504/, https://www.mail-archive.com/netdev@vger.kernel.org/msg171251.html), I'd prefer these links to be in the commit message.

nit: I'm not convinced about the changes, but if I *were*, I'd probably suggest to write the condition like
`if (!recv_error && (errno == EINTR || errno == EAGAIN))`
and have commit subject as `ping: Resolved packet loss issue` (lower letter in ping and no dot).

pevik:
@apriya-git Could you please describe what you were trying to solve and how to reproduce the issue?

pevik:
@apriya-git Gently ping, could you please describe what you were trying to solve and how to reproduce the issue? Does the problem still exists on the current master or the latest release? How to reproduce it?

Issue #290, opened 2020-08-10
Reconsider introduced incompatibility in handling numbers in options


pevik:
Handling -i and -W parameter changed over the time, recently introducing
several *incompatible* changes:

### -i interval option
Until s20121114 ping used `.` (dot) as a decimal separator.

Then in 04d3f79 ("ping,ping6: Do not assume radix point is denoted by `.` (-i option).") released in s20121121 decided to use decimal point from locales (depended on `$LC_NUMERIC`). But even then it depend on `USE_IDN=yes` which does not make much sense.

Recently d865d4c ("ping: Use C locale for parsing -i/-W without IDN") addressed that and called setlocale() unconditionally. Maybe condition based on ENABLE_NLS definition should have been discussed, but that would probably lead just to LC_ALL=C setup in the scripts. People also get confused by old behavior when decimal separator depends on locale as some distros e.g. compile without IDN but other don't (see issue #273) But that introduced *incompatible* behavior for next release, thus documenting it in this commit.

562e0d5 fixed problems for -W, but also introduced yet another *incompatible* behavior as invalid number is no longer fatal error (see below).

d196177 ("ping: Add missing preprocesses checks around `setlocale()` calls") unintentionally restored the old behavior, which as fixed by 7a90bd5 ("ping: Fix handling decimal separator for -i/-W opt").

### -W timeout option
Until s20180629 there was undocumented behavior:

1) timeout silently rounded down to the next lower integral number
2) -W N for N < 1 was rounded to 0 => infinite timeout (obviously a bug)
3) -W Ns worked as expected (sleep for N sec, as 's' was dropped,
this could be considered as a feature)
4) -W Nm -W Nh also slept for N sec instead of N min or hour (obviously
a bug)

918e824 ("ping: add support for sub-second timeouts") released in s20190324 fixed 1) and 2) by allowing real number with locale based decimal separator when USE_IDN=yes configured. But that introduced *incompatible* behavior as broke 3 and 4). As it was not documented in release notes https://github.com/iputils/iputils/releases/tag/s20190324, people complained (issue #236).

562e0d5 ("ping: allow -i and -W option arguments contain garbage input") restored 3) and 4) support (everything including and after first non-numeric character is dropped, warning printed).

### integer number options fail when passed real number (previously decimal part was dropped)
[...]

PR #306, opened 2021-01-13
ping: Restore -W 0 as infinite timeout

pevik:
Fixes: 918e824 ("ping: add support for sub-second timeouts")

References: #290

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -685,7 +685,8 @@ xml:id="man.ping">
           otherwise
           <command>ping</command> waits for two RTTs.
           Real number allowed with dot as a decimal separator
-          (regardless locale setup).</para>
+          (regardless locale setup).
+          0 means infinite timeout.</para>
         </listitem>
       </varlistentry>
     </variablelist>
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -488,7 +488,7 @@ main(int argc, char **argv)
 			double optval;
 
 			optval = ping_strtod(optarg, _("bad linger time"));
-			if (isless(optval, 0.001) || isgreater(optval, (double)INT_MAX / 1000))
+			if (isless(optval, 0) || isgreater(optval, (double)INT_MAX / 1000))
 				error(2, 0, _("bad linger time: %s"), optarg);
 			/* lingertime will be converted to usec later */
 			rts.lingertime = (int)(optval * 1000);


pevik:
@kerolasa @MofX @nmeyerhans any objections to this?

nmeyerhans:
> @kerolasa @MofX @nmeyerhans any objections to this?

0 as an infinite linger time is sensible, especially with the updates to the documentation.

The rest of #290 remains rather messy, but I think this change can happen regardless of how we handle the rest of the issues.

pevik:
@nmeyerhans thanks for your review. Actually most of https://github.com/iputils/iputils/issues/290 has been addressed in last release, the only missing item is:
* make wrong user input on ping -i -W fatal (revert 562e0d5). It'd be good to further improve error message on wrong input, but that can wait till future.
(see https://github.com/iputils/iputils/issues/290#issuecomment-675551087)
I'm thinking about doing release this month, we can discuss whether remove warning now or still keep it.

MofX:
Change looks good to me, although I don't understand linger time or the implementation does not do what I would expect

It looks like it is only used for the first packet, because for the next call `__schedule_exit` will have `waittime != 0` and just return next.
That would also mean, the "two RTTs" part will not work as documented, because the RTT is only available after the first echo was received, but if no lingertime is given, default lingertime `MAXWAIT = 10` is used.

pevik:
> Change looks good to me, although I don't understand linger time or the implementation does not do what I would expect
> 
> It looks like it is only used for the first packet, because for the next call `__schedule_exit` will have `waittime != 0` and just return `next`.

Yes, `setitimer()` is called only once, unless `waittime == 0`.

> That would also mean, the "two RTTs" part will not work as documented, because the RTT is only available after the first echo was received, but if no lingertime is given, default lingertime `MAXWAIT = 10` is used and `waittime` is already set to `10000` in the first call to `__schedule_exit`.

I'll have to look more deeply into the implementation. But I guess the change is not relevant to your question (might be a separate bug).

MofX:
> I'll have to look more deeply into the implementation. But I guess the change is not relevant to your question (might be a separate bug).

You are right, that's why I wrote, that the changes look good to me. I think this can be merged.

pevik:
Investigate RTT is my TODO.

PR #310, opened 2021-01-25
tftpd: install into sbindir

floppym:
The xinet.d config expects the daemon to live in sbin.


floppym:
@pevik Could you merge this one before the next release as well?

pevik:
`Reviewed-by: Petr Vorel <pvorel@suse.cz>`.
Sure, good catch. Thanks!

PR #313, opened 2021-02-22
Limit setcap search to PATH for cross builds

jatedev:
For cross compile builds, set the argument for find_program to allow searching
the PATH only.

The hard coded paths in find_program mean that the search for setcap extends
beyond the cross build environment meaning possible host contamination.

Signed-off-by: Jate Sujjavanich <jatedev@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -215,7 +215,14 @@ config_h = configure_file(
 	output : 'config.h',
 	configuration : conf)
 
-setcap = find_program('setcap', '/usr/sbin/setcap', '/sbin/setcap', required : false)
+if meson.is_cross_build()
+	message('limit setcap search to PATH')
+	path_setcap = ['setcap']
+else
+	message('search for setcap in PATH and admin dirs')
+	path_setcap = ['setcap', '/usr/sbin/setcap', '/sbin/setcap']
+endif
+setcap = find_program(path_setcap, required : false)
 if cap_dep.found() and setcap.found()
 	perm_type = 'caps'
 	setcap_path = setcap.path()


pevik:
I thought that `setcap` is being run on the host, thus should be ok to use system dirs.
@kerolasa @ffontaine @nmeyerhans can you please have a look?

jatedev:
In the case of openembedded, the cross build tightly controls the tools available, so searching the system path results in non-determinism in the build.

I wanted to see if this makes sense for upstream.

pevik:
Non-determinism? I still don't understand the problem. IMHO host's `setcap` is being used in cross-compilation, right?
I haven't checked the implementation but IMHO it first search in the `$PATH` and I guess the other `sbin` dirs are for these who don't have them in `$PATH`.

jatedev:
In openembedded, the tools available to iputils during the build are tightly controlled by limiting the PATH. It excludes /usr/sbin, for example. So we excluded setcap from PATH for the cross build, but find_program finds it via the absolute paths in arguments two and three. This means part of the build can be determined by the build machine outside of the build environment.

We might be able to work around this by using the configuration of libcap via the test cap_dep.found(), I believe.
> if cap_dep.found() and setcap.found()
> 	perm_type = 'caps'
> 	setcap_path = setcap.path()
> else
> 	perm_type = 'setuid'
> 	setcap_path = '/dev/null'
> endif
>

nmeyerhans:
When the binaries are being installed on the same system where they were built, then I think the current logic around setcap is fine. However, where the target is a package management system (i.e. Debian, OpenWRT, and presumably OpenEmbedded), the decision about what sort of permissions to set and how to set them should typically be left to the package manager. Whatever environment is present on the build host, and whatever options it has available for setting the permissions/capabilities/etc, may differ from the target system.

When building iputils packages for Debian and OpenWRT, I set `-DNO_SETCAP_OR_SUID` in the build. If it's appropriate to set capabilities or permissions for the target system, that happens during the installation. I think Open Embedded should probably do something similar.

The current logic in meson.build is fine, in my opinion. My only suggestion would be to consider switching the default, so no binaries are installed with extra capabilities unless explicitly requested.

pevik:
> When building iputils packages for Debian and OpenWRT, I set `-DNO_SETCAP_OR_SUID` in the build. If it's appropriate to set capabilities or permissions for the target system, that happens during the installation. I think Open Embedded should probably do something similar.
+1

> The current logic in meson.build is fine, in my opinion. My only suggestion would be to consider switching the default, so no binaries are installed with extra capabilities unless explicitly requested.

Make sense, I'll prepare PR.

pevik:
`NO_SETCAP_OR_SUID` is default false now (9cec7363bb163ae6c90eb088e242a4639f4aec24), thus closing this.

Issue #314, opened 2021-02-27
Socket :operation not permitted


xxllp:
systemd rdisc service start error 
log: Socket :operation not permitted
and when delete PrivateUsers=yes  in rdisc.service  the service is ok 
but why?

pevik:
I wonder if this was caused by missing `CAP_NET_ADMIN` capability, which was added in https://github.com/iputils/iputils/commit/710a161770aabd13e4f5028c325c41f63a417e4c. Probably not as the error looks different.

PR #316, opened 2021-03-03
rdisc: remove PrivateUsers=yes from systemd service file

kerolasa:
Quoting systemd.exec(5) manual page 'Specifically this means that the
process will have zero process capabilities on the host's user namespace'.
That does not combine will with CAP_NET_RAW that needs to take effect host's
namespace.

Secondlly add CapabilityBoundingSet that is will ensure capabilities are
limited to the one and only capability it needs.

Fixes: https://github.com/iputils/iputils/issues/314
Reference: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateUsers=


pevik:
LGTM.
nit: s/Secondlly/Secondly/g in the commit message.

Issue #320, opened 2021-03-26
Reply payload and truncation checking is wrong and bypasses gather_statistics() when it should not


paul-demo:
All past versions of ping on OSX and Ubuntu have allowed for truncated responses. For example, google.com truncates to 96B (68B ICMP payload + 8B ICMP header + 20B IPv4 header).

However, we found that moving from Ubuntu 18 to Ubuntu 20, ping has changed, and now doesn't report the info message "truncated...got X bytes instead of Y", but rather just doesn't show anything when a reply is truncated.

Why did you change this behavior?

paul-demo:
Tracing the execution manually:

https://github.com/iputils/iputils/blob/master/ping/ping.c#L1554
```c++
if (gather_statistics(rts, (uint8_t *)icp, sizeof(*icp), cc,
				      ntohs(icp->un.echo.sequence),
				      reply_ttl, 0, tv, pr_addr(rts, from, sizeof *from),
				      pr_echo_reply, rts->multicast)) {
			fflush(stdout);
			return 0;
		}
```

The 7th argument is "0".

This goes into gather_statistics:
https://github.com/iputils/iputils/blob/665782e1d3c77df2c90f144b586da1ac2da58302/ping_common.c#L763

```c++
int gather_statistics(__u8 *icmph, int icmplen,
		      int cc, __u16 seq, int hops,
		      int csfailed, struct timeval *tv, char *from,
		      void (*pr_reply)(__u8 *icmph, int cc))
```

Let's say timing = 1, then sizeof(struct timeval) would maybe be 0 since timeval *tv is passed the literal value 0 which I guess means NULL. So then it's possible that ping gets stuck in an infinite loop here:

```
	if (timing && cc >= 8+sizeof(struct timeval)) {
		struct timeval tmp_tv;
		memcpy(&tmp_tv, ptr, sizeof(tmp_tv));

restamp:
		tvsub(tv, &tmp_tv);
		triptime = tv->tv_sec * 1000000 + tv->tv_usec;
		if (triptime < 0) {
			fprintf(stderr, "Warning: time of day goes back (%ldus), taking countermeasures.\n", triptime);
			triptime = 0;
			if (!(options & F_LATENCY)) {
				gettimeofday(tv, NULL);
				options |= F_LATENCY;
				goto restamp;
			}
		}
```

This is just a guess that it is getting stuck in an infinite loop in this "goto restamp". Maybe the authors of the code could comment.

paul-demo:
[...]

PR #321, opened 2021-03-27
Roll back f7710a1 which checks payloads incorrectly, ignores all mism…

paul-demo:
…atch replies

- gather_statistics() already checks payloads and reports truncated packets
- f7710a1 skips all payload comparisons in gather_statistics() when it should not
- f7710a1 ignores truncated packets when it should not
- f7710a1 broke ping's historical behavior that reply payloads are checked
  against request payloads and differences (including reply truncation) are
  reported by gather_statistics(), because the change prevents execution of
  gather_statistics() entirely unless payload match is complete and the reply
  is not truncated

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1549,8 +1549,6 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 			return 1;
 		if (!is_ours(rts, sock, icp->un.echo.id))
 			return 1;			/* 'Twas not our ECHO */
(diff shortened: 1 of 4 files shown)


pevik:
f7710a17c4d5994313a64583f511bcdb9559f2a9 was reverted in dff5d82dadab1b04400b2f9e1eb10a0d124868ed (your PR https://github.com/iputils/iputils/issues/320), thus closing.

PR #328, opened 2021-04-29
Add RFC 8335 (Probe) Request and Response Functionality

slang21:
[RFC 8335](https://tools.ietf.org/html/rfc8335) describes the use of extended echo requests to probe specific network interfaces. This PR adds the -e flag to the ping command to probe specific interfaces and receive responses from those interfaces. Functionality for RFC 8335 was recently included in Linux net-next.

Usage:

ping -4 -e [interface] [address]

Interface can be specified by name, index, or address.

Example:

ping -4 -e 1 192.168.1.201
PING 192.168.1.201 (192.168.1.201) 56(84) bytes of data.
64 bytes from 192.168.1.201: icmp_seq=263 ttl=64
64 bytes from 192.168.1.201: icmp_seq=519 ttl=64
64 bytes from 192.168.1.201: icmp_seq=775 ttl=64
--- 192.168.1.201 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -64,21 +64,37 @@ struct ping_rts *global_rts;
 #ifndef ICMP_FILTER
 #define ICMP_FILTER	1
 struct icmp_filter {
-	uint32_t	data;
+	uint64_t	data;
 };
 #endif
+
+
+struct exthdr {
+    uint16_t    v_rsvd;
+    uint16_t    checksum;
+};
+
+struct iiohdr {
+    uint16_t    len;
+    uint8_t     class;
+    uint8_t     ctype;
+};
 
 ping_func_set_st ping4_func_set = {
 	.send_probe = ping4_send_probe,
 	.receive_error_msg = ping4_receive_error_msg,
 	.parse_reply = ping4_parse_reply,
-	.install_filter = ping4_install_filter
+	.install_filter = ping4_install_filter,
+	.install_probe_filter = probe4_install_filter,
 };
 
 #define	MAXIPLEN	60
 #define	MAXICMPLEN	76
 #define	NROUTES		9		/* number of record route slots */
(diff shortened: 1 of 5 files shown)


ping/ping.c
```diff
@@ -1703,3 +2059,36 @@ void ping4_install_filter(struct ping_rts *rts, socket_st *sock)
...
+
+	// if (once)
+	// 	return;
+	// once = 1;
+
+	// /* Patch bpflet for current identifier. */
+	// insns[2] = (struct sock_filter)BPF_JUMP(BPF_JMP | BPF_JEQ | BPF_K, htons(rts->ident), 0, 1);
+
+	// if (setsockopt(sock->fd, SOL_SOCKET, SO_ATTACH_FILTER, &filter, sizeof(filter)))
+	// 	error(0, errno, _("WARNING: failed to install socket filter"));
```
nmeyerhans, line 2092: This doesn't seem necessary.

nmeyerhans (commented):
At first glance it looks OK.

One question: When I run this code and inspect the transmitted packets using wireshark, I see the expected interface identification object, but then another unknown object:

```
ICMP Multi-Part Extensions
    Version: 2
    .... 0000 0011 1000 = Reserved: 0x038
    [Checksum: [missing]]
    [Checksum Status: Not present]
    Interface Identification Object
        Length: 8
        Class: Interface Identification Object (3)
        C-Type: 2 (Identifies Interface By Index)
        Interface Index: 2
    Unknown object (139/96) (truncated)
        Length: 50720
        Class: Unknown (139)
        C-Type: 96
        Data: 0000000071d10200000000001c1d1e1f202122232425262728292a2b2c2d2e2f30313233…
```

What is the second object? Is it expected?

PR #331, opened 2021-05-14
Revert "Add strict pattern matching on response when pattern was provided"

pevik:
This reverts commit f7710a17c4d5994313a64583f511bcdb9559f2a9.

Commit broke report of truncated packets:
```
$ ping -c2 -s100 google.com
PING google.com (142.250.185.238) 100(128) bytes of data.
```

Running ping from both s20161105 (which does not contain f7710a1) and
reverted f7710a1 on master reports truncated packets:

```
$ ping -c2 -s100 google.com
PING google.com (142.250.185.238) 100(128) bytes of data.
76 bytes from fra16s53-in-f14.1e100.net (142.250.185.238): icmp_seq=1 ttl=116 (truncated)
76 bytes from fra16s53-in-f14.1e100.net (142.250.185.238): icmp_seq=2 ttl=116 (truncated)
```
"truncated" statistics never happend and
there was unreachable code in gather_statistics() because contains_pattern_in_payload() added in
f7710a1 always found a mismatch first and nothing got printed.

There was unreachable code in gather_statistics() because
contains_pattern_in_payload() added in f7710a1 always found a mismatch
first. Due that all of these did not work:
* updating counters for statistics generation
* keeping track of timestamps and time-of-flight using the first section
  of the payload
* checking for duplicate replies and report them
* printing basic info about the reply
* printing "(truncated)" if the reply was truncated
* checking the checksum
* validating the rest of the payload (bytes 17 and above) against the
  ICMP request that was sent, and report any differences

Fixes: f7710a1 ("Add strict pattern matching on response when pattern was provided")
Closes: https://github.com/iputils/iputils/issues/320

Reported-by: paul-demo

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1549,8 +1549,6 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 			return 1;
 		if (!is_ours(rts, sock, icp->un.echo.id))
 			return 1;			/* 'Twas not our ECHO */
-		if (!contains_pattern_in_payload(rts, (uint8_t *)(icp + 1)))
-			return 1;			/* 'Twas really not our ECHO */
 		if (gather_statistics(rts, (uint8_t *)icp, sizeof(*icp), cc,
 				      ntohs(icp->un.echo.sequence),
 				      reply_ttl, 0, tv, pr_addr(rts, from, sizeof *from),
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -380,7 +380,6 @@ int is_ours(struct ping_rts *rts, socket_st *sock, uint16_t id);
 extern int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock);
 extern void sock_setbufs(struct ping_rts *rts, socket_st *, int alloc);
 extern void setup(struct ping_rts *rts, socket_st *);
-extern int contains_pattern_in_payload(struct ping_rts *rts, uint8_t *ptr);
 extern int main_loop(struct ping_rts *rts, ping_func_set_st *fset, socket_st*,
 		     uint8_t *packet, int packlen);
 extern int finish(struct ping_rts *rts);
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -824,8 +824,6 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
 			return 1;
 		if (!is_ours(rts, sock, icmph->icmp6_id))
 			return 1;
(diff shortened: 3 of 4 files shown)


pevik:
@pavlix @kerolasa @jsynacek @nmeyerhans @vapier @nmav @paul-demo Feel free to review.
@paul-demo could you reveal your name and email address for `Reported-by:` tag?

paul-demo:
> @pavlix @kerolasa @jsynacek @nmeyerhans @vapier @nmav @paul-demo Feel free to review.
> @paul-demo could you reveal your name and email address for `Reported-by:` tag?

Yes: Paul Swirhun, paulswirhun@gmail.com

I'd like one of the owners to take responsibility for checking this though; I don't want or need any credit I just want it fixed!

Thanks,
Paul

nmeyerhans (approved):
This looks good to me. The behavior when processing truncated packets is restored.

We might consider whether we want to report the expected packet length in the output when we log a truncated packet, but that doesn't need to happen here.

I am curious about the issue that #67 was trying to fix in the first place. I wonder if it was an actual observed problem or something theoretical. I'd love to see a repro case for it.

pevik:
> This looks good to me. The behavior when processing truncated packets is restored.

Thanks for a review!

> We might consider whether we want to report the expected packet length in the output when we log a truncated packet, but that doesn't need to happen here.

+1, yep, I'd add it as a separate PR.
 
> I am curious about the issue that #67 was trying to fix in the first place. I wonder if it was an actual observed problem or something theoretical. I'd love to see a repro case for it.

That'd be great to have. @locasity care to post more info or even a reproducer.

Sunjeet:
Thanks for looping me in (@locasity here)

IIRC the intention of the pattern matching code was to only be called if requested using the "strict mode" flag. 

This was in response to a production issue. The production code was a perl script shelling out to run ping and traceroute concurrently at a large scale. At that scale we were seeing traceroute responses being misconstrued for ping responses once every few days (due to pid warp around), leading to a "hung" ping process.

Repro instructions from when I reported it in 2016 are here- https://bugs.launchpad.net/ubuntu/+source/iputils/+bug/1551020

pevik:
@Sunjeet thanks for info. I tested your reproducer mentioned in https://bugs.launchpad.net/ubuntu/+source/iputils/+bug/1551020, but it's obviously problematic to some specific host you didn't mention. Thus I decide to revert your implementation to get back truncate packets. Please, if the problem you tried to solve persists, report it again with host, thus we can reproduce it and write test for it.

Sunjeet:
@pevik sounds good to me

PR #333, opened 2021-05-17
Fix variance estimate

jack-champagne:
Should fix issue iputils#332 . For more on why this change, this wikipedia article is good: https://en.wikipedia.org/wiki/Bessel%27s_correction

Added a check to ensure more than 1 pings happened for variance to be calculated. Otherwise, there will be no variance.


jack-champagne:
Sorry this was a duplicate

pevik:
> Sorry this was a duplicate

Please next time just use `git push -f` to overwrite your code instead opening new PR.

PR #334, opened 2021-05-18
Fix Variance Estimate

### LINKED ISSUE #332: Standard Deviation not calculated Correctly
Using ping, the statistic mdev given on completion is inaccurate as an estimate for standard deviation.

As an example:
```
 ping dlford.io
PING dlford.io (174.126.87.71) 56(84) bytes of data.
64 bytes from 174-126-87-71.cpe.sparklight.net (174.126.87.71): icmp_seq=1 ttl=54 time=81.7 ms
64 bytes from 174-126-87-71.cpe.sparklight.net (174.126.87.71): icmp_seq=2 ttl=54 time=90.9 ms
64 bytes from 174-126-87-71.cpe.sparklight.net (174.126.87.71): icmp_seq=3 ttl=54 time=81.3 ms
64 bytes from 174-126-87-71.cpe.sparklight.net (174.126.87.71): icmp_seq=4 ttl=54 time=82.8 ms
64 bytes from 174-126-87-71.cpe.sparklight.net (174.126.87.71): icmp_seq=5 ttl=54 time=83.8 ms
^C
--- dlford.io ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4002ms
rtt min/avg/max/mdev = 81.274/84.104/90.949/3.537 ms
```

The standard deviation for this sample of pings should actually be closer to 3.92 ms (if I were to for example put in the travel times into excel and get the std. dev.)

jack-champagne:
Should fix issue #332 . For more on why this change, this wikipedia article is good: https://en.wikipedia.org/wiki/Bessel%27s_correction

Added a check to ensure more than 1 pings happened for variance to be calculated. Otherwise, there will be no variance.

Closes #332

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -903,16 +903,16 @@ int finish(struct ping_rts *rts)
 		long long tmvar;
 
 		// Add check for more than one to calculate std. dev. Will be zero otherwise.
-		if (total <= 1)
+		if (total <= 1) {
 			tmvar = 0;
-		else
+		} else {
 			if (rts->tsum < INT_MAX)
 				/* This slightly clumsy computation order is important to avoid
 			 	* integer rounding errors for small ping times. */
 				tmvar = (rts->tsum2 - ((rts->tsum * rts->tsum) / total)) / (total - 1);
 			else
-				tmvar = (rts->tsum2 / (total - 1)) - (tmavg * tmavg)*(total / (total - 1);
-
+				tmvar = (rts->tsum2 / (total - 1)) - (tmavg * tmavg) * (total / (total - 1));
+		}
 		tmdev = llsqrt(tmvar);
 
 		printf(_("rtt min/avg/max/mdev = %ld.%03ld/%lu.%03ld/%ld.%03ld/%ld.%03ld ms"),
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -902,12 +902,16 @@ int finish(struct ping_rts *rts)
 		long tmavg = rts->tsum / total;
 		long long tmvar;
 
-		if (rts->tsum < INT_MAX)
-			/* This slightly clumsy computation order is important to avoid
-			 * integer rounding errors for small ping times. */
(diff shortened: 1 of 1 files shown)


ping/ping_common.c
```diff
@@ -902,13 +902,17 @@ int finish(struct ping_rts *rts)
...
+		if (total <= 1) {
+			tmvar = 0;
+		} else {
+			if (rts->tsum < INT_MAX)
+				/* This slightly clumsy computation order is important to avoid
+			 	* integer rounding errors for small ping times. */
+				tmvar = (rts->tsum2 - ((rts->tsum * rts->tsum) / total)) / (total - 1);
+			else
+				tmvar = (rts->tsum2 / (total - 1)) - (tmavg * tmavg) * (total / (total - 1));
+		}
```
pevik, line 915: https://github.com/iputils/iputils/issues/332#issuecomment-848020207
> I'd be very hesitant to change the meaning of the existing fields, since they're well established at this point.  However, collecting and/or reporting additional statistics might be interesting.  It might be preferable to report the additional information only if explicitly requested at runtime (via a flag, etc), but I haven't thought too deeply about it.

I'd also be very careful to change this.


Just commenting style - reducing else part makes code more readable:
```
long long tmvar = 0;
...
if (total > 1) {
	if (rts->tsum < INT_MAX)
		/* This slightly clumsy computation order is important to avoid
		 * integer rounding errors for small ping times. */
		tmvar = (rts->tsum2 - ((rts->tsum * rts->tsum) / total)) / (total - 1);
	else
		tmvar = (rts->tsum2 / (total - 1)) - (tmavg * tmavg) * (total / (total - 1));
}
```
Also I'd just quash the commits into one. You also didn't put your `Signed-off-by: Your Name <your@email>`, see other commits.

jack-champagne: Thanks for you comments @pevik, this PR was from a long time ago before I knew what I know now, I honestly forgot about it. I presume since I deleted my fork, it notified you. Thank you for your comments however!

Issue #243, opened 2019-12-25
Implement tests to cover basic functionality


pevik:
Write posix shell test framework to cover basic functionality. For start I'd cover `ping` or `arping`, which are IMHO mostly used.

kerolasa:
I am not sure what is meant with shell test framework. Taking what meson has to offer is likely the fitting way to add tests to this project.

https://mesonbuild.com/Unit-tests.html

MikeHigginbottom:
I'd like to take this issue. Not sure that Meson's unit test support will be enough on its own though. It seems to assume you have some kind of test framework already in place (e.g. CUnit) in which the test executables are written. Meson would then call out to these exes as part of the build process. Would a combination of the two be appropriate?

pevik:
@MikeHigginbottom sure, go ahead :). Unfortunately I'm too busy atm to work on it.

BTW my original idea was to have functional testing (e.g. all options work as expected; because that's what affect users). IMHO code is quite small (and it's not a library) therefore I don't see a priority for unit testing. But maybe I'm wrong.

For functional testing I planned to start with `ping` and `arping`. I intend to reuse shell part of LTP test framework (https://github.com/linux-test-project/ltp/blob/master/testcases/lib/tst_test.sh it would require to take few parts of C code from C API as well), with it's network extension (https://github.com/linux-test-project/ltp/blob/master/testcases/lib/tst_net.sh). This framework is well established, network namespaces testing requires zero setup from user (except root privileges for setting netns).
Here is some general info about LTP shell API https://github.com/linux-test-project/ltp/wiki/Test-Writing-Guidelines#23-writing-a-testcase-in-shell (unfortunately network extension hasn't been documented yet).

These are just my thoughts, which I planned to implement. But feel free to do something completely different.

pevik:
We should also test code on kernel without IPv6 support. Although I expect IPv6 everywhere, there are obviously people running IPv4 only: https://github.com/iputils/iputils/issues/293

I don't have much time for a development, but consider tests as a high priority. Hope I'll find time to prepare subset of LTP shell API and write at least some tests soon.

pevik:
I've started working on testsuite. Our testsuite should be probably based on `meson test` command run in `builddir`

[...]

PR #335, opened 2021-05-27
Meson unit tests for ping

pevik:
Start of implementing https://github.com/iputils/iputils/issues/243

These tests are simple meson unit tests for `ping`, which just check expected pass or fail.
This is just a start, because we need
1) cover all programs
2) implementing test framework which allows more complicated tests (e.g. to compare ping output)
I have started to adapt LTP framework (cleanup unimportant things), because for comparing output etc. I'd like use already written framework instead of reinventing a wheel in `meson`.


pevik:
NOTE: errors aren't related to changes in this PR. It's docker pull rate limit. We need to switch to github actions, which hopefully will not be affected. I plan to do it soon.
https://www.docker.com/increase-rate-limit

pevik:
After rebase github does not reflect my changes. I need to open new PR.

PR #336, opened 2021-06-01
CI: Move from Travis to GitHub Actions

pevik:
Before merging other PR we need to fix CI.

Travis is unreliable due "pull rate limit" issue.

Also using GitHub native CI allows more advanced features (e.g. update wiki with doc/*.txt, nightly build docparser doc).

Nice bonus is that manual podman activation for distros using glibc >= 2.33 (e.g. openSUSE Tumbleweed, Fedora) it's not needed in GitHub.

Tested: https://github.com/pevik/iputils/actions/runs/894732214


pevik:
The only disadvantage of GitHub actions is a that it does not allow to rerun individual (failed) job. The only way is to restart all jobs. But it's not a big issue for iputils, as build is quite quick.
https://github.com/actions/runner/issues/432

I'll merge this tomorrow unless any feedback.
@nmeyerhans nice improvement would be to add OpenWrt docker image into testing (as a separate effort).

Issue #339, opened 2021-06-09
ping: No errors if an interface is specified


tashanamatl:
When the ICMP socket is bound to an interface, if trying to ping an
address that is unreachable then no errors will be received, as they come in
through the lo interface. To be able to get useful errors the binding of the icmp socket to the device added in https://github.com/iputils/iputils/commit/f455fee41c077d4b700a473b2f5b3487b8febc1d may need to be removed.

Before that patch:
ping -4 192.168.1.2 -I vlan1000 -c 5
PING 192.168.1.2 (192.168.1.2) from 192.168.1.100 vlan1000: 56(84) bytes of data.
From 192.168.1.100 icmp_seq=1 Destination Host Unreachable
From 192.168.1.100 icmp_seq=2 Destination Host Unreachable
From 192.168.1.100 icmp_seq=3 Destination Host Unreachable
From 192.168.1.100 icmp_seq=4 Destination Host Unreachable
From 192.168.1.100 icmp_seq=5 Destination Host Unreachable

--- 192.168.1.2 ping statistics ---
5 packets transmitted, 0 received, +5 errors, 100% packet loss, time 4179ms

After the patch:
ping -4 192.168.1.2 -I vlan1000 -c 5
PING 192.168.1.2 (192.168.1.2) from 192.168.1.100 vlan1000: 56(84) bytes of data.

--- 192.168.1.2 ping statistics ---
5 packets transmitted, 0 received, 100% packet loss, time 4130ms

pevik:
@tashanamatl any change you specify steps to reproduce the issue? Ideally with using network namespaces, see some examples of netns based reproducers:

https://github.com/iputils/iputils/pull/340#issuecomment-987219698
https://github.com/iputils/iputils/pull/296#issuecomment-710044824
https://github.com/iputils/iputils/issues/371#issuecomment-922582756
https://github.com/iputils/iputils/pull/344
https://github.com/iputils/iputils/issues/300

pevik:
@tashanamatl I guess it's the same issue as https://github.com/iputils/iputils/issues/398? I.e. fixed in 5.17 and stable kernels? (https://github.com/iputils/iputils/pull/405#issuecomment-1146359974)

tashanamatl:
> 
I'm not sure, it looks like this issue is  just IPv6 but the issue I was seeing was only IPv4. I actually think #391 might be related.

pevik:
[...]

PR #341, opened 2021-06-18
Ping: Remove binding of ICMP socket

tashanamatl:
When specifying an interface in the ping command, if trying to ping an
address that is not reachable then no errors would be received. This was
because the ICMP socket was being bound to the source interface but the
errors were coming through the loopback interface.
To fix this the ICMP socket should not be bound to the interface.
Instead IPv4 ping should copy IPv6's use of cmsg. A cmsg header will
contain information about the interface and will then be included in the
msg header when sending a ping. This means the outgoing ping uses the
correct interface but incoming messages from other interfaces can still
be received on the socket.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -75,6 +75,9 @@ ping_func_set_st ping4_func_set = {
 	.install_filter = ping4_install_filter
 };
 
+static unsigned char cmsgbuf[4096];
+static size_t cmsglen = 0;
+
 #define	MAXIPLEN	60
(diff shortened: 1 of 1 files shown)


tashanamatl:
This commit would fix https://github.com/iputils/iputils/issues/339

pevik:
This commit don't even compile.

pevik:
@tashanamatl Does the problem still exists in the latest release and in the current master? If yes, how to reproduce it?

PR #342, opened 2021-06-22
ping: add RFC 8335 PROBE request and response functionality

aroeseler:
This is a continuation of a project with slang21 and an update to the existing pull request found here: https://github.com/iputils/iputils/pull/328

This pull request adds timestamp info to the probe messages, fixes a bug in calculating the ctype of the message, and adds probing by ICMPV6 messages. 

Original pull request blurb:
RFC 8335 describes the use of extended echo requests to probe specific network interfaces. This PR adds the -e flag to the ping command to probe specific interfaces and receive responses from those interfaces. Functionality for RFC 8335 was recently included in Linux net-next.

Usage:

ping -e [interface] [address]

Interface can be specified by name, ifindex, or IPV4 or IPV6 address.

ping -e 1 10.0.0.89
PING 10.0.0.89 (10.0.0.89) 56(84) bytes of data.
Interface: 1
Status: ACTIVE IPV4 IPV6
64 bytes from 10.0.0.89: icmp_seq=1 ttl=64 time=2.30 ms

--- 10.0.0.89 ping statistics ---
rtt min/avg/max/mdev = 2.298/2.298/2.298/0.000 ms

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -64,7 +64,7 @@ struct ping_rts *global_rts;
 #ifndef ICMP_FILTER
 #define ICMP_FILTER	1
 struct icmp_filter {
-	uint32_t	data;
+	uint64_t	data;
 };
 #endif
 
@@ -311,7 +311,7 @@ main(int argc, char **argv)
 		hints.ai_family = AF_INET6;
 
 	/* Parse command line options */
-	while ((ch = getopt(argc, argv, "h?" "4bRT:" "6F:N:" "aABc:dDfi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:")) != EOF) {
+	while ((ch = getopt(argc, argv, "h?" "4bRT:" "6F:N:" "aABc:dDfi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:e:")) != EOF) {
 		switch(ch) {
 		/* IPv4 specific options */
 		case '4':
@@ -494,6 +494,12 @@ main(int argc, char **argv)
 			rts.lingertime = (int)(optval * 1000);
 		}
 			break;
+		case 'e':
+		{
+			rts.probe = 1;
+			rts.interface = optarg;
+		break;
+		}
 		default:
 			usage();
 			break;
@@ -887,7 +893,6 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 			tmp_rspace = (uint32_t *)&rspace[4 + i * 4];
 			*tmp_rspace = rts->route[i];
 		}
-
 		if (setsockopt(sock->fd, IPPROTO_IP, IP_OPTIONS, rspace, 4 + rts->nroute * 4) < 0)
 			error(2, errno, "record route");
 		rts->optlen = 40;
@@ -1435,16 +1440,21 @@ in_cksum(const unsigned short *addr, int len, unsigned short csum)
 int ping4_send_probe(struct ping_rts *rts, socket_st *sock, void *packet,
 		     unsigned packet_size __attribute__((__unused__)))
 {
-	struct icmphdr *icp;
+	struct exthdr   *extbase, ext;
+	struct iiohdr   *iiobase, iio;
+	uint32_t iio_ip_hdr = 0;
+	struct icmphdr  *icp;
 	int cc;
 	int i;
 
 	icp = (struct icmphdr *)packet;
-	icp->type = ICMP_ECHO;
 	icp->code = 0;
 	icp->checksum = 0;
-	icp->un.echo.sequence = htons(rts->ntransmitted + 1);
 	icp->un.echo.id = rts->ident;			/* ID */
+	if (rts->probe == 1)
+		goto build_probe;
+	icp->type = ICMP_ECHO;
+	icp->un.echo.sequence = htons(rts->ntransmitted + 1);
 
 	rcvd_clear(rts, rts->ntransmitted + 1);
 
@@ -1458,6 +1468,7 @@ int ping4_send_probe(struct ping_rts *rts, socket_st *sock, void *packet,
 		}
 	}
 
+send_msg:
 	cc = rts->datalen + 8;			/* skips ICMP portion */
 
 	/* compute ICMP checksum here */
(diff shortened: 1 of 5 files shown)


ping/ping.c
```diff
@@ -1466,13 +1477,82 @@ int ping4_send_probe(struct ping_rts *rts, socket_st *sock, void *packet,
...
+build_probe:
+	extbase = (struct exthdr *)(icp + 1);
+	iiobase = (struct iiohdr *)((char *)extbase + sizeof(struct exthdr));
+	icp->type = ICMP_EXT_ECHO;
+	/* PROBE messages use only the first 8 bits as sequence number */
+	icp->un.echo.sequence = htons((rts->ntransmitted + 1) << 8);
+	icp->un.echo.sequence |= htons(1);	/* Set L-bit */
+	WRITE_VERSION(ext.v_rsvd , 2);
+	ext.v_rsvd = htons(ext.v_rsvd);
+	ext.checksum = 0;
```
fenner, line 1500: You must fill in the checksum later. While it appears that the Linux kernel does not check the checksum, there are other probe implementations that do.

My tcpdump update https://github.com/the-tcpdump-group/tcpdump/pull/1115 will check that it's filled in properly.

ping/ping_common.c
```diff
@@ -293,6 +294,48 @@ void print_timestamp(struct ping_rts *rts)
...
+/* Helper functions for constructing RFC 8335 PROBE messages */
+/* Verifies name is a valid linux interface name
+ * Copied from iproute2's if name check
+ */
+int check_ifname(const char *name)
+{
+	/* These checks mimic kernel checks in dev_valid_name */
+	if (*name == '\0')
+		return -1;
+	if (strlen(name) >= IFNAMSIZ)
```
fenner, line 306: This limit only exists in Linux.  Imagine running this against a router with an interface name TwentyFiveGigabitEthernet3/0/1

ping/ping_common.c
```diff
@@ -293,6 +294,48 @@ void print_timestamp(struct ping_rts *rts)
...
+		return -1;
+
+	while (*name) {
+		if (*name == '/' || isspace(*name))
```
fenner, line 310: As mentioned above, slashes are common in interface names.  RFC8335 only limits interface names by referring to RFC7223, which says that they are yang strings.

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
 		}
+	} else if (icmph->icmp6_type == ICMPV6_EXT_ECHO_REPLY) {
+		if (!rts->multicast &&
+		    memcmp(&from->sin6_addr.s6_addr, &rts->whereto6.sin6_addr.s6_addr, 16))
+			return 1;
+		if (!is_ours(rts, sock, icmph->icmp6_id))
+			return 1;
+
+		sequence = ntohs(icmph->icmp6_seq);
+		state = icmph->icmp6_seq & 0xe0;
```
fenner, line 943: You forgot to shift this right by 5, so the switch below won't hit any of the cases.

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
+			case 2:
+				printf("Error: No Such Interface\n");
+				break;
+			case 3:
+				printf("Error: No Such Table Entry\n");
+				break;
+			case 4:
+				printf("Error: Multiple Interfaces Satisfy Query\n");
+				break;
+			default:
```
fenner, line 958: An invalid code (5 or above) will result in no error, and no output from the "code == 0" case below.  This will likely be fairly confusing.  Sure, it shouldn't happen, but, let's try to confuse the user less if it does occur.

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
+				break;
+		}
+		if (icmph->icmp6_code == 0) {
+			if ((sequence & ICMP_EXT_ECHOREPLY_ACTIVE) != 0) {
+				printf("Status: ACTIVE");
+				if (sequence & ICMP_EXT_ECHOREPLY_IPV4)
+					printf(" IPV4");
+				if (sequence & ICMP_EXT_ECHOREPLY_IPV6)
+					printf(" IPV6");
```
fenner, line 989: There's an unusual situation when the status is active, but there is no v4 or v6 configured.  I suggest printing out something specific when `(!sequence & (ICMP_EXT_ECHOREPLY_IPV4| ICMP_EXT_ECHOREPLY_IPV6))`.

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
+			printf("\n");
+			}
```
fenner, line 991: The else case here is significant: if you sent a query about an interface, and A=0, that is something the user probably really wants to know.  Maybe
```
 } else {
   printf("Status: INACTIVE\n");
 }
```

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
+			case 3:
+				printf("Error: No Such Table Entry\n");
+				break;
+			case 4:
+				printf("Error: Multiple Interfaces Satisfy Query\n");
+				break;
+			default:
+				break;
+		}
+		switch (state) {
```
fenner, line 961: The returned state value is only valid if code == 0 and the transmitted L flag is 0; I recommend checking that here.

ping/ping6_common.c
```diff
@@ -832,6 +932,71 @@ int ping6_parse_reply(struct ping_rts *rts, socket_st *sock,
...
+			case 5:
+				printf("State: Probe\n");
+				break;
+			case 6:
+				printf("State: Failed\n");
+				break;
+			default:
+				break;
+		}
+		if (icmph->icmp6_code == 0) {
```
fenner, line 983: The returned A/4/6 values are only valid if code=0 and L=1, I recommend checking that here.

ping/ping6_common.c
```diff
@@ -571,6 +577,90 @@ int build_echo(struct ping_rts *rts, uint8_t *_icmph,
...
+	icmph = (struct icmp6_hdr *)_icmph;
+	extbase = (struct exthdr *)(icmph + 1);
+	iiobase = (struct iiohdr *)((char *)extbase + sizeof(struct exthdr));
+	icmph->icmp6_type = ICMPV6_EXT_ECHO_REQUEST;
+	icmph->icmp6_code = 0;
+	icmph->icmp6_cksum = 0;
+	icmph->icmp6_id = rts->ident;
+	/* PROBE messages use only the first 8 bits as sequence number */
+	icmph->icmp6_dataun.icmp6_un_data8[2] = rts->ntransmitted + 1;
+	icmph->icmp6_dataun.icmp6_un_data8[3] = 1;	/* Set L-bit */
```
fenner, line 598: I recommend giving the user an argument to control this (e.g., "--remote") - this is one of the most important pieces of PROBE functionality.  In the case where the L-bit is clear, the only valid ctype is ICMP_EXT_ECHO_CTYPE_ADDR, but otherwise all the rest of the code applies.

fenner (changes requested):
Thanks for the great start on PROBE support!  While I have several comments, I believe all of them are minor.

I have two different test resources to help with testing these changes:
1. A synthetic probe responder, which responds with a different response to each request, then a timeout, then starts cycling around from the beginning again.  This way, you don't have to manually create a situation that invokes a given response; just run against an instance of this responder.  It's a standalone script written in python, that uses the scapy packet library.  It's available from https://github.com/aristanetworks/probe-tools
2. tcpdump patches that print PROBE requests/replies, including checking the checksum, etc.

fenner:
Oh, one more thought here: when running the client on a system without the necessary kernel updates (pre-5.14, afaik), the `send()` returns with `EINVAL` and the request never gets sent, but there is no indication of a problem.

```
./ping -e 1 -c 5 127.0.0.1

5 packets transmitted, 0 received, 100% packet loss, time 4094ms
```

`100% packet loss` is probably not the best way to represent "your system can not send this type of packet".

strace says:
```
sendto(3, "*\0\353\16\0\0\5\1  \0\0\0\10\3\2\0\0\0\1\261\321\256e\0\0\0\0\"\310\0\0"..., 64, 0, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("127.0.0.1")}, 16) = -1 EINVAL (Invalid argument)
```

fenner:
I've created issues for all of my comments at https://github.com/aristanetworks/iputils/issues and intend to work on addressing them.

See https://github.com/aristanetworks/iputils/wiki for more information.

PR #345, opened 2021-07-01
ping: connmark is unsigned int, not signed

luizluca:
See: https://github.com/torvalds/linux/blob/master/include/uapi/linux/netfilter/xt_connmark.h

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -420,7 +420,7 @@ main(int argc, char **argv)
 			rts.opt_noloop = 1;
 			break;
 		case 'm':
-			rts.mark = strtol_or_err(optarg, _("invalid argument"), 0, INT_MAX);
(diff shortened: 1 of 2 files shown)


nmeyerhans (approved):
LGTM

luizluca:
Rebased and signed the commit

pevik:
Well, `SO_MARK` is `__u32 sk_mark`, i.e. `uint32_t` in userspace. I'm not sure if `unsigned int` is 4 bytes on all platforms.
And therefore not sure if `UINT32_MAX` is always the same on all platforms, thus maybe check for 4294967295 (the value of `UINT32_MAX` on x86_64).

metan-ucw:
Agree with @luizluca here I do not think that we have to worry about UINT_MAX not being 32bit on Linux. So this can go in as it is and the code could be cleaned up to use the stdint.h types later on.

PR #346, opened 2021-07-01
systemd unit: rdisc: Fixes of capabilities and type

6eh01der:
-t option is for testing purposes, so it's unknown for what reason it was added. So it much more nicer to replace it with -v option. -v is for detailed output verbosity. Additionally rdisc requires CAP_NET_ADMIN capability for ability to modify routing table. Without it there will be error: "ioctl (add/delete route): Operation not permitted"
Also service type must be set to "forking". Without it service will not start.


6eh01der:
I will create new request without debian specific files

pevik:
> I will create new request without debian specific files

@6eh01der FYI You may want to discuss Debian specific packaging with @nmeyerhans (Debian iputils maintainer).

pevik:
@6eh01der FYI Debian itself (as most of distros, if not even all distros) does not use `build-aux/setcap-setuid.sh`:
https://salsa.debian.org/debian/iputils/-/blob/master/debian/rules#L23

6eh01der:
@pevik  Yes, setcap-setuid.sh needed only for build from sources and installing then (make && make install). For packages like deb, rpm etc it's not needed ofcourse because there is ability to create post installation scripts for example.

pevik:
I actually can see no scenario where arbitrary users in the system should be allowed to run rdisc to (possibly) change routing information of the system. That's why I hesitate to merge this patch.

PR #347, opened 2021-07-05
meson_options.txt: rdisc: Added SETCAP_OR_SUID_RDISC option

6eh01der:


pevik:
I'll move this to https://github.com/iputils/iputils/pull/348. It does not make sense to have it separated.
Also, you could put your all commits into single pull request :).

PR #348, opened 2021-07-05
rdisc: setcap

IBeholderI:
Added ability to execute setcap-setuid.sh on rdisc binary

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -51,11 +51,13 @@ setcap_arping = false
 setcap_clockdiff = false
 setcap_ping = false
 setcap_traceroute6 = false
+setcap_rdisc = false
 if (not get_option('NO_SETCAP_OR_SUID'))
   setcap_arping = get_option('SETCAP_OR_SUID_ARPING')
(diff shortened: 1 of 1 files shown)


pevik:
@6eh01der While the code is correct, can you explain why this is needed? Most of users nowadays have systemd, thus if they need this legacy software, they run it as a systemd service. Sure, there are few users without systemd, but why don't they run it just as root?

6eh01der:
@pevik Security. Why to run service as root if it is possible to run it as non-root? By the way systemd service also run as root by default too if no special directives was set like User=, DynamicUser= etc.

pevik:
@6eh01der isn't it just adding `DynamicUser` enough: https://github.com/iputils/iputils/pull/358
Do you really use other init than systemd that you require file based capabilities? Because our  script can also enable setuid, which I'm not sure if safer than running service as root.

6eh01der:
@pevik OpenRC for example which is default init system in distros like Alpine, Gentoo and some others

pevik:
Closed with 2d8d9a4e88ed5d436aedad3cbd8f68afdaabc71a

PR #349, opened 2021-07-05
setcap-setuid.sh: rdisc: Set correct caps

6eh01der:
rdisc requires cap_net_admin for ability to modify routing table (https://man7.org/linux/man-pages/man8/rdisc.8.html)
Added condition that checks that binary is rdisc and sets cap_net_raw and cap_net_admin to it.


6eh01der:
I'll create new request with one commit

pevik:
FYI instead of creating new PR where you fix something on current PR you can amend your git branch and push it with `-f`. Opening many PRs just pollutes the history.

PR #351, opened 2021-07-09
systemd unit: rdisc: Fixes of capabilities and type

6eh01der:
-t option is for testing purposes, so it's unknown for what reason it was added. So it much more nicer to replace it with -v option. -v is for detailed output verbosity. Additionally rdisc requires CAP_NET_ADMIN capability for ability to modify routing table. Without it there will be error: "ioctl (add/delete route): Operation not permitted"
Also service type must be set to "forking". Without it service will not start.

### DIFF
--- a/systemd/rdisc.service.in
+++ b/systemd/rdisc.service.in
@@ -5,11 +5,11 @@ Requires=network.target
 After=network.target
 
 [Service]
+Type=forking
 EnvironmentFile=-/etc/sysconfig/rdisc
-ExecStart=@sbindir@/rdisc -f -t $OPTIONS $SEND_ADDRESS $RECEIVE_ADDRESS
-
-AmbientCapabilities=CAP_NET_RAW
-CapabilityBoundingSet=CAP_NET_RAW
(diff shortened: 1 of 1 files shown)


pevik:
@iputils/maintainers LGTM, but could anybody test it?

pevik:
I suppose `-t` option is here due not going into background. But I need to verify it.

6eh01der:
Exactly. So i don't understand for what -t option is placed. I mean i don't understand for what to not place rdisc daemon into background.

pevik:
@6eh01der I took the only relevant fix:

```
-AmbientCapabilities=CAP_NET_RAW
-CapabilityBoundingSet=CAP_NET_RAW
+AmbientCapabilities=CAP_NET_RAW CAP_NET_ADMIN
+CapabilityBoundingSet=CAP_NET_RAW CAP_NET_ADMIN
```
Systemd service is working with `-t` which also don't require `Type=forking`. I'm not systemd expert and open to add also these changes, but please provide explanation *why* it's needed.

PR #352, opened 2021-07-09
setcap-setuid.sh: rdisc: Set correct capabilities

6eh01der:
rdisc requires cap_net_admin for ability to modify routing table (https://man7.org/linux/man-pages/man8/rdisc.8.html)
Added condition that checks that binary is rdisc and sets cap_net_raw and cap_net_admin to it.
Capabilities must be set with +ep because if rdisc running NOT by root or privileged user there will be error "socket: Operation not permitted" and process will be exited.

### DIFF
--- a/build-aux/setcap-setuid.sh
+++ b/build-aux/setcap-setuid.sh
@@ -15,9 +15,15 @@ _log() {
 
 case "$perm_type" in
 	caps)
-		params="cap_net_raw+p"
-		_log "calling: $setcap $params $exec_path"
-		"$setcap" $params "$exec_path"
+		if [ "$2" = "rdisc" ]; then
+			params="cap_net_raw,cap_net_admin+ep"
+			_log "calling: $setcap $params $exec_path"
+			"$setcap" $params "$exec_path"
+		else
+			params="cap_net_raw+p"
+			_log "calling: $setcap $params $exec_path"
+			"$setcap" $params "$exec_path"
+		fi
 	;;
 	setuid)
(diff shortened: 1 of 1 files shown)


build-aux/setcap-setuid.sh
```diff
@@ -15,9 +15,15 @@ _log() {
...
+		if [ "$2" = "rdisc" ]; then
+			params="cap_net_raw,cap_net_admin+ep"
+			_log "calling: $setcap $params $exec_path"
+			"$setcap" $params "$exec_path"
+		else
+			params="cap_net_raw+p"
+			_log "calling: $setcap $params $exec_path"
+			"$setcap" $params "$exec_path"
+		fi
 	;;
```
pevik, line 27: Why do you duplicate `_log` and `$setcap` call? Why not just:
```
params="cap_net_raw+p"
[ "$2" = "rdisc" ] && params="cap_net_raw,cap_net_admin+ep"
_log "calling: $setcap $params $exec_path"
"$setcap" $params "$exec_path"

```

pevik (approved):
LGTM (with cleanup to remove duplicity I suggested)
@iputils/maintainers any feedback?

build-aux/setcap-setuid.sh
pevik: I also wonder if `cap_net_raw,cap_net_admin+p` would be enough.

6eh01der: Yes, it must be +ep, not only +p or it will be error if you will try to run rdisc by unprivileged user

PR #356, opened 2021-07-16
rdisc: support /etc/default/rdisc environment file

nmeyerhans:
Red Hat derived distributions use /etc/sysconfig/, which Debian
derived distributions use /etc/default/ for a similar purpose.
Support Debian distributions by adding support for /etc/default/rdisc
as an optional EnvironmentFile in rdisc.service

Signed-off-by: Noah Meyerhans <noahm@debian.org>

### DIFF
--- a/systemd/rdisc.service.in
+++ b/systemd/rdisc.service.in
@@ -6,6 +6,7 @@ After=network.target
 
 [Service]
 EnvironmentFile=-/etc/sysconfig/rdisc
+EnvironmentFile=-/etc/default/rdisc
 ExecStart=@sbindir@/rdisc -f -t $OPTIONS $SEND_ADDRESS $RECEIVE_ADDRESS
 
 AmbientCapabilities=CAP_NET_RAW


pevik (approved):

pevik:
@nmeyerhans Do you plan to submit the same change for `systemd/rarpd.service.in`? Maybe we should add it even you don't plan to add `rarpd` to Debian package. Note: `ninfod` does not have any option to be added.

nmeyerhans:
> @nmeyerhans Do you plan to submit the same change for `systemd/rarpd.service.in`? Maybe we should add it even you don't plan to add `rarpd` to Debian package. Note: `ninfod` does not have any option to be added.

I hadn't considered it, as I don't plan on packaging ninfod for Debian. The protocols that it implements are long dead at this point and IMO we could probably make sense to delete or otherwise archive them within this repo.  But, as long as we're shipping it, it's probably reasonable to provide a complete unit file for it.

pevik:
I'm ok to remove `ninfod` and `rarpd` from iputils after release. Frankly speaking most of utils from iputils are pretty legacy (I personally use just `ping` and sometimes `arping`).

pevik:
@nmeyerhans FYI Debian old version of `rarpd` (probably version before adopted by iputils), which is still in unstable: https://tracker.debian.org/pkg/rarpd
IMHO it'd make sense to either remove it or replace it with version from iputils.

PR #358, opened 2021-07-21
systemd unit: rdisc: Add DynamicUser=yes

pevik:
to add extra security (sandboxing the service [1]).

Distro support: DynamicUser is supported from systemd 232 (Nov 2016),
also used in all other services.

NOTE: DynamicUser=yes implies PrivateTmp=yes (specified in all other
services) and implies RemoveIPC=yes (*not* specified).

[1] http://0pointer.net/blog/dynamic-users-with-systemd.html

### DIFF
--- a/systemd/rdisc.service.in
+++ b/systemd/rdisc.service.in
@@ -10,6 +10,7 @@ EnvironmentFile=-/etc/default/rdisc
 ExecStart=@sbindir@/rdisc -f -t $OPTIONS $SEND_ADDRESS $RECEIVE_ADDRESS
 AmbientCapabilities=CAP_NET_RAW CAP_NET_ADMIN
 CapabilityBoundingSet=CAP_NET_RAW CAP_NET_ADMIN
+DynamicUser=yes
 PrivateTmp=yes
 ProtectSystem=strict
 ProtectHome=yes


pevik:
@6eh01der FYI

6eh01der (commented):
Tested on systems with systemd version >=232 and everything is fine, but it does not work in distros like CentOS 7 for example. CentOS 7 has systemd version 219 and it is still popular. So for that distro and similar like RedHat 7 and Oracle Linux 7 there should be used User= and Group= directives and appropriate user and group creation, for example:

useradd --system --shell /usr/sbin/nologin --home-dir /dev/null rdisc

pevik:
Thanks for your review, I'll make a note at PR. But we already use `DynamicUser` in other profiles. Also newest iputils will not be updated on old CentOS, there are other dependencies: ninja, etc. which needs epel.
I'll test `rdist.service` on CentOS 7 and it's working with `DynamicUser=yes` (when installed into `/usr/local/sbin/`, not working, when installed into `/root/iputils-install/`).

6eh01der:
@pevik Strange, i've got CentOS 7 7.9.2009 with kernel 3.10.0-1160 and systemd 219 with rdisc in /usr/sbin and it doesn't work. More precisely rdisc works fine but it runs as root that is not right. In distros with systemd >= 232 with DynamicUser=yes rdisc runs as rdisc autocreated user.

pevik:
I tested it with very old 7.4.1708, then upgraded to recent 7.9.2009. In both cases `DynamicUser` just somehow gets ignored:

```
# systemctl status rdisc.service
● rdisc.service - Network Router Discovery Daemon
   Loaded: loaded (/usr/lib/systemd/system/rdisc.service; enabled; vendor preset: disabled)
   Active: active (running) since Čt 2021-07-22 14:17:55 BST; 3min 57s ago
     Docs: man:rdisc(8)
 Main PID: 1349 (rdisc)
   CGroup: /system.slice/rdisc.service
           └─1349 /usr/local/sbin/rdisc -f -t

# systemctl show rdisc.service |grep DynamicUser # nothing found

# grep DynamicUser /usr/lib/systemd/system/rdisc.service
DynamicUser=yes
```
I have no idea why.
Anyway, as I said, other units already use `DynamicUser` and nobody complained, thus it does not even make sense to mention it in release notes + I've mentioned it in the commit message.

6eh01der:
@pevik Just for info. I think that the answer is in your first message. DynamicUser directive was added in systemd version 232, but CentOS 7 has version 219. I tried to set some non-existent directive like "Test=yes" and service successfully start. So it seems like unknown directives just ignores.

pevik:
Yes: https://www.freedesktop.org/software/systemd/man/systemd.unit.html
> Unit files may contain additional options on top of those listed here. If systemd encounters an unknown option, it will write a warning log message but continue loading the unit.

PR #359, opened 2021-07-23
Make tests optional

heirecka:
Allows to avoid meson erroring out when ip isn't installed.


pevik:
Also alternative would be just to skip when `ip` not installed, this is perfectly ok. Thanks!

pevik:
I'm just going to rename variable to `SKIP_TESTS` (default false), amend a description a bit (tests are going to be for more tools than just for ping).

PR #360, opened 2021-07-27
arping: Fix ARP protocol field for AX.25 and NETROM

ralfbaechle:
AX.25 and NETROM differ from other, more ethernet-like protocols in that
they are not using a DIX protocol number but the AX.25 PID.  The arping code
doesn't handle this special case resulting in invalid ARP packets being sent.

The interface bpq0 is an AX.25-over-ethernet interface.  Without this
fix:

\# arping -c 1 -I bpq0 172.20.1.3
ARPING 172.20.1.3 from 172.20.1.2 bpq0
Sent 1 probes (1 broadcast(s))
Received 0 response(s)

With this fix:

\# arping -c 1 -I bpq0 172.20.1.3
ARPING 172.20.1.3 from 172.20.1.2 bpq0
Unicast reply from 172.20.1.3 [88:98:60:A0:92:40:02]  1.402ms
Sent 1 probes (1 broadcast(s))
Received 1 response(s)

Signed-off-by: Ralf Baechle <ralf@linux-mips.org>

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -37,6 +37,14 @@
 
 #include "iputils_common.h"
 
+/*
+ * As of July 2021 AX.25 PID values are not currently defined in any
+ * userspace headers.
+ */
+#ifndef AX25_P_IP
+#define AX25_P_IP		0xcc	/* ARPA Internet Protocol     */
+#endif
+
 #ifdef DEFAULT_DEVICE
 # define DEFAULT_DEVICE_STR	DEFAULT_DEVICE
 #else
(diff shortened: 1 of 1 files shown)


pevik:

pevik:
@ralfbaechle I suppose there is no virtual driver for AX25 in the linux kernel, right? Something like `vxcan`, `vxlan`. At least `net/ax25/Kconfig` does not mention any. I'd be surprised, but asking just in case.

ralfbaechle:
> @ralfbaechle I suppose there is no virtual driver for AX25 in the linux kernel, right? Something like `vxcan`, `vxlan`. At least `net/ax25/Kconfig` does not mention any. I'd be surprised, but asking just in case.

What you want is probably the bpqether driver.  BPQ (which isn't an acronym but the suffix of the ham radio callsign of the original developer) is a simple AX.25-over-ethernet encapsulation method.  To use it you only need to load the bpqether kernel module.
To enable, first enable CONFIG_HAMRADIO, CONFIG_AX25 and finally CONFIG_BPQETHER.
Loading the bpqether module then will create a bpq%d device for every ethernet device in the system.  Which is crazy but that's the way they did it in the old days and I have patches to clean that up.  Btw. you need to use ifconfig to configure the AX.25 bits of a bpq device.  Ip(8) from iproute2 doesn't support that yet.  I have patches for that and the associated kernel bits as well in the queue to be posted.

PR #362, opened 2021-07-28
[RFC] Remove traceroute6

pevik:
Lets drop software which does not make much sense to carry out.

NOTE: later something else to remove? I'd suggest `tftpd` (`dnsmasq` has much better alternative) and consider also `rarpd`, `rdisc`.


pevik:
@6eh01der @sevan @zenczykowski @floppym @nmav @vapier @yurchor @neheb FYI

zenczykowski:
Probably ifenslave too - ip should be used instead.

pevik:
> Probably `ifenslave; too - ip should be used instead.

`ifenslave` is not from iputils.

zenczykowski:
# rpm -q --filesbypkg iputils-20210202-2.fc34.x86_64 | egrep bin
iputils                   /usr/bin/arping
iputils                   /usr/bin/clockdiff
iputils                   /usr/bin/ping
iputils                   /usr/bin/tracepath
iputils                   /usr/sbin/arping
iputils                   /usr/sbin/ifenslave
iputils                   /usr/sbin/ping
iputils                   /usr/sbin/ping6
iputils                   /usr/sbin/rdisc
iputils                   /usr/sbin/tracepath
iputils                   /usr/sbin/tracepath6

pevik:
That's fedora downstream specific, it has never been part of iputils upstream.
https://src.fedoraproject.org/rpms/iputils/blob/rawhide/f/iputils.spec#_13
https://src.fedoraproject.org/rpms/iputils/blob/rawhide/f/sources
https://src.fedoraproject.org/lookaside/pkgs/iputils/ifenslave.tar.gz/6702c58e970e6a73fc8d9b7c9b63af35/ifenslave.tar.gz

nmeyerhans:
I'm in favor of dropping traceroute6.

I'd also be in favor of dropping tftpd and ninfod. Possibly rarpd and rdisc as well, but I'm familiar with the quality of the alternatives.

okias:
I would propose link to suitable replacement solution from build process, when traceroute6 will be requested

nmeyerhans (approved):

neheb:
Fine with me.

pevik:
Thanks! I'll wait few more days before removing it. I also created separated ticket to move the discussion there.

Issue #293, opened 2020-09-02
ping: socket: Address family not supported by protocol


lekto:
Hi, I got message "ping: socket: Address family not supported by protocol" when using ping without "-4" after update to s20200821.
```
lekto@blaszak ~ $ ping github.com
ping: socket: Address family not supported by protocol
PING github.com (140.82.121.3) 56(84) bytes of data.
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=1 ttl=50 time=19.5 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=2 ttl=50 time=18.7 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=3 ttl=50 time=18.3 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=4 ttl=50 time=17.8 ms
^C
--- github.com ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3003ms
rtt min/avg/max/mdev = 17.818/18.561/19.457/0.601 ms

lekto@blaszak ~ $ ping -4 github.com
PING github.com (140.82.121.3) 56(84) bytes of data.
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=1 ttl=50 time=19.4 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=2 ttl=50 time=18.2 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=3 ttl=50 time=27.6 ms
64 bytes from lb-140-82-121-3-fra.github.com (140.82.121.3): icmp_seq=4 ttl=50 time=18.7 ms
^C
--- github.com ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3002ms
rtt min/avg/max/mdev = 18.240/20.992/27.623/3.852 ms
```

I did git bisect and it's showing me that this commit is causing this:
```
904cdb6dc9bd511847e31689873f9c57172f6b8d is the first bad commit
commit 904cdb6dc9bd511847e31689873f9c57172f6b8d
Author: Sami Kerola <kerolasa@iki.fi>
Date:   Sat Nov 30 19:14:39 2019 +0000

    ping: AF_INET6 is address family not socket type [lgtm scan]
    
[...]

PR #370, opened 2021-09-16
ping: Remove 'unsupported IPv6' warning on disabled IPv6

pevik:
Regression was introduced in d141cb6 as introduced condition

`if ((errno == EAFNOSUPPORT && socktype == AF_INET6) || options & F_VERBOSE || requisite)`

was wrong, it should have been:

`if ((errno == EAFNOSUPPORT && family == AF_INET6 && requisite) || options & F_VERBOSE)`

but bug was hidden as `family == AF_INET6' was always false until otherwise correct fix 904cdb6 ("ping: AF_INET6 is address family not socket type [lgtm scan]") propagated the error.

Tested on kernel booted with ipv6.disable=1 (disabling via sysctl, i.e. `sysctl -w net.ipv6.conf.all.disable_ipv6=1; sysctl -w net.ipv6.conf.default.disable_ipv6=1` does not trigger the issue as it exit with "socket: Address family not supported by protocol" - errno EADDRNOTAVAIL).

Fixes: d141cb6 ("ping: work with older kernels that don't support ping sockets")
Closes: https://github.com/iputils/iputils/issues/293

Reported-by: lekto <lekto@o2.pl>
Suggested-by: klynastor

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -150,8 +150,8 @@ static void create_socket(struct ping_rts *rts, socket_st *sock, int family,
 		/* Report error related to disabled IPv6 only when IPv6 also failed or in
 		 * verbose mode. Report other errors always.
 		 */
-		if ((errno == EAFNOSUPPORT && family == AF_INET6) ||
-		    rts->opt_verbose || requisite)
+		if ((errno == EAFNOSUPPORT && family == AF_INET6 && requisite) ||
+		    rts->opt_verbose)
 			error(0, errno, "socket");
 		if (requisite)
 			exit(2);


pevik:
@klynastor would you mind to share your email address, so I can reference it in the commit message?

ac000:
Reviewed-by: Andrew Clayton <andrew@digital-domain.net>

klynastor:
> @klynastor would you mind to share your email address, so I can reference it in the commit message? I'll make you as the author of the patch, with myself as reviewer only (+ credit to the commit message).

Sure: gandalf@winds.org

Issue #326, opened 2021-04-19
Clockdiff host is down


LurenAA:
I have two computers，in a Campus Network.  they can ping each other. I can use ssh to connect one using the other one. but when I use clockdiff , both show "host is down". is there any thing wrong ? thanks 

ubuntu 20.04
5.4.0-42-generic
![image](https://user-images.githubusercontent.com/45759613/115235231-8bcd3500-a14c-11eb-8493-39f94dcc3bc8.png)

JokerShao:
same problem.

JokerShao:
but i can get right result use clockdiff from apt repository

pevik:
Can you please reproduce the problem on publicly visible hosts? That'd be much easier to reproduce for us. 

Can you test it on current master and then try to bisect if older version works?

if problem found on `clockdiff` in package which version and distro you use? Never versions (>= `s20190324`) support print version (`clockdiff -V`, for other versions we rely on distro package info.

@LurenAA Also please next time copy paste output. Github allows code formatting with `< >` button and text can be further copied.

JokerShao:
1. sorry, i can't reproduce the problem on publicly visible hosts, many router will ignore the ip and icmp timestamp;
    ```
    In my case, i have a dual network card laptop, one is ethernet(192.168.10.0/24), another is wireless networkcard(128.1.0.0/24).
    on ethernet card, it work well, but not work on wifi, echo is host down.
    ```
2. i have test the release version from 20210202 to 20180629, they're not work;
3. i don't known which version in apt repository, but i downloaded a version from `http://www.skbuff.net/iputils/`, iputils-s20151218.tar.bz2, it works well.

6eh01der:
@LurenAA I suppose that ICMP type 13 is blocked on target machine. Try to disable ufw. ICMP type 13 -  ICMP Timestamp. It used by default in clockdiff. Also you could add option -o and clockdiff will send ICMP ECHO packets with IP TIMESTAMP option instead of ICMP Timestamp packets. Successfully tested on Ubuntu 20.04.
@JokerShao Maybe apt version use -o option by default.

yueryoufeng:
@6eh01der Thank you for your reply.
@pevik  @kerolasa  The changed commit is https://github.com/iputils/iputils/commit/acb7fbbc48c368551d2f2da7d446d12ad7e1e172
Probably 323-327

themoses:
@pevik I sent you a mail containing the details. Thanks for pinging me

[...]

PR #374, opened 2021-10-14
fix clockdiff host id down

yueryoufeng:
The clockdiff command always reports an error： host is down
Plus parameters -o and -o1 have problems
Change the code logic in this commit:https://github.com/iputils/iputils/commit/acb7fbbc48c368551d2f2da7d446d12ad7e1e172
but the commit message :The measure() was a bit too long and deeply indented (8 steps).

the issues:https://github.com/iputils/iputils/issues/326

### DIFF
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -328,7 +328,7 @@ static int measure_inner_loop(struct run_state *ctl, struct measure_vars *mv)
 		if (diff < RANGE) {
 			mv->min1 = delta1;
 			mv->min2 = delta2;
-			return BREAK;
+			return GOOD;
 		}
 	}
 	return CONTINUE;
@@ -420,6 +420,8 @@ static int measure(struct run_state *ctl)
 				case BREAK:
 					escape = 1;
 					break;
+				case GOOD:
+					goto good_exit;
 				case CONTINUE:
 					continue;
 				default:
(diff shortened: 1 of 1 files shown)


6eh01der:
There is another problem with this variant (it's the same old code that was before 20180629 version inclusive). If rtt to host is around 1ms then clockdiff without -o option will return strange rtt values like 750(187) ms, 562(280) ms, 178(265) ms etc. And this is not correct i think. Or maybe i don't understand something. In later versions rtt value in same situation will be returned as 0. I have some idea about it and i'll test it in evening today. And what the problems do you mean with -o and -o1 options?

clockdiff.c
```diff
@@ -420,13 +420,16 @@ static int measure(struct run_state *ctl)
 				case BREAK:
 					escape = 1;
 					break;
+				case GOOD:
```
6eh01der, line 423: For what reason you add GOOD case?

clockdiff.c
```diff
@@ -328,7 +328,7 @@ static int measure_inner_loop(struct run_state *ctl, struct measure_vars *mv)
 		if (diff < RANGE) {
 			mv->min1 = delta1;
 			mv->min2 = delta2;
-			return BREAK;
+			return GOOD;
```
6eh01der, line 331: Maybe goto good_exit; instead as it was before 20190324 ?

6eh01der (commented):
I will create new pull request with your code and plus rtt fix. Please close this request.

PR #373, opened 2021-09-20
ping: allow pinging IPv6 multicast addresses

teknoraver:
Allow to ping IPv6 multicast and anycast addresses with the `-b` flag,
as it's already possible with IPv4.

Fixes: 5e052ad ("ping: discard packets with wrong source address")
Closes: #371
Signed-off-by: Matteo Croce <mcroce@microsoft.com>

PR #376, opened 2021-10-15
ping: allow pinging IPv6 multicast/anycast addresses (improved)

pevik:
Based on https://github.com/iputils/iputils/pull/373, with changes:
* relax dropping reply (from 5e052ad), print `DIFFERENT ADDRESS!` instead - needs to update translations. Original version of this PR printed `WRONG SOURCE!`, but that could be misleading in case address is not *wrong* for other sources, also mention address is a bit more informative. 
NOTE: it'd be nice to put all warnings into `ping` man page (currently none of them is there).
* IT detect Subnet-Router anycast only for the default prefix 64. I wonder if (because Subnet-Router anycast should be on local network) prefix could be detected with netlink from local interface.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1504,6 +1504,7 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 	int reply_ttl;
 	uint8_t *opts, *tmp_ttl;
 	int olen;
+	int wrong_source = 0;
 
 	/* Check the IP header */
 	ip = (struct iphdr *)buf;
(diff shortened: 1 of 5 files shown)


teknoraver:
Reviewed-by: Matteo Croce <mcroce@microsoft.com>

pevik:
> > Thanks for your ack, feel free to post your tags (`Reviewed-by:`). I might try to add just detection for Subnet-Router anycast address on default prefix (64).
> 
> It would be nice to add this.

Thanks.

> 
> In the current implementation the `WRONG SOURCE` phrase is misleading in the subnet-router case. Perhaps a more generic message about it being different from the target?

Yes, subnet-router case would could be detected, thus this could be fixed.

I'm sorry I need to rethink whole case (but this should be resolved before release).
1) Because the original reason for adding -b for broadcast was to add sockopt `SO_BROADCAST`. I'm not sure if this is correct for multicast and anycast. Also we detect IPv6 multicast with `IN6_IS_ADDR_MULTICAST()`, thus adding multicast to `-b` does not sound good to me.
2) Also you added meaning for `-b` to not hide reply with wrong source address regardless whether it's broadcast - you can use `-b` for your original case in 5e052ada96c3fc457efe66b2fed5c14ed8de6ae3 (which is something different than broadcast).
3) I'm not sure if reporting address (change in this PR) or hiding it (as currently implemented) should be only for echo reply.
4) Maybe broadcast should be detected.

Also `WARNING: pinging broadcast address` on `-b` would need to be updated, but that's trivial.

pevik:
@nmeyerhans @iputils/maintainers could you please review this?

pevik:
Fixed with commits: 15a5e5c7aace5a7a782ff802988e04ed4c1148a5 5f6bec5ab57cc8beaa78f5756a0ffbdf01f28d36 dd0bcac98d10ea340dfe8aff43f26fe1a7f77931.

PR #378, opened 2021-11-09
ping: cleanup (DRY)

pevik:
To reduce duplicity.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -594,6 +594,46 @@ main(int argc, char **argv)
 	return ret_val;
 }
 
+static int iface_name2index(struct ping_rts *rts, int fd)
+{
+	struct ifreq ifr;
+
+	memset(&ifr, 0, sizeof(ifr));
+	strncpy(ifr.ifr_name, rts->device, IFNAMSIZ - 1);
+
+	if (ioctl(fd, SIOCGIFINDEX, &ifr) < 0)
+		error(2, 0, _("unknown iface: %s"), rts->device);
+
+	return ifr.ifr_ifindex;
+}
+
+static void bind_to_device(struct ping_rts *rts, int fd, in_addr_t addr)
+{
+	int rc;
+	int errno_save;
+
+	enable_capability_raw();
+	rc = setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, rts->device,
+			strlen(rts->device) + 1);
+	errno_save = errno;
+	disable_capability_raw();
+
+	if (rc != -1)
+		return;
+
+	if (IN_MULTICAST(ntohl(addr))) {
+		struct ip_mreqn imr;
+
+		memset(&imr, 0, sizeof(imr));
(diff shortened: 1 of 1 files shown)


ping/ping.c
```diff
@@ -758,47 +772,14 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
...
-	if (rts->device) {
-		struct ifreq ifr;
-
-		memset(&ifr, 0, sizeof(ifr));
-		strncpy(ifr.ifr_name, rts->device, IFNAMSIZ - 1);
-		if (ioctl(sock->fd, SIOCGIFINDEX, &ifr) < 0)
-			error(2, 0, _("unknown iface: %s"), rts->device);
-	}
+	if (rts->device)
+		iface_name2index(rts, sock->fd);
```
metan-ucw, line 782: Looking at this part of the code, is the result even used? Can't we just delete this part?

We get the result in the ifr.ifr_ifindex but before it was not used at all, now it's returned from the function but again the return value is not used.

ping/ping.c
```diff
@@ -665,39 +705,13 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
...
-						imr.imr_ifindex = ifr.ifr_ifindex;
-						if (setsockopt(fd, SOL_IP, IP_MULTICAST_IF,
-							       &imr, sizeof(imr)) == -1)
-							error(2, errno, "IP_MULTICAST_IF");
-					} else
-						error(2, errno_save, "SO_BINDTODEVICE %s", rts->device);
-				}
-			}
+			for (i = 0; i < 2; i++)
+				bind_to_device(rts, fds[i], dst.sin_addr.s_addr);
```
metan-ucw, line 714: I guess that we can get rid of the fds[] array as well, since we have put the code into the function we can do:
```
bind_to_device(rts, probe_fd, dst.sin_addr.s_addr);
bind_to_device(rts, sock->fd, dst.sin_addr.s_addr);
```

Other than that the code looks good to me.

ping/ping.c
pevik: Here is used to check for unknown interface in case of ` -I <VRF interface> and -I <local IPv4 address>`:

```
$ ./builddir/ping/ping 1:2::3:4 -I nonexisting -I 1:2::3:4 -c 1
builddir/ping/ping: unknown iface: nonexisting
```
Maybe it's a bit confusing use the same function for both: 1) getting interface index 2) just for checking interface exists. Any idea how to do it better?

Also, This check should be moved to getopts, but I'd prefer to do this after release. `-I` can contain interface or address.

pevik: Actually, you're right this check is probably not needed as it's used only for `I <VRF interface> and -I <local IPv4 address>`, but check for invalid device has been done earlier as part of `bind_to_device()` call in ping.c:786 `ping4_run()`.

PR #380, opened 2021-11-12
Clockdiff: fix loop and rtt

6eh01der:
Fix of #326 

It's almost the same code that was used before [s20190324](https://github.com/iputils/iputils/releases/tag/s20190324) version (look at [s20180629](https://github.com/iputils/iputils/releases/tag/s20180629) ) for situation when diff is less then 1 ms. Additionally it include little fix for avoid return incorrect rtt value (750(187)ms and etc).

### DIFF
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -328,7 +328,7 @@ static int measure_inner_loop(struct run_state *ctl, struct measure_vars *mv)
 		if (diff < RANGE) {
 			mv->min1 = delta1;
 			mv->min2 = delta2;
-			return BREAK;
+			return GOOD;
 		}
 	}
 	return CONTINUE;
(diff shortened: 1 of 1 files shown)


yueryoufeng:
I don't know why the loop has to be executed 50 times.

yueryoufeng:
@6eh01der Is there anything else I need to do with this PR? Or have been waiting?

6eh01der:
> @6eh01der Is there anything else I need to do with this PR? Or have been waiting?

Just wait when commit will be approved. Or declined)

pevik:
@yueryoufeng @6eh01der I'll try to have look on this tomorrow/this week. But clockdiff is a bit hard for me to test (steps to reproduce is welcome).

yueryoufeng:
Thanks for your reply. on my machine, the two machines are in the same network environment, and they appear probably every 10 times.
@pevik

pevik:
Leaving this for next release.

pevik:
Thanks for your effort. In the end we took https://github.com/iputils/iputils/pull/459.

Issue #302, opened 2020-12-03
Due to EINTR & EAGAIN error found ping packet loss of 1


apriya-stack:
Idea- Replace "break" with  "continue"  for below condition- 
 if ((errno == EINTR && !recv_error) ||
      (errno == EAGAIN && !recv_error))
       continue;

This is mostly an issue when running "fluid ping test" (ping -f)-
errno=4 recv_error=0receive_msg=-1
errno=11 recv_error=0
--- 10.0.0.1 ping statistics ---
100 packets transmitted, 99 received, 1% packet loss, time 71ms

kerolasa:
@apriya-stack it looks like you read the code and figured out what needs to be changed. Could you send pull request instead of adding an issue.

6eh01der:
@apriya-stack for some reasons didn't create request so i will do it

pevik:
@6eh01der Could you please add some example how to reproduce the issue? Ideally using network namespaces?

6eh01der:
> @6eh01der Could you please add some example how to reproduce the issue? Ideally using network namespaces?

As mentioned in issue's description - ping -f ntp.org for example

pevik:
I'm sorry, I'm somehow slow, is the issue `1% packet loss`? Do you expect `O% packet loss`?

Also, maybe better to test localhost to not influence network:
```
# ping -f ::1 -c 30000
PING ::1(::1) 56 data bytes
 
--- ::1 ping statistics ---
30000 packets transmitted, 30000 received, 0% packet loss, time 822ms
rtt min/avg/max/mdev = 0.003/0.006/0.583/0.004 ms, ipg/ewma 0.027/0.006 ms
```

6eh01der:
I'm just pulled code implemented by @apriya-stack without deep dive into issue and that's my mistake.

@apriya-stack talk about that the EAGAIN is treated as packet loss in flood ping scenario if i understand right. And i think that the right question is should it be treated as loss, or not because in fact EAGAIN does not mean that the packet was lost...

`sudo strace ping -f 192.168.1.1`

`gettimeofday({tv_sec=1636919514, tv_usec=398432}, NULL) = 0
[...]

PR #382, opened 2021-11-12
Ping: fix of EINTR & EAGAIN

6eh01der:
Fix of issue #302 authored by @apriya-stack


pevik:
If you look into CI results, you see it broke `ping -c5 -i0.1 127.0.0.1` on all native tests (which can actually run tests). Which means that code is probably not correct.

6eh01der:
Yes, it stucks, code not correct, you're right...

PR #390, opened 2022-01-02
tests: do not error out if ip program cannot be found

### LINKED ISSUE #383: 20210722: build fails on systems without iproute2 installed
build fails on systems without iproute2 installed and iproute seem to be only needed for test (so that ip shouldn't be needed just for building)

````
[...]
Run-time dependency threads found: YES
Has header "error.h" : YES
Program /usr/bin/meson found: YES (/usr/bin/meson)
Configuring config.h using configuration
Program /sbin/setcap found: YES (/sbin/setcap)

ping/meson.build:38:0: ERROR: Program or command 'ip' not found or not executable

A full log can be found at /tmp/B.x7dvybsu/BUILD/iputils-20210722/build/meson-logs/meson-log.txt
````

### PR DESCRIPTION
**Pull request by eli-schwartz**
Failure to run `ip` should produce the same results as running it and discovering that, indeed, ipv6 is not available. But meson aborted because a missing program is a stricter issue than one that simply returns non-zero. So, do an upfront check for the program and disable those tests if it cannot be found.

While we are at it, optimize the function call chain to drop an unneeded .strip() before .contains()

Fixes #383

### DIFF
--- a/test/ping/meson.build
+++ b/test/ping/meson.build
@@ -5,8 +5,8 @@
 # https://github.com/actions/virtual-environments/issues/668
 ipv6_dst = []
 ipv6_switch = []
-r = run_command('ip', '-6', 'a')
-if r.stdout().strip().contains('::1')
+ip = find_program('ip', required: false)
+if ip.found() and run_command(ip, '-6', 'a').stdout().contains('::1')
   message('IPv6 enabled')
   ipv6_dst = [ '::1' ]
   ipv6_switch = [ '-6' ]

### CODE REVIEW

**Review by nmeyerhans — CHANGES_REQUESTED** [review]
LGTM, but could you add a DCO Signed-off-by line to your commit?

**Comment by eli-schwartz** [comment]
Err... good point. :D I am much better at remembering signoffs when I'm doing it for mailing list / patchwork based projects.

**Comment by eli-schwartz** [comment]
(Actually I don't even see this mentioned in any kind of contribution guidelines or in the README. It is obvious from a glance at git log...)

**Comment by pevik** [comment]
LGTM, I'm going to merge with this diff  (https://github.com/pevik/iputils/commits/gh%23383):

```
diff --git a/test/ping/meson.build b/test/ping/meson.build
index f41dedd..4e4761f 100644
--- a/test/ping/meson.build
+++ b/test/ping/meson.build
@@ -10,6 +10,8 @@ if ip.found() and run_command(ip, '-6', 'a').stdout().contains('::1')
   message('IPv6 enabled')
   ipv6_dst = [ '::1' ]
   ipv6_switch = [ '-6' ]
+elif not ip.found()
+  message('WARNING: ip binary not found => disable IPv6 tests')
 else
   message('WARNING: IPv6 disabled')
 endif
```
We could also expect IPv6 to be enabled if we don't find `ip`, but let's keep it this way to avoid false positives.

NOTE: I preferred to to use `/proc/net/if_inet6` for detection (see https://github.com/pevik/iputils/commits/gh%23383.if_inet6), but that code creates infinite loop. Thus I'll postpone using it after porting LTP shell test library.

Issue #125, opened 2018-04-18
Question: why do we need `probe_fd` in ping.c?


amitsaha:
I was going through `ping.c` to understand its working. Is there any reason to use a different (new socket) `probe_fd`? Why can't we use the `sock->fd` which is what we eventually use to send the ICMP packets over. Basically:

```
 git diff
diff --git a/ping.c b/ping.c
index efc4c58..6ff19e2 100644
--- a/ping.c
+++ b/ping.c
@@ -599,7 +599,7 @@ int ping4_run(int argc, char **argv, struct addrinfo *ai, socket_st *sock)
        if (source.sin_addr.s_addr == 0) {
                socklen_t alen;
                struct sockaddr_in dst = whereto;
-           int probe_fd = socket(AF_INET, SOCK_DGRAM, 0);
+         int probe_fd = sock->fd;

                if (probe_fd < 0) {
                        perror("socket");
@@ -700,7 +700,7 @@ int ping4_run(int argc, char **argv, struct addrinfo *ai, socket_st *sock)
                                fprintf(stderr, "ping: Warning: source address might be selected on device other than %s.\n", device);
                }
 #endif
-           close(probe_fd);
+         //close(probe_fd);
        } while (0);

        if (whereto.sin_addr.s_addr == 0)
```

Obviously there is a good reason for using a different socket. Wondering what that is?

Thanks.

svagner:
Hi! I have the same question. Could someone explain why we have  here separate socket?

pevik:
Hm, that code origin from the code imported by YOSHIFUJI Hideaki from ftp://ftp.inr.ac.ru/ip-routing/iputils-ss020927.tar.gz (see 33370345c7).

I guess it's because different protocol: the "working" socket `sock->fd` use `IPPROTO_ICMP` (value `1`) or `IPPROTO_ICMPV6` (value `58`), but `probe_fd` socket use `0` (`IPPROTO_IP`).

I'm not really familiar with socket protocols, see also [protocols(5)](http://man7.org/linux/man-pages/man5/protocols.5.html) and [socket(2)](http://man7.org/linux/man-pages/man2/socket.2.html).

zenczykowski:
It's a well known socket programming trick to use a UDP socket() connect() getsockname() close() to do a route lookup (this generates no network traffic).

The kernel did not support per-protocol specific routing.  (however that's no longer the case)
So you could simply use a udp socket.
Now we should be using an icmp socket...

[...]

PR #391, opened 2022-01-10
ping: Call connect() before sending/receiving

gal-pressman:
Once a socket is created, the connect() syscall must be called to
connect to the remote before using send()/recv() or sendto()/recvfrom().

Note that there is an existing connect() call on the probe_fd socket,
but it is not enough as it is not the socket that is later used to
transmit/receive the ping packets.

Fixes: 33370345c7d8 ("Initial import of iputils")
Signed-off-by: Gal Pressman <gal@nvidia.com>

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -31,6 +31,10 @@ xml:id="man.ping">
         <replaceable>count</replaceable></option>
       </arg>
       <arg choice="opt" rep="norepeat">
+        <option>-C
+        <replaceable>connect_sk</replaceable></option>
+      </arg>
+      <arg choice="opt" rep="norepeat">
         <option>-F
         <replaceable>flowlabel</replaceable></option>
       </arg>
@@ -192,6 +196,15 @@ xml:id="man.ping">
           <command>ping</command> waits for
           <emphasis remap="I">count</emphasis> ECHO_REPLY packets,
           until the timeout expires.</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+          <option>-C</option>
+          <emphasis remap="I">connect_sk</emphasis>
+        </term>
+        <listitem>
+          <para>Call connect() syscall on socket creation</para>
         </listitem>
       </varlistentry>
       <varlistentry>
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -357,7 +357,7 @@ main(int argc, char **argv)
 		hints.ai_family = AF_INET6;
 
 	/* Parse command line options */
-	while ((ch = getopt(argc, argv, "h?" "4bRT:" "6F:N:" "aABc:dDe:fi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:")) != EOF) {
+	while ((ch = getopt(argc, argv, "h?" "4bRT:" "6F:N:" "aABc:CdDe:fi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:")) != EOF) {
 		switch(ch) {
 		/* IPv4 specific options */
 		case '4':
@@ -417,6 +417,9 @@ main(int argc, char **argv)
 			break;
 		case 'c':
 			rts.npackets = strtol_or_err(optarg, _("invalid argument"), 1, LONG_MAX);
+			break;
+		case 'C':
+			rts.opt_connect_sk = 1;
 			break;
 		case 'd':
 			rts.opt_so_debug = 1;
@@ -738,6 +741,7 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 	char hnamebuf[NI_MAXHOST];
 	unsigned char rspace[3 + 4 * NROUTES + 1];	/* record route space */
 	uint32_t *tmp_rspace;
+	struct sockaddr_in dst;
 
 	if (argc > 1) {
 		if (rts->opt_rroute)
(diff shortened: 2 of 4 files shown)


pavlix:
Care to explain why would you need to `connect()` in the first place? What is the problem, error, expected and actual behavior?

galpress:
@pavlix Sure!

I'm setting up an ipip tunnel with a higher mtu than the physical interface and testing the pmtu discovery using the '-M want' flag.
Since the kernel does the pmtu assignment for each route during connect(), setting the routing tables explicitly instead of the default rules result in packet drops.

For example (replace the ips for remote side):
```
ifconfig eth2 1.1.1.1/16 up
ip tunnel add ipinip1 mode ipip remote 1.1.1.2 local 1.1.1.1 dev eth2
ip link set dev ipinip1 up
ip addr add 2.1.1.1/16 dev ipinip1
ip rule add iif eth2 lookup local pref 0
ip rule add from 1.1.1.1 table 10 pref 10
ip route add 1.1.0.0/16 dev eth2 src 1.1.1.1 table 10
ip route add local 1.1.1.1 dev eth2 src 1.1.1.1 table 10
ip rule add iif ipinip1 lookup local pref 0
ip rule add from 2.1.1.1 table 12 pref 10
ip route add 2.1.0.0/16 dev ipinip1 src 2.1.1.1 table 12
ip route add local 2.1.1.1 dev ipinip1 src 2.1.1.1 table 12
ifconfig ipinip1 mtu 9600
ping -c 5 2.1.1.2 -I ipinip1 -i 0.2 -s 1453 -M want -W 1
```

This snippet sets the ipip tunnel's mtu to 9600 (eth2 default mtu 1500), and adds explicit routing rules to direct the traffic through the appropriate interfaces.
In this case, all ping packets are lost. Adding the connect() sets the pmtu for the '2.1.0.0/16' rule in table 12 and fixes it.

pavlix:
I also wonder whether the question about separate `probe_fd` still applies.

https://github.com/iputils/iputils/issues/125

pevik:
> I'm setting up an ipip tunnel with a higher mtu than the physical interface and testing the pmtu discovery using the '-M want' flag.
> Since the kernel does the pmtu assignment for each route during connect(), setting the routing tables explicitly instead of the > default rules result in packet drops.

@galpress Thanks for a bug report and code. The description should be in the commit (these special case must be well documented, i.e. describe *why* change is needed).

@pavlix welcome back, your review and code is highly appreciated :).

I'll try next week to look into this. And also reproduce it using network namespaces.

NOTE: we use network namespaces to be able to reproduce the issue regardless the environment.
https://github.com/iputils/iputils/pull/340#issuecomment-987219698
https://github.com/iputils/iputils/pull/296#issuecomment-710044824

pavlix:
> @pavlix welcome back, your review and code is highly appreciated :).

Thank you for the warm welcome. Feel free to reach out to me on Twitter/Discord/whatever fits. Regarding the review, I already tried to reach out to a few folks who are into IP protocols, let us see what they have to say.

galpress:
Thank you @pavlix @pevik, updated the commit message.

galpress:
Hey, kind reminder.

galpress:
Did anyone get a chance to take a look at this?

pevik:
I'll try to have look after https://github.com/iputils/iputils/issues/398. To be honest, first I'd like to find time to write tests for ping, because I'm not a network expert and there has already been many regressions in ping.

galpress:
@pevik any updates?

galpress:
Hey, what can I do to get this patch accepted?

pevik:
@galpress I'm sorry, lack of time on my side (there is no other maintainer). Understand that there has been many regressions in ping, thus careful review and extensive testing must be done.

pevik:
@galpress I'm trying to rewrite this to use network namespaces to be able to test it.

galpress:
@pevik thanks for the update!

richiejp:
If we call connect on a broadcast or multicast address does the kernel then filter responses to be inside the relevant address ranges? 

The ping util itself doesn't seem to try validating the source of a response if unicast or multicast are enabled.

Also I guess if we call connect on a unicast address then we will no longer see responses from the wrong source, which seems to be collected and printed as part of the statistics.

For reference see the man page. I haven't found the relevant code in the kernel.

mkubecek:
I don't think that calling `connect()` unconditionally is a good idea, the side effects (e.g. those mentioned by richiejp) would IMHO do more harm than the issue this is trying to address. I'm not even sure current behaviour is actually wrong, after all the `-M` parameter is intended to serve as an interface to `IP_MTU_DISCOVER` and `IPV6_MTU_DISCOVER` socket options so it should IMHO follow its semantics.

pevik:
David Ahern, the IPv4/IPv6 kernel maintainer was also against calling `connect()` unconditionally. He suggested to add command line option to call `connect()`. That sounds reasonable.
https://lore.kernel.org/netdev/d47c3f41-2977-3ffb-5c99-953088727a4b@gmail.com/

galpress:
Thanks for the link to the netdev thread, wasn't aware of it.
Please let me know how you wish to proceed, should I add a new CLI knob?

I'd love to get this issue fixed, but it's there since the beginning of time so I guess it won't be the end of the world if we leave it as is.

pevik:
> Thanks for the link to the netdev thread, wasn't aware of it.

I'm sorry I thought I Cc you, but obviously I didn't.

> Please let me know how you wish to proceed, should I add a new CLI knob?

Please add new `-C` option as `-c` has been taken for count and add new boolean opt `opt_connect:1` into `struct ping_rts`.
 
> I'd love to get this issue fixed, but it's there since the beginning of time so I guess it won't be the end of the world if we leave it as is.

If you don't have time for this, I can finish it. I'd like to freeze git before release this or next week.

galpress:
@pevik thanks, done.

PR #394, opened 2022-02-15
ping: fix -m (SO_MARK) in 32-bit archs

luizluca:
strtol_or_err() uses signed int for checking boundaries. However, '-m'
uses UINT_MAX as upper limit. For 32-bit archs, that UINT_MAX becomes a
negative value, invalidating all possibe values.

A new strtoul_or_err() is now in use for '-m'. It is equal to
strtol_or_err(), but it uses unsigned variables. Most, if not all,
uses of strtol_or_err() could be replaced by strtoul_or_err().

The error message in strtol_or_err() was also fixed. It was printing
signed boundaries as unsigned ones, hiding the fact that the upper
boundary was treated as negative. The opposite was happening with
sock_setmark().

Although setsockopt(,,SO_MARK,,) treat that value as signed int, the
correct value eventually get into the kernel u32 variable.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/iputils_common.c
+++ b/iputils_common.c
@@ -79,8 +79,29 @@ long strtol_or_err(char const *const str, char const *const errmesg,
 	if (errno || str == end || (end && *end))
 		goto err;
 	if (num < min || max < num)
-		error(EXIT_FAILURE, 0, "%s: '%s': out of range: %lu <= value <= %lu",
+		error(EXIT_FAILURE, 0, "%s: '%s': out of range: %ld <= value <= %ld",
 		      errmesg, str,  min, max);
+	return num;
+ err:
+	error(EXIT_FAILURE, errno, "%s: '%s'", errmesg, str);
+	abort();
+}
+
+unsigned long strtoul_or_err(char const *const str, char const *const errmesg,
+		   const unsigned long min, const unsigned long max)
+{
+	unsigned long num;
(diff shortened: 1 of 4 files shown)


luizluca:
I'm working in the kernel side fix. However, I believe all uses of strtol_or_err() can be blindly replaced by strtoul_or_err() and that old function can be dropped (or, depending on the perspective, renamed).

pevik:
@luizluca LGTM, thx!

pevik:
@luizluca LGTM, thx!

pevik:
> I'm working in the kernel side fix. However, I believe all uses of strtol_or_err() can be blindly replaced by strtoul_or_err() and that old function can be dropped (or, depending on the perspective, renamed).

@luizluca Agree, everything can use your `strtoul_or_err()` as there is no negative value allowed. Also probably all variables holding it these values could be made `unsigned int` (but that needs to be checked). Do you plan to finish this or shell I do it?

And there could be even macro to not repeat code a lot:
```
#define SAFE_STRTOUL(num, min, max) strtol_or_err(num, _("invalid argument"), min, max)
```

metan-ucw:
The patch itself looks good to me: `Reviewed-by: Cyril Hrubis <chrubis@suse.cz>`

PR #396, opened 2022-03-07
ping: revert "ping: do not bind to device when destination IP is on d…

lxin:
…evice"

The issue should be fixed in kernell side, see:

https://lore.kernel.org/all/ea03066bc7256ab86df8d3501f3440819305be57.1644988852.git.lucien.xin@gmail.com/

Otherwise it would cause regressions in some kernel selftests.

Signed-off-by: Xin Long <lucien.xin@gmail.com>


pevik:
@lxin FYI we consider reverting cc44f4c as you did in this PR (see https://github.com/iputils/iputils/pull/405#issuecomment-1119768883), but IMHO nobody from the community knows all the corner cases which need to be tested. Given you fixed it in the kernel:

* [2afc3b5a31f9 ("ping: fix the sk_bound_dev_if match in ping_lookup")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2afc3b5a31f9edf3ef0f374f5d70610c79c93a42) (v5.17-rc2)
* [35a79e64de29 ("ping: fix the dif and sdif check in ping_lookup")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=35a79e64de29e8d57a5989aac57611c0cd29e13e) (v5.17-rc5)
* [cd33bdcbead8 ("ping: remove pr_err from ping_lookup")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cd33bdcbead882c2e58fdb4a54a7bd75b610a452) (v5.17-rc6)

 and these fixes has been backported to all stable / LTS, most of the users which install updates have problem solved. It's just a question whether all enterprise distros backport these fixes and how many users will continue using unfixed kernels. Also users with unfixed kernel will probably not update iputils either. To be clear, I'm not against fixing it also in userspace, but we must be sure we avoid regression.

nit: it'd be better to run `git revert cc44f4c` and put your extra info bellow. The reason is that `git revert` produces typical message (easy to find with `git grep --log`, puts full hash of the reverted commit).

pevik:
@lxin patch in kernel you reference was merged into kernel mainline in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=35a79e64de29e8d57a5989aac57611c0cd29e13e (v5.17-rc5, released in v5.17). Do you still think this should be reverted? That would fix older kernels which does not have fix, but OTOH most of linux distros use LTS kernels whic got this fix.

lxin:
Hi, Petr

I'm not sure. cc44f4c has also caused some regression though it's only seen
in the kselftests, which is also why I made the kernel fix. I don't know
which regression is more serious regardless of the patch's correctness:).

Note that, there's a workaround for the original issue, not using the PING
socket by:

# sysctl -w net.ipv4.ping_group_range="1 0"

Also, we should also be aware of how the iputils version and kernel version
are related in the distro. like the latest iptuils may require the
kernel version to be newer than kernel-x.y, or we can put some changelog to
say that this revert should be backported together with the kernel commit.

Thanks.

pevik:
David Ahern, the IPv4/IPv6 kernel maintainer also agreed with reverting this
https://lore.kernel.org/netdev/d47c3f41-2977-3ffb-5c99-953088727a4b@gmail.com/
Two strong voices to revert, I'll accept that.

Issue #398, opened 2022-03-10
ping: Allow to use -I<ifname|scope-id> for IPv6 link-local scope address


pevik:
[man getaddrinfo(3)](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html) states support for the `address%scope-id` notation for specifying the IPv6 scope-ID. man ping(8) used to require to use it.  During socket creation refactoring in 243b6e1 it was possible to ping link-local scope address without `%<scope-id|ifname>` notation.  Everything worked as expected because the default was still raw socket with capability:
    
	$ ip -6 addr show vnet2
	20: vnet2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master virbr0 state UNKNOWN group default qlen 1000
		inet6 fe80::fc54:ff:fe9e:936c/64 scope link
		   valid_lft forever preferred_lft forever

	$ sudo sysctl net.ipv4.ping_group_range="0 1" # disable ICMP socket
	$ sudo setcap cap_net_raw+p ./builddir/ping/ping # CAP_NET_RAW for normal user

	$ ./builddir/ping/ping -c1 fe80::fc54:ff:fe9e:936c # link-local without %scope-id
	PING fe80::fc54:ff:fe9e:936c(fe80::fc54:ff:fe9e:936c) 56 data bytes
	64 bytes from fe80::fc54:ff:fe9e:936c%vnet2: icmp_seq=1 ttl=64 time=0.058 ms

	--- fe80::fc54:ff:fe9e:936c ping statistics ---
	1 packets transmitted, 1 received, 0% packet loss, time 0ms
    
But later when distros moved to ICMP socket it's required to use %<scope-id>:
    
	$ sudo sysctl net.ipv4.ping_group_range="0 2147483647" # enable ICMP socket

	$ ./builddir/ping/ping -c1 fe80::fc54:ff:fe9e:936c
	PING fe80::fc54:ff:fe9e:936c(fe80::fc54:ff:fe9e:936c) 56 data bytes

	--- fe80::fc54:ff:fe9e:936c ping statistics ---
	1 packets transmitted, 0 received, 100% packet loss, time 0ms
    
We could set scope-id or ifname only via % notation:
    
1) `destination%<ifname>`

```
./builddir/ping/ping -c1 fe80::fc54:ff:fe9e:936c%vnet2
[...]

PR #399, opened 2022-03-10
ping: Allow to use -I<ifname|scope-id> for IPv6 link-local scope address

pevik:
See https://github.com/iputils/iputils/issues/398
**Superseded by https://github.com/iputils/iputils/pull/405**

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -555,6 +555,24 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
+	/*
+	 * Allow to use -I<ifname|scope-id> for IPv6 link-local scope address
+	 * when passing it to getaddrinfo():
+	 * i.e. -I<ifname|scope-id> => %<ifname|scope-id>
+	 */
+	unsigned char buf[sizeof(struct in6_addr)];
+	int free_target = 0;
+	if (!strchr(target, '%') && inet_pton(AF_INET6, target, buf) > 0 &&
+		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
+		if (rts.device) {
+			asprintf(&target, "%s%%%s", target, rts.device);
+			rts.device = NULL;
+			free_target = 1;
+		} else if (sock6.socktype == SOCK_DGRAM) {
+			error(0, 0, _("Warning: IPv6 link-local address requires scope-id or ifname => use address%%<ifname|scope-id> or address -I<ifname|scope-id>"));
+		}
+	}
+
 	ret_val = getaddrinfo(target, NULL, &hints, &result);
 	if (ret_val)
 		error(2, 0, "%s: %s", target, gai_strerror(ret_val));
@@ -590,6 +608,9 @@ main(int argc, char **argv)
 
 	freeaddrinfo(result);
 	free(rts.outpack);
+
+	if (free_target)
+		free(target);
 
 	return ret_val;
 }
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -280,10 +280,17 @@ xml:id="man.ping">
           name, each packet is routed using the corresponding
           routing table; in this case, the <option>-I</option> option
           can be repeated to specify a source address.
-          NOTE: For IPv6, when doing ping to a link-local scope
-          address, link specification (by the '%'-notation in
-          <emphasis remap="I">destination</emphasis>, or by this
-          option) can be used but it is no longer required.</para>
+          </para>
+
+          <para>
+          NOTE: For ping to a IPv6 link-local scope address when using ICMP socket
+          (not required on legacy raw socket), link specification is required on
+          one of following forms:
+          <emphasis remap="I">destination%&lt;ifname|scope-id&gt;</emphasis> or
+          <emphasis remap="I">destination -I&lt;ifname|scope-id&gt;</emphasis>,
+          where &lt;ifname&gt; is interface name, &lt;scope-id&gt; is it's index.
+          E.g. fe80::fc54:ff:fe9e:936c -Ivnet2 or fe80::fc54:ff:fe9e:936c%20.
+          </para>
         </listitem>
       </varlistentry>
       <varlistentry>


pevik:
@nmeyerhans @pavlix @kerolasa could you please have a look?

ping/ping.c
```diff
@@ -555,6 +555,23 @@ main(int argc, char **argv)
...
 
+	/*
+	 * Allow to use -I<ifname|scope-id> for IPv6 link-local scope address
+	 * when passing it to getaddrinfo():
+	 * i.e. -I<ifname|scope-id> => %<ifname|scope-id>
+	 */
+	unsigned char buf[sizeof(struct in6_addr)];
+	if (!strchr(target, '%') && inet_pton(AF_INET6, target, buf) > 0 &&
+		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
+		if (rts.device) {
```
pevik, line 566: @jonharrisonms @gobenji @nmeyerhans I haven't tested this with VRF (9e08707d743b29e853df81bd7def1729e3afe55d, 2705c8248281fbb8efaa5326ab1d0ed0a670bd3d).

It looks like basic getopts parsing for device works at least for `-I <source_addr> -I <ifname>`:
https://github.com/iputils/iputils/blob/854873bdd28fcdd9cc3fe0c2d29c083a07d07a86/ping/ping.c#L388-L413

because all of these works (`2620:113:80c0:8000:10:161:28:49` is local IPv6 global scope dynamic noprefixroute, `fe80::5054:ff:fe55:be42` is remote IPv6 link scope) - verified with `tcpdump`:
```
ping fe80::5054:ff:fe55:be42%eth0 -I 2620:113:80c0:8000:10:161:28:49
ping fe80::5054:ff:fe55:be42 -I eth0 -I 2620:113:80c0:8000:10:161:28:49
ping fe80::5054:ff:fe55:be42 -I 2620:113:80c0:8000:10:161:28:49 -I eth0
```
(also verified it works with this change).

Also maybe I should add here also check whether `rts.device` and not apply it when invalid ifname, not sure if new functionality is not misleading:
```
$ ping fe80::5054:ff:fe55:be42 -I noexisting # original
ping: unknown iface: noexisting

$ ./builddir/ping/ping fe80::5054:ff:fe55:be42 -I noexisting # with this change
./builddir/ping/ping: fe80::5054:ff:fe55:be42%noexisting: Name or service not known
```

pevik, line 566: @jonharrisonms @gobenji @nmeyerhans I haven't tested this with VRF (9e08707d743b29e853df81bd7def1729e3afe55d, 2705c8248281fbb8efaa5326ab1d0ed0a670bd3d).

It looks like basic getopts parsing for device works at least for `-I <source_addr> -I <ifname>`:
https://github.com/iputils/iputils/blob/854873bdd28fcdd9cc3fe0c2d29c083a07d07a86/ping/ping.c#L388-L413

because all of these works (`2620:113:80c0:8000:10:161:28:49` is local IPv6 global scope dynamic noprefixroute, `fe80::5054:ff:fe55:be42` is remote IPv6 link scope) - verified with `tcpdump`:
```
ping fe80::5054:ff:fe55:be42%eth0 -I 2620:113:80c0:8000:10:161:28:49
ping fe80::5054:ff:fe55:be42 -I eth0 -I 2620:113:80c0:8000:10:161:28:49
ping fe80::5054:ff:fe55:be42 -I 2620:113:80c0:8000:10:161:28:49 -I eth0
```
(also verified it works with this change).

Also maybe I should add here also check whether `rts.device` and not apply it when invalid ifname, not sure if new functionality is not misleading:
```
$ ping fe80::5054:ff:fe55:be42 -I noexisting # original
ping: unknown iface: noexisting

$ ./builddir/ping/ping fe80::5054:ff:fe55:be42 -I noexisting # with this change
./builddir/ping/ping: fe80::5054:ff:fe55:be42%noexisting: Name or service not known
```

ping/ping.c
```diff
@@ -555,6 +555,23 @@ main(int argc, char **argv)
...
+			strcat(target, "%");
+			strcat(target, rts.device);
```
pevik, line 568: As @metan-ucw reported using `strcat()` is wrong because  of previously `target = argv[argc - 1]`. There is no space to append anything to it. Although it "works" I probably rewrote some env variables passed to the process by kernel or something. I should allocate the buffer, e.g. use `asprintf()` which prints into a newly allocated buffer, but that would require to make `_GNU_SOURCE` mandatory. I wonder if it's ok (we have meson check for it).

nmeyerhans: I think `asprintf()` should be available on all the libc implementations we're likely to be linking against (glibc, MUSL, uClibc, anyway).  However, I think `asprintf()` would be wasteful due to the amount of memory copying it would need to do. We could instead do something like the following (untested) and only allocate memory once:

```C
		char *target;
		if (rts.device) {
			target = (char*)malloc(strlen(rts.device) + strlen(argv[argc - 1]) + 2);
			sprintf(target, "%s%%%s", argv[argc - 1], rts.device);
		} else {
			target = (char*)malloc(strlen(argv[argc - 1]) + 1);
			strcat(target, argv[argc - 1]);
		}
```

Not that it matters significantly for the amount of memory we're talking about here.

pevik: I don't get why [`asprintf()`](https://man7.org/linux/man-pages/man3/asprintf.3.html) should allocate more memory than `malloc()` + `sprintf()`. 

Actually @metan-ucw suggested to use [his `asprintf()` implementation](https://github.com/gfxprim/gfxprim/blob/master/libs/widgets/gp_string.c#L43) (requires also [proper signature](https://github.com/gfxprim/gfxprim/blob/master/include/widgets/gp_string.h#L39)) in case GNU extension is a problem (I don't think so but it's nice to have a backup).

IMHO this would be needed only `if (rts.device)`, when we modify `target`, right? By default we could just reuse `argv[argc - 1]`.

pevik: Also, we already depend on GNU extensions - building without it results to many failures in `ping` and `arping`. Just to list some of them:
```
../ping/ping.h:319:9: warning: implicit declaration of function ‘sigaction’ [-Wimplicit-function-declaration]
  319 |         sigaction(signo, &sa, NULL);
../ping/ping.c: In function ‘pr_addr’:
../ping/ping.c:1630:19: error: ‘NI_MAXHOST’ undeclared (first use in this function)
 1630 |         char name[NI_MAXHOST] = "";
      |                   ^~~~~~~~~~
../ping/ping.c:1641:9: warning: implicit declaration of function ‘getnameinfo’ [-Wimplicit-function-declaration]
 1641 |         getnameinfo(sa, salen, address, sizeof address, NULL, 0, getnameinfo_flags | NI_NUMERICHOST);

../arping.c:289:9: warning: implicit declaration of function ‘clock_gettime’ [-Wimplicit-function-declaration]
  289 |         clock_gettime(CLOCK_MONOTONIC, &now);
      |         ^~~~~~~~~~~~~
../arping.c:592:36: error: ‘IFF_LOOPBACK’ undeclared (first use in this function); did you mean ‘IN_LOOPBACKNET’?
  592 |         if (ifflags & (IFF_NOARP | IFF_LOOPBACK)) {
      |                                    ^~~~~~~~~~~~
      |                                    IN_LOOPBACKNET
../arping.c: In function ‘event_loop’:
../arping.c:731:9: warning: implicit declaration of function ‘sigemptyset’ [-Wimplicit-function-declaration]
  731 |         sigemptyset(&mask);
      |         ^~~~~~~~~~~
../arping.c:732:9: warning: implicit declaration of function ‘sigaddset’ [-Wimplicit-function-declaration]
  732 |         sigaddset(&mask, SIGINT);
../iputils_common.c:92:25: error: storage size of ‘ts’ isn’t known
   92 |         struct timespec ts;
      |                         ^~
../iputils_common.c:94:9: warning: implicit declaration of function ‘clock_gettime’ [-Wimplicit-function-declaration]
   94 |         clock_gettime(CLOCK_REALTIME, &ts);
      |         ^~~~~~~~~~~~~
../iputils_common.c:94:23: error: ‘CLOCK_REALTIME’ undeclared (first use in this function)
   94 |         clock_gettime(CLOCK_REALTIME, &ts);
```
=> we should depend on GNU anyway s these are old changes (e.g. ad0256f3543a69ff0d44bca838bade6aac04bcbd - released in [s20200821](https://github.com/iputils/iputils/releases/tag/s20200821) - I'll fix this as well).

gobenji:
With this patch, because rts.device is cleared in some cases, a SO_BINDTODEVICE may be skipped. This can break some usecases. Using the configuration from commit 7c65999f98bc ("ping: Fix ping6 binding to VRF and address"):
```
root@vsid:~# ip netns exec temp_netns ping fe80::5c68:9dff:feed:e586 -I lo10 -I 1:2::3:4 -c1
PING fe80::5c68:9dff:feed:e586(fe80::5c68:9dff:feed:e586) from 1:2::3:4 lo10: 56 data bytes
64 bytes from fe80::5c68:9dff:feed:e586%vrf_1: icmp_seq=1 ttl=64 time=0.025 ms

--- fe80::5c68:9dff:feed:e586 ping statistics ---
rtt min/avg/max/mdev = 0.025/0.025/0.025/0.000 ms
root@vsid:~# ip netns exec temp_netns /src/iputils/build/ping/ping fe80::5c68:9dff:feed:e586 -I lo10 -I 1:2::3:4 -c1
/src/iputils/build/ping/ping: bind icmp socket: Cannot assign requested address
```

I wouldn't bet that this is a common usecase but it's something to consider.

Basically, the objective of this patch is to enable `-I <scope-id>` when the destination is an ipv6 link-local address, right?

pevik:
> With this patch, because rts.device is cleared in some cases, a SO_BINDTODEVICE may be skipped. This can break some usecases. 

Good catch, I need to have look into it!

> Using the configuration from commit [7c65999](https://github.com/iputils/iputils/commit/7c65999f98bc4a1984594b7fad1af0eaf0b9d34b) ("ping: Fix ping6 binding to VRF and address"):
> 
> ```
> root@vsid:~# ip netns exec temp_netns ping fe80::5c68:9dff:feed:e586 -I lo10 -I 1:2::3:4 -c1
> PING fe80::5c68:9dff:feed:e586(fe80::5c68:9dff:feed:e586) from 1:2::3:4 lo10: 56 data bytes
> 64 bytes from fe80::5c68:9dff:feed:e586%vrf_1: icmp_seq=1 ttl=64 time=0.025 ms
> 
> --- fe80::5c68:9dff:feed:e586 ping statistics ---
> 1 packets transmitted, 1 received, 0% packet loss, time 0ms
> rtt min/avg/max/mdev = 0.025/0.025/0.025/0.000 ms
> root@vsid:~# ip netns exec temp_netns /src/iputils/build/ping/ping fe80::5c68:9dff:feed:e586 -I lo10 -I 1:2::3:4 -c1
> /src/iputils/build/ping/ping: bind icmp socket: Cannot assign requested address
> ```
> 
> I wouldn't bet that this is a common usecase but it's something to consider.

Sure, no regression.
> 
> Basically, the objective of this patch is to enable `-I <scope-id>` when the destination is an ipv6 link-local address, right?

Yes. I wasn't sure if it's a good idea, but it'd be better to not force to use `%` notation when we wrongly advertised some time ago it's possible.

gobenji:
> > Basically, the objective of this patch is to enable `-I <scope-id>` when the destination is an ipv6 link-local address, right?
> 
> Yes. I wasn't sure if it's a good idea, but it'd be better to not force to use `%` notation when we wrongly advertised some time ago it's possible.

```
Fixes: 243b6e1 ("ping: refactor ping socket code")
Fixes: 9f26885 ("doc/ping: Update Interface section")
Fixes: https://github.com/iputils/iputils/issues/398
```
To me, it seems more like a new feature than a fix. The documentation never said that you could do `-I <scope-id>`, right?

From my point of view, `-I` is about binding the socket whereas `%` controls the scope of the destination address. I don't think it's a good idea to further couple the two and rewrite one option into the other. It seems risky to introduce unintended behavior changes like the one I showed above.

The "scope-id" is basically the same as "ifindex". If you want to add support for `-I <ifindex>`, maybe it can be implemented in a way that's limited to the parsing of that option.

pevik:
> To me, it seems more like a new feature than a fix. The documentation never said that you could do `-I <scope-id>`, right?

Yes, this is a new feature. I added `Fixes:` before just wrongly claimed in man page in commit 9f26885ba120dc68c9168f128b241d1281950de2 that `%` is not needed. It's not needed for the old approach (when using raw socket), but that it's required for recommended use (ping socket). ATM not sure why.

> From my point of view, `-I` is about binding the socket whereas `%` controls the scope of the destination address. I don't think it's a good idea to further couple the two and rewrite one option into the other. It seems risky to introduce unintended behavior changes like the one I showed above.

Agree, as I wrote I'm not sure myself it's a good idea to implement it. The main motivation was that people can mix up these two. But encouraging to use `-I` for both source and destination address does not right to me.

Also `-I` is a bit complex even now, we allow to use `-I <ifname> -I <VRF>`, not sure if it even makes sense to have
`-I <ifname> -I <scope_id>` or even `-I <ifname> -I <scope_id> -I <VRF>`.

Maybe to have extra option would be better than trying this (option is more visible than `%` notation). But instead of adding new option I'll probably fix the docs and make it more visible.

> The "scope-id" is basically the same as "ifindex". If you want to add support for `-I <ifindex>`, maybe it can be implemented in a way that's limited to the parsing of that option.

Just for clarify, do you mean to handle this case in `getopt()` loop?

ping/ping.c
nmeyerhans: > I don't get why [asprintf()](https://man7.org/linux/man-pages/man3/asprintf.3.html) should allocate more memory than malloc() + sprintf().

It'll allocate memory each time you call it, and as your code was written you were going to end up allocating memory twice: once for the copy of the target from `argv`, and again to create a new, larger, buffer to hold the target with `%IFNAME` appended.

But really the amount of memory isn't critical here. I was more trying to avoid using `_GNU_SOURCE`. But I think it's supported by all the libc implementations we care about. (glibc and MUSL support it, as does uClibc afaik. Are there others?)

pevik: > I was more trying to avoid using `_GNU_SOURCE`

I'm sorry I didn't wait for you before merge https://github.com/iputils/iputils/pull/402, but we have really depended on `_GNU_SOURCE` for quite long time

> ut I think it's supported by all the libc implementations we care about. (glibc and MUSL support it, as does uClibc afaik. Are there others?)

others: [dietlibc](https://www.fefe.de/dietlibc/) looks to be dead. [μClinux](https://en.wikipedia.org/wiki/%CE%9CClinux) as well.

Android moved some time ago to use iputils s20121221 (and they didn't upgrade), thus we can ignore it, but it also supports asprintf()
https://android.googlesource.com/platform/bionic.git/+/refs/heads/android-t-preview-1/libc/stdio/stdio.cpp#736
https://android.googlesource.com/platform/external/ping/+/refs/heads/master
https://android.googlesource.com/platform/external/iputils/+/refs/heads/android-t-preview-1/README.version

pevik:
@nmeyerhans @pavlix  ATM I'm more for not implementing `%` -> `-I` replacement and rather state the need in the docs 
(https://github.com/iputils/iputils/pull/399#issuecomment-1072386701). WDYT?

gobenji:
> > The "scope-id" is basically the same as "ifindex". If you want to add support for `-I <ifindex>`, maybe it can be implemented in a way that's limited to the parsing of that option.
> 
> Just for clarify, do you mean to handle this case in `getopt()` loop?

Yes, that's what I was thinking. But to be clear, I don't personally see a need to add support for `-I <ifindex>`.

pevik:
> > Just for clarify, do you mean to handle this case in `getopt()` loop?
> 
> Yes, that's what I was thinking. But to be clear, I don't personally see a need to add support for `-I <ifindex>`.

Agree. @nmeyerhans unless you're strongly for implementing `-I <ifindex>` I'll just fix man page where I state `%` is required (for icmp socket, which is the default).

I guess the main confusion to use `-I` for it is that `%` is mentioned in `-I` section and there is no description for `destination`. Maybe we should create it and mention `%` notation there.

Also it'd be worth to mention in man the differences between raw and icmp socket use, but that's another story.

pevik:
BTW if we chose only to fix man page, it'd also make sense to add warning from this PR
`error(0, 0, _("Warning: IPv6 link-local address requires scope-id or ifname => use address%%<ifname|scope-id> or address -I<ifname|scope-id>"));`

(obviously ` or address -I<ifname|scope-id>` part would be removed)

pevik:
Superseded by https://github.com/iputils/iputils/pull/405

PR #404, opened 2022-03-18
arping: Fix exit code on -w option when count * interval > timeout

eaglegai:
Signed-off-by: eaglegai <eaglegai@163.com>

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -850,7 +850,7 @@ static int event_loop(struct run_state *ctl)
 	else if (ctl->dad && ctl->quit_on_reply)
 		/* Duplicate address detection mode return value */
 		rc |= !(ctl->brd_sent != ctl->received);
-	else if (ctl->timeout && !(ctl->count > 0))
+	else if (ctl->timeout && (!(ctl->count > 0) || (ctl->interval > ctl->timeout / ctl->count)))
 		rc |= !(ctl->received > 0);
 	else
 		rc |= (ctl->sent != ctl->received);


eaglegai:
since commit 854873 , as long as we receive reply packages，and -w parameter is setted  &&  -c parameter is not setted，the return value is 0.
but if we use -c and -w together，and count value is larger than timeout value，in fact，timeout value will take effect，in this case it also should return 0  if we could receive  packages，no matter how many packages they are

eaglegai:
before:
root@localhost ~]# arping 99.99.75.56 -c 10 -w 2
ARPING 99.99.75.56 from 99.99.75.57 enp3s0
Unicast reply from 99.99.75.56 [x:x:x:x:x:x]  2.370ms
Unicast reply from 99.99.75.56 [x:x:x:x:x:x]  0.970ms
Sent 3 probes (1 broadcast(s))
Received 2 response(s)
[root@localhost ~]# echo $?
1

arping.c
```diff
@@ -850,7 +850,7 @@ static int event_loop(struct run_state *ctl)
 	else if (ctl->dad && ctl->quit_on_reply)
 		/* Duplicate address detection mode return value */
 		rc |= !(ctl->brd_sent != ctl->received);
-	else if (ctl->timeout && !(ctl->count > 0))
+	else if (ctl->timeout && (!(ctl->count > 0) || (ctl->count > ctl->timeout)))
```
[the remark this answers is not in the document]
eaglegai: oh, i didn't consider interval value, after considering interval value， comparing count * interval and timeout make sense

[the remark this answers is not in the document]
eaglegai: I change patch to compare  -w timeout and  total time to send the packages(-c count) ,when timeout is smaller than total time, timeout take effect，return 0 if we receive reply packages

pevik:
LGTM, but I'll need to have a closer look. Also your second commit should have been in the commit message (I'll update it before merge).

eaglegai:
> LGTM, but I'll need to have a closer look. Also your second commit should have been in the commit message (I'll update it before merge).

okay

eaglegai:
i just change the commit message to be consistent with the issue title

eaglegai:
@pevik hi, Is there any thing else that need to be modified?

eaglegai:
@pevik  change  'count * interval > timeout' to 'interval > timeout / count' to avoid integer overflow.

DenisKirjanov:
Looks like the commit message should be updates and then finally merged, right?

arping.c
```diff
@@ -850,7 +850,7 @@ static int event_loop(struct run_state *ctl)
 	else if (ctl->dad && ctl->quit_on_reply)
 		/* Duplicate address detection mode return value */
 		rc |= !(ctl->brd_sent != ctl->received);
-	else if (ctl->timeout && !(ctl->count > 0))
+	else if (ctl->timeout && (!(ctl->count > 0) || (ctl->interval > ctl->timeout / ctl->count)))
```
pevik, line 853: Maybe instead of `ctl->interval > ctl->timeout / ctl->count` express it as: `ctl->interval * ctl->count > ctl->timeout`? IMHO slightly more readable.

pevik: ah, valid point

pevik: OK, both variants are wrong:
* `ctl->interval > ctl->timeout / ctl->count` division of two `int` is ceiled: `2/3 = 0`. It needs to be casted to `float` or `double`: 
```
2/3.0 = 0.666667
2/(float)3 = 0.666667
```
Also `float` is not precise, for higher values `double` is needed:
```
2147483647 / (float)2: 1073741824.00 // wrong
2147483647 / (double)2: 1073741823.50// correct
```

* `ctl->interval * ctl->count > ctl->timeout` would indeed have overflow if the result is bigger than `INT_MAX` (i.e. `2147483647`). But again, casting will fix it.

Therefore either of these is correct, I'd slightly prefer the second one (multiplication with casting to `long`):
```
ctl->interval > ctl->timeout / (double) ctl->count
(long)ctl->interval * ctl->count > ctl->timeout
```
@metan-ucw @gobenji @DenisKirjanov feel free to correct me.

gobenji: 1)
```
(long)ctl->interval * ctl->count > ctl->timeout
```
Are 32bit arches still supported? If so, then a type like uint64_t would be needed rather than `long`.

2)
In commit 854873bdd28f Noah indicated that the goal was to be compatible with the exit status of ping. AFAIK, ping has the same behavior as what's desribed here for arping, ie. if count > deadline: exit status != 0:
```
root@vsid:/src/iputils/builddir# ./ping/ping -w2 -c3 192.168.15.1
PING 192.168.15.1 (192.168.15.1) 56(84) bytes of data.
64 bytes from 192.168.15.1: icmp_seq=1 ttl=64 time=0.052 ms
64 bytes from 192.168.15.1: icmp_seq=2 ttl=64 time=0.212 ms

--- 192.168.15.1 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1026ms
rtt min/avg/max/mdev = 0.052/0.132/0.212/0.080 ms
root@vsid:/src/iputils/builddir# echo $?
1

root@vsid:/src/iputils/builddir# ./arping -w2 -c3 192.168.15.1
ARPING 192.168.15.1 from 192.168.15.156 eth0
Unicast reply from 192.168.15.1 [3A:2E:EC:7B:AE:3F]  0.573ms
Unicast reply from 192.168.15.1 [3A:2E:EC:7B:AE:3F]  0.720ms
Sent 3 probes (1 broadcast(s))
Received 2 response(s)
root@vsid:/src/iputils/builddir# echo $?
1
```
Do you really want to take in this change and make arping different from ping in this regard?

eaglegai: > OK, both variants are wrong:
> 
> * `ctl->interval > ctl->timeout / ctl->count` division of two `int` is ceiled: `2/3 = 0`. It needs to be casted to `float` or `double`:
> 
> ```
> 2/3.0 = 0.666667
> 2/(float)3 = 0.666667
> ```
> 
> Also `float` is not precise, for higher values `double` is needed:
> 
> ```
> 2147483647 / (float)2: 1073741824.00 // wrong
> 2147483647 / (double)2: 1073741823.50// correct
> ```
> 
> * `ctl->interval * ctl->count > ctl->timeout` would indeed have overflow if the result is bigger than `INT_MAX` (i.e. `2147483647`). But again, casting will fix it.
> 
> Therefore either of these is correct, I'd slightly prefer the second one (multiplication with casting to `long`):
> 
> ```
> ctl->interval > ctl->timeout / (double) ctl->count
> (long)ctl->interval * ctl->count > ctl->timeout
> ```
> 
> @metan-ucw @gobenji @DenisKirjanov feel free to correct me.

1. when interval >= 1, 2/3 = 0 or 2/3.0 = 0.666667 will not have any effect on the result, ctl->timeout / ctl->count needn't cast
2. when interval == 0, it will just send and receive one package, and $? == 0 (i don't know if send one package is reasonable,which is different from ping)
before patch,i test like this:
interval    timeout    count   echo$?     sendpackages   receivepackages
3              30            10          0            10                    10
3              29            10          0            10                    10
3              28            10          0            10                    10
3              27            10          1            10                    9
3              26            10          0            9                      9
3              25            10          0            9                      9
3              24            10          1            9                      8
3              9             10           1            4                      3
0              24            10          0            1                      1
after patch,  all $? == 0
3. we can make the reproduce range more precise，if （timeout%interval == 0）&& （interval * count >timeout）, then $? != 0 
4. since 97926373，arping add -i option，which logic is very different from ping，for example, ping's interval allowed with dot as a decimal separator, so is it necessary to keep them consistent?

pevik: > Are 32bit arches still supported? If so, then a type like `int64_t` would be needed rather than `long`.

I'd keep it. Last intel atom processors were sold few years ago. I wanted to postpone types cleanup after the release: https://github.com/iputils/iputils/issues/410

pevik: > > In commit [854873b](https://github.com/iputils/iputils/commit/854873bdd28fcdd9cc3fe0c2d29c083a07d07a86) Noah indicated that the goal was to be compatible with the exit status of ping. AFAIK, ping has the same behavior as what's desribed here for arping, ie. if count > deadline: exit status != 0:
> > ...
> > Do you really want to take in this change and make arping different from ping in this regard?

> since https://github.com/iputils/iputils/commit/97926373401e4e794fa90f87b42c6cac9c35daf7 arping add `-i` option, which logic is very different from ping, for example, ping's interval allowed with dot as a decimal separator, so is it necessary to keep them consistent?

I need to have a closer look, but generally I'd prefer consistency:
1) `arping` has long history, we should be careful when introducing unnecessary changes
2) there is alternative implementation in busybox project, again unnecessary changes breaks compatibility (bad for scripts aiming portability between these two)

eaglegai:
hello, is there any thing new about this issue?@pevik

pevik:
After long thinking I prefer to be compatible with ping, thus not accepting this. Actually I'm thinking to to make busybox arping implementation to behave the same as our implementation (their ping also follows our behavior).
I also have look at `-i` differences in `ping` and `arping` after release.
I'll moved using fixed data types (https://github.com/iputils/iputils/issues/410) to after future.

Issue #104, opened 2017-10-11
ping -I device_name device_ip_address (self) does not work


jsynacek:
It's a rather cosmetic issue, but still...
```
# ip addr show dev eth0 | grep inet
    inet 192.168.122.100/24 brd 192.168.122.255 scope global eth0
    inet6 fe80::8817:157e:68cc:d912/64 scope link 

# ping -I eth0 192.168.122.100
PING 192.168.122.100 (192.168.122.100) from 192.168.122.100 eth0: 56(84) bytes of data.
<nothing... loops forever>

# ping -I 192.168.122.100 192.168.122.100
PING 192.168.122.100 (192.168.122.100) from 192.168.122.100 : 56(84) bytes of data.
64 bytes from 192.168.122.100: icmp_seq=1 ttl=64 time=0.029 ms
64 bytes from 192.168.122.100: icmp_seq=2 ttl=64 time=0.048 ms
64 bytes from 192.168.122.100: icmp_seq=3 ttl=64 time=0.024 ms
<works just fine>
```
Note that using the device name when pinging the IP address of the same device results in ping not working, whereas using the device IP address instead of its name does work.

This is a result of https://github.com/iputils/iputils/pull/56. I didn't really find a problem with the code itself. The described problem seems to happen because of the `setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, device, strlen(device)+1);` is called on the `sock->fd` in addition to the probe fd. I'm not sure if this is a problem in the kernel, or in iputils code, or it's simply not meant to be used like that.

kerolasa:
I guess cancelling socket option when it does not work is an option. See commit 4082ba9e780320e3c6e31a21c7d7ad164c229b5b.

kerolasa:
Merged: cc44f4c5c2e72f7aa833658a6978d924ff9e059d

threexc:
Hello,

I am encountering an issue similar to this using release s20190709, but with IPv6. I've confirmed that the packets are actually appearing on the loopback interface with tcpdump, and that specifying the device as e.g. ping -6 fe80::92e2:baff:fe08:3738%eth0 works, but ping -6 -I eth0 fe80::92e2:baff:fe08:3738 does not. Both ping options work for me on s20180629. I've also tried writing a patch similar to [cc44f4c](https://github.com/iputils/iputils/commit/cc44f4c5c2e72f7aa833658a6978d924ff9e059d), but it didn't affect the result.

[...]

PR #405, opened 2022-03-29
ping: document '%'-notation for IPv6 link-local scope address, print warning, man: add destination

pevik:
Man reason for this commit were problems described in https://github.com/iputils/iputils/issues/398.

But it looks like '%'-notation should not have an alternative in `-I<ifname|scope-id>` option, therefore in this PR:
* man/ping: document %-notation for IPv6 link-local scope address
* ping: print warning when missing on non-raw ICMP socket
* man/{ar,}ping: add destination section

### DIFF
--- a/doc/arping.xml
+++ b/doc/arping.xml
@@ -58,6 +58,11 @@ xml:id="man.arping">
     <emphasis remap="I">interface</emphasis> by ARP packets, using
     source address
     <emphasis remap="I">source</emphasis>.</para>
+    <para>arping supports IPv4 addresses only. For IPv6, see
+    <citerefentry>
+      <refentrytitle>ndisc6</refentrytitle>
+      <manvolnum>8</manvolnum>
+    </citerefentry>.</para>
   </refsection>
 
   <refsection>
@@ -227,6 +232,10 @@ xml:id="man.arping">
   <refsect1 id='see_also'>
     <title>SEE ALSO</title>
     <para>
+    <citerefentry>
+      <refentrytitle>ndisc6</refentrytitle>
+      <manvolnum>8</manvolnum>
+    </citerefentry>,
     <citerefentry>
       <refentrytitle>ping</refentrytitle>
       <manvolnum>8</manvolnum>
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -923,7 +923,7 @@ xml:id="man.ping">
     <command>ping</command> requires CAP_NET_RAW capability to be
     executed 1) if the program is used for non-echo queries (See
     <option>-N</option> option), or 2) if kernel does not support
-    non-raw ICMP sockets, or 3) if the user is not allowed to
+    ICMP datagram sockets, or 3) if the user is not allowed to
     create an ICMP echo socket. The program may be used as set-uid
     root.</para>
   </refsection>
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -120,6 +120,46 @@ xml:id="man.ping">
 
   <refsection>
     <info>
+      <title>IPV6 LINK-LOCAL DESTINATIONS</title>
+    </info>
+    <para>For IPv6, when the destination address has link-local scope and
+    ping is using <emphasis remap="I">ICMP datagram sockets</emphasis>,
+    the output interface must be specified.
+    When <command>ping</command> is using raw sockets, it is not strictly
+    necessary to specify the output interface but it should be done to avoid
+    ambiguity when there are multiple possible output interfaces.</para>
+    <para>There are two ways to specify the output interface:</para>
+    <variablelist remap="TP">
+      <varlistentry>
+        <term>
+          • using the
+          <emphasis remap="I">% notation</emphasis>
+        </term>
+        <listitem>
+        <para>The destination address is postfixed with
+        <emphasis remap="I">%</emphasis>
+         and the output interface name or ifindex, for example:</para>
+        <para><command>ping fe80::5054:ff:fe70:67bc%eth0</command></para>
+        <para><command>ping fe80::5054:ff:fe70:67bc%2</command></para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+          • using the
+          <emphasis remap="I">-I option</emphasis>
+        </term>
+        <listitem>
+          <para>When using <emphasis remap="I">ICMP datagram sockets</emphasis>,
+          this method is only supported on some kernel versions: 5.17,
+          5.15.19, 5.10.96, 5.4.176, 4.19.228, 4.14.265.
+          Also it is not supported on musl libc.</para>
+        </listitem>
+      </varlistentry>
+    </variablelist>
+  </refsection>
+
+  <refsection>
+    <info>
       <title>OPTIONS</title>
     </info>
     <variablelist remap="TP">
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -558,6 +558,15 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
+	unsigned char buf[sizeof(struct in6_addr)];
+	if (!strchr(target, '%') && sock6.socktype == SOCK_DGRAM &&
+		inet_pton(AF_INET6, target, buf) > 0 &&
+		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
+			error(0, 0, _(
+				"Warning: IPv6 link-local address on ICMP datagram socket may require ifname or scope-id"
+				" => use: address%%<ifname|scope-id>"));
+	}
+
 	ret_val = getaddrinfo(target, NULL, &hints, &result);
 	if (ret_val)
 		error(2, 0, "%s: %s", target, gai_strerror(ret_val));
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -44,7 +44,7 @@
  *	Public Domain.  Distribution Unlimited.
  * Bugs -
  *	More statistics could always be gathered.
- *	If kernel does not support non-raw ICMP sockets,
+ *	If kernel does not support ICMP datagram sockets,
  *	this program has to run SUID to ROOT or with
  *	net_cap_raw enabled.
  */
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -53,7 +53,7 @@
  *	Public Domain.  Distribution Unlimited.
  * Bugs -
  *	More statistics could always be gathered.
- *	If kernel does not support non-raw ICMP sockets or
+ *	If kernel does not support ICMP datagram sockets or
  *	if -N option is used, this program has to run SUID to ROOT or
  *	with net_cap_raw enabled.
  */


pevik:
@gobenji could you please review it?

doc/ping.xml
```diff
@@ -107,15 +107,27 @@ xml:id="man.ping">
...
     <command>ping</command> works with both IPv4 and IPv6. Using
     only one of them explicitly can be enforced by specifying
     <option>-4</option> or
     <option>-6</option>.</para>
     <para>
-    <command>ping</command> can also send IPv6 Node Information
-    Queries (RFC4620). Intermediate
-    <emphasis remap="I">hop</emphasis>s may not be allowed, because
-    IPv6 source routing was deprecated (RFC5095).</para>
+    NOTE: For when doing ping to a IPv6 link-local scope
```
gobenji, line 126: I'm not a native English speaker but it seems that the previous formulation ("For IPv6, when doing ping [...]") was more natural.

ping/ping.c
```diff
@@ -557,6 +557,15 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
+	unsigned char buf[sizeof(struct in6_addr)];
+	if (!strchr(target, '%') && sock6.socktype == SOCK_DGRAM &&
+		inet_pton(AF_INET6, target, buf) > 0 &&
+		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
+			error(0, 0, _(
+				"Warning: IPv6 link-local address on non-raw ICMP socket requires ifname or scope-id"
+				" => use address%%<ifname|scope-id>"));
```
gobenji, line 566: The warning is appreciated but not completely accurate:
```
root@vsid:/src/iputils# ./builddir/ping/ping -I eth0 fe80::1 -c1
./builddir/ping/ping: Warning: IPv6 link-local address on non-raw ICMP socket requires ifname or scope-id => use address%<ifname|scope-id>
./builddir/ping/ping: Warning: source address might be selected on device other than: eth0
PING fe80::1(fe80::1) from :: eth0: 56 data bytes
64 bytes from fe80::1%eth0: icmp_seq=1 ttl=64 time=0.090 ms

--- fe80::1 ping statistics ---
rtt min/avg/max/mdev = 0.090/0.090/0.090/0.000 ms
```
ie. it requires `%` **or** `-I`

doc/ping.xml
```diff
@@ -107,15 +107,27 @@ xml:id="man.ping">
...
+    <emphasis remap="I">destination</emphasis>, non-raw ICMP socket
```
gobenji, line 127: "non-raw ICMP socket" is more complicated than necessary. There are no "stream icmp sockets" so it's not necessary to say "non-raw". I would say "ICMP datagram socket". That is closer to the [kernel documentation](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt) for ping_group_range:

> ping_group_range - 2 INTEGERS
> 	Restrict ICMP_PROTO datagram sockets to users in the group range.

Same comment for the message in the second patch.

pevik: FYI I used that because `non-raw ICMP sockets` term was used already in man page:

https://github.com/iputils/iputils/blob/master/doc/ping.xml#L923-L928

But you're right, thus I use `ICMP datagram socket` as you suggested. Thanks!

ping/ping.c
pevik: 1) while this works on some of my x86_64 machine, it's a problem on ppc64le
2) not mixing `-I ifname` with % notation was the reason to not taking https://github.com/iputils/iputils/issues/398, right? Do we want to reconsider it? With fixing `-I` parsing in getopts it would save VRF issue. That PR fixes ppc64le issues to make sure `-I ifname` works everywhere, even `-I ifindex` works with it).

*  `-I ifname` works on x86_64 machine:
```
$ ip -6 a show vnet0
7: vnet0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master virbr0 state UNKNOWN group default qlen 1000
    inet6 fe80::fc54:ff:feeb:ac2f/64 scope link 
       valid_lft forever preferred_lft forever

$ ip -6 a show wlp2s0 
2: wlp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    inet6 fe80::5054:ff:febb:a3d4/64 scope link 
       valid_lft forever preferred_lft forever
    inet6 fe80::766a:125a:c587:ab18/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```
*  `-I ifname` broken on ppc64le machine:
```
$ ip -6 a show eth0 
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    altname enp0s1
    inet6 XXXX:XXX:XXXX:XXXX:XX:XXX:XX:XX/64 scope global dynamic noprefixroute 
       valid_lft 11374sec preferred_lft 5974sec
    inet6 fe80::5054:ff:febb:a3d4/64 scope link 
       valid_lft forever preferred_lft forever

# global IPv6 hidden

$ ./builddir/ping/ping fe80::5054:ff:febb:a3d4 -I eth0
./builddir/ping/ping: Warning: source address might be selected on device other than: eth0
PING fe80::5054:ff:febb:a3d4(fe80::5054:ff:febb:a3d4) from :: eth0: 56 data bytes

--- fe80::5054:ff:febb:a3d4 ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 2062ms
```

Obviously % notation works (both ifname and ifindex):
```
$ ./builddir/ping/ping -c1 fe80::5054:ff:febb:a3d4%eth0
PING fe80::5054:ff:febb:a3d4%eth0(fe80::5054:ff:febb:a3d4%eth0) 56 data bytes
64 bytes from fe80::5054:ff:febb:a3d4%eth0: icmp_seq=1 ttl=64 time=0.085 ms

--- fe80::5054:ff:febb:a3d4%eth0 ping statistics ---
rtt min/avg/max/mdev = 0.085/0.085/0.085/0.000 ms

$ ./builddir/ping/ping -c1 fe80::5054:ff:febb:a3d4%2
PING fe80::5054:ff:febb:a3d4%2(fe80::5054:ff:febb:a3d4%eth0) 56 data bytes
64 bytes from fe80::5054:ff:febb:a3d4%eth0: icmp_seq=1 ttl=64 time=0.073 ms

--- fe80::5054:ff:febb:a3d4%2 ping statistics ---
rtt min/avg/max/mdev = 0.073/0.073/0.073/0.000 ms

```
I'm not sure whether it's ppc64le issue or issue with global IPv6 address affecting it. But deleting the global IPv6 address does not help.

I also added another iface on that ppc64le machine to test it, but it does not help
```
3: ltp_ns_veth2@if4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 fd00:1:1:1::2/64 scope global nodad 
       valid_lft forever preferred_lft forever
    inet6 fe80::98cd:89ff:fee9:23b4/64 scope link 
       valid_lft forever preferred_lft forever

$ ./builddir/ping/ping fe80::98cd:89ff:fee9:23b4 -I ltp_ns_veth2
./builddir/ping/ping: Warning: source address might be selected on device other than: ltp_ns_veth2
PING fe80::98cd:89ff:fee9:23b4(fe80::98cd:89ff:fee9:23b4) from :: ltp_ns_veth2: 56 data bytes

--- fe80::98cd:89ff:fee9:23b4 ping statistics ---
8 packets transmitted, 0 received, 100% packet loss, time 7282ms
```
Obviously there is nothing ppc64le specific in iputils, IMHO the issue is not caused by network setup. Thus it's probably some ppc64le specific kernel network internals (or glibc), but unlikely a bug. Maybe I should ask on linuxppc-dev and netdev ML.

gobenji: > * `-I ifname` broken on ppc64le machine:

Interesting. I was curious so I tried using qemu. `-I` worked for me:
```
root@debian:~# ping fe80::1
PING fe80::1(fe80::1) 56 data bytes
^C
--- fe80::1 ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 2041ms

root@debian:~# ping -I enp0s2 fe80::1 -c1
ping: Warning: source address might be selected on device other than: enp0s2
PING fe80::1(fe80::1) from :: enp0s2: 56 data bytes
64 bytes from fe80::1%enp0s2: icmp_seq=1 ttl=64 time=0.412 ms

--- fe80::1 ping statistics ---
rtt min/avg/max/mdev = 0.412/0.412/0.412/0.000 ms
root@debian:~# uname -m
ppc64le
root@debian:~# lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye
root@debian:~# ping -V
ping from iputils 20210202
```

pevik: @gobenji Thanks! Maybe it's openSUSE HW specific. Could you please post also kernel version (`uname -r`). And even a used config (or which config did you use)?

gobenji: I simply installed Debian Bullseye and changed the sysctl to enable ping sockets, no other special configuration:

```
root@debian:~# sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 0   10000
root@debian:~# uname -a
Linux debian 5.10.0-13-powerpc64le #1 SMP Debian 5.10.106-1 (2022-03-17) ppc64le GNU/Linux
root@debian:~# ip -6 addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enp0s2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 fd0e:224b:829:4050:5054:ff:feb7:5a00/64 scope global dynamic mngtmpaddr
       valid_lft 86111sec preferred_lft 14111sec
    inet6 fe80::5054:ff:feb7:5a00/64 scope link
       valid_lft forever preferred_lft forever
root@debian:~# ip -6 rou
::1 dev lo proto kernel metric 256 pref medium
fd0e:224b:829:4050::/64 dev enp0s2 proto kernel metric 256 expires 86105sec pref medium
fe80::/64 dev enp0s2 proto kernel metric 256 pref medium
default via fe80::8885:5803:fdbd:c285 dev enp0s2 proto ra metric 1024 expires 1505sec hoplimit 64 pref medium
root@debian:~# ping -I enp0s2 fe80::1 -c1
ping: Warning: source address might be selected on device other than: enp0s2
PING fe80::1(fe80::1) from :: enp0s2: 56 data bytes
64 bytes from fe80::1%enp0s2: icmp_seq=1 ttl=64 time=4.23 ms

--- fe80::1 ping statistics ---
rtt min/avg/max/mdev = 4.229/4.229/4.229/0.000 ms
```

pevik:
@gobenji Sorry, for very long delay, continue below to not get lost.

In the end I found two issues with ppc64le:
1) There is **kernel fix from v5.17-rc5 [35a79e64de29 ("ping: fix the dif and sdif check in ping_lookup")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=35a79e64de29e8d57a5989aac57611c0cd29e13e)**, **[link to lore](https://lore.kernel.org/netdev/ea03066bc7256ab86df8d3501f3440819305be57.1644988852.git.lucien.xin@gmail.com/)**. I tested it's backport from [5.15 LTS (5.15.25)](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v5.15.37&id=b9ff5667e116dc56fc60b1e8bf21039f73a2ab42) via buildroot.
Strange enough, the problem is ppc64le specific (it works at least on x86_64), where recvmsg() blocks on with EAGAIN until signaled:

```
sendmsg(4, {msg_name={sa_family=AF_INET6, sin6_port=htons(58), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "fe80::5054:ff:febb:a3d4", &sin6_addr), sin6_scope_id=0}, msg_namelen=28, msg_iov=[{iov_base="\200\0\0\0\0\0\0\2\237\316sb\0\0\0\0\36\16\f\0\0\0\0\0\20\21\22\23\24\25\26\27"..., iov_len=64}], msg_iovlen=1, msg_control=[{cmsg_len=36, cmsg_level=SOL_IPV6, cmsg_type=0x32}], msg_controllen=40, msg_flags=0}, 0) = 64
setitimer(ITIMER_REAL, {it_interval={tv_sec=0, tv_usec=0}, it_value={tv_sec=10, tv_usec=0}}, NULL) = 0
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EAGAIN (Resource temporarily unavailable)
recvmsg(4, {msg_namelen=128}, 0)        = -1 EINTR (Interrupted system call)
--- SIGALRM {si_signo=SIGALRM, si_code=SI_KERNEL} ---
rt_sigreturn({mask=[]})                 = -1 EINTR (Interrupted system call)
write(1, "\n", 1)                       = 1
write(1, "--- fe80::5054:ff:febb:a3d4 ping"..., 48) = 48
write(1, "2 packets transmitted, 0 receive"..., 65) = 65
write(1, "\n", 1)                       = 1
close(1)                                = 0
close(2)                                = 0
exit_group(1)                           = ?
+++ exited with 1 +++
```
It took me some time, because before I expected it's glibc issue and therefore tests various glibc versions, via buildroot, openSUSE and compiling from git.

I tested also ppc64el Debian bullseye 5.10.0-14-powerpc64le 5.10.113-1 (working because contains backport of this fix v5.10.102: 8dec3c4e7350) and 5.17.0-1-powerpc64le 5.17.3-1 bookworm/sid (also working as have the fix).

[35a79e64de29](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=35a79e64de29e8d57a5989aac57611c0cd29e13e) commit message correctly mentions it should have been fixed in user-space:

>  The issue was also reported on: https://github.com/iputils/iputils/issues/104
>  But fixed in iputils in a wrong way by not binding to device when destination IP is on device, and it will cause some of kselftests to fail, as Jianlin noticed.

=> **should we reconsider reverting cc44f4c ("ping: do not bind to device when destination IP is on device")?** ATM all mainline, stable and LTS kernels backported kernel fix (https://lore.kernel.org/lkml/?q=%22ping%3A+fix+the+dif+and+sdif+check+in+ping_lookup%22).

2) **musl fails even with backported fix** (tested on 5.15.33, which contains the fix; `Try again` is from [`gai_strerror()`](https://git.musl-libc.org/cgit/musl/tree/src/network/gai_strerror.c#n7)):
```
# sysctl -w net.ipv4.ping_group_range='0 2147483647'
# strace ping -c2 fe80::5054:ff:fe12:3456 -Ieth0
execve("/bin/ping", ["ping", "-c2", "fe80::5054:ff:fe12:3456", "-Ieth0"], 0x7ffd0a8225d8 /* 10 vars */) = 0
arch_prctl(ARCH_SET_FS, 0x7f305c139b48) = 0
set_tid_address(0x7f305c139f90)         = 121
brk(NULL)                               = 0x55c74e7e7000
brk(0x55c74e7e9000)                     = 0x55c74e7e9000
mmap(0x55c74e7e7000, 4096, PROT_NONE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x55c74e7e7000
mprotect(0x7f305c136000, 4096, PROT_READ) = 0
mprotect(0x55c74e7e4000, 4096, PROT_READ) = 0
geteuid()                               = 0
getuid()                                = 0
rt_sigprocmask(SIG_BLOCK, ~[RTMIN RT_1 RT_2], [], 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], NULL, 8) = 0
setresuid(-1, 0, -1)                    = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
getrandom(0x7ffce6a49c64, 4, GRND_NONBLOCK) = -1 EAGAIN (Resource temporarily unavailable)
getpid()                                = 121
getuid()                                = 0
rt_sigprocmask(SIG_BLOCK, ~[RTMIN RT_1 RT_2], [], 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], NULL, 8) = 0
setresuid(-1, 0, -1)                    = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP) = 3
socket(AF_INET6, SOCK_DGRAM, IPPROTO_ICMPV6) = 4
getuid()                                = 0
rt_sigprocmask(SIG_BLOCK, ~[RTMIN RT_1 RT_2], [], 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], NULL, 8) = 0
setresuid(-1, 0, -1)                    = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
open("/etc/hosts", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 5
fcntl(5, F_SETFD, FD_CLOEXEC)           = 0
read(5, "127.0.0.1\tlocalhost\n127.0.1.1\tbu"..., 1024) = 40
read(5, "", 1024)                       = 0
close(5)                                = 0
open("/etc/resolv.conf", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 5
fcntl(5, F_SETFD, FD_CLOEXEC)           = 0
read(5, "nameserver 10.0.2.3 # eth0\n", 248) = 27
read(5, "", 248)                        = 0
close(5)                                = 0
socket(AF_INET, SOCK_DGRAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 5
bind(5, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("0.0.0.0")}, 16) = 0
sendto(5, "3\6\1\0\0\1\0\0\0\0\0\0\6-Ieth0\0\0\1\0\1", 24, MSG_NOSIGNAL, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("10.0.2.3")}, 16) = 24
sendto(5, "\252)\1\0\0\1\0\0\0\0\0\0\6-Ieth0\0\0\34\0\1", 24, MSG_NOSIGNAL, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("10.0.2.3")}, 16) = 24
poll([{fd=5, events=POLLIN}], 1, 2500)  = 0 (Timeout)
sendto(5, "3\6\1\0\0\1\0\0\0\0\0\0\6-Ieth0\0\0\1\0\1", 24, MSG_NOSIGNAL, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("10.0.2.3")}, 16) = 24
sendto(5, "\252)\1\0\0\1\0\0\0\0\0\0\6-Ieth0\0\0\34\0\1", 24, MSG_NOSIGNAL, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("10.0.2.3")}, 16) = 24
poll([{fd=5, events=POLLIN}], 1, 2500)  = 0 (Timeout)
close(5)                                = 0
writev(2, [{iov_base="ping: ", iov_len=6}, {iov_base=NULL, iov_len=0}], 2ping: ) = 6
writev(2, [{iov_base="-Ieth0: Try again", iov_len=17}, {iov_base=NULL, iov_len=0}], 2-Ieth0: Try again) = 17
writev(2, [{iov_base="", iov_len=0}, {iov_base="\n", iov_len=1}], 2
) = 1
close(1)                                = 0
close(2)                                = 0
exit_group(2)                           = ?
+++ exited with 2 +++
```

gobenji:
Regarding 1, IIUC you mean that ping to link-local address using ICMP sockets and `-I` on ppc64 was not working for you previously because you were using a kernel without
```
35a79e64de29 ping: fix the dif and sdif check in ping_lookup (v5.17-rc5)
```
or one of its backports, is that right?

I spun up the same bullseye ppc64 image that I used before and tried with an [earlier kernel package](https://snapshot.debian.org/archive/debian/20220310T221404Z/pool/main/l/linux/linux-image-5.10.0-11-powerpc64le_5.10.92-2_ppc64el.deb) that should not contain that fix but the result was still good for me. I would've expected the opposite, no? So far I haven't been able to reproduce the failure you saw.

I would be cautious about reverting cc44f4c because of the potential for regression but it's up to you and the commit author, @kerolasa.

Regarding 2, IIUC you're saying that it looks like it doesn't work because of an error in getaddrinfo() in musl? Did you try `ltrace` to check what is the failing call and its arguments?

Regarding the last point, I don't see a strong need to encourage one way or another but I don't have a firm stance about it. My opinion about replacing one option by another is still as posted in https://github.com/iputils/iputils/pull/399#issuecomment-1068641619 and following comments, I'm not a fan.

pevik:
> Regarding 1, IIUC you mean that ping to link-local address using ICMP sockets and `-I` on ppc64 was not working for you previously because you were using a kernel without
> 
> ```
> 35a79e64de29 ping: fix the dif and sdif check in ping_lookup (v5.17-rc5)
> ```
> 
> or one of its backports, is that right?

Yes
 
> I spun up the same bullseye ppc64 image that I used before and tried with an [earlier kernel package](https://snapshot.debian.org/archive/debian/20220310T221404Z/pool/main/l/linux/linux-image-5.10.0-11-powerpc64le_5.10.92-2_ppc64el.deb) that should not contain that fix but the result was still good for me. I would've expected the opposite, no? So far I haven't been able to reproduce the failure you saw.

Are you sure you enabled ICMP datagram socket with `sysctl -w net.ipv4.ping_group_range='0 2147483647'`. Because with it `5.10.0-11-powerpc64le (5.10.92-2)` and older fail with `-I`, while `5.10.0-13-powerpc64le (5.10.106-1)` works (fix was backported in LTS v5.10.102). I also cloned https://salsa.debian.org/kernel-team/linux and checked there are no ping related patches.

> I would be cautious about reverting [cc44f4c](https://github.com/iputils/iputils/commit/cc44f4c5c2e72f7aa833658a6978d924ff9e059d) because of the potential for regression but it's up to you and the commit author, @kerolasa.

Agree. Given also this is a corner case issue and kernel fixes are for anybody who keeps system updated.

> Regarding 2, IIUC you're saying that it looks like it doesn't work because of an error in getaddrinfo() in musl? Did you try `ltrace` to check what is the failing call and its arguments?

Unfortunately ltrace is not available for ppc64le (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=744752):
https://git.busybox.net/buildroot/tree/package/ltrace/Config.in
https://packages.debian.org/sid/ltrace
https://tracker.debian.org/media/packages/l/ltrace/control-0.7.3-6
openSUSE has it but it prints nothing due no arch support. (Obviously I'd need it in buildroot or in alpine to test musl.)

> Regarding the last point, I don't see a strong need to encourage one way or another but I don't have a firm stance about it. My opinion about replacing one option by another is still as posted in [#399 (comment)](https://github.com/iputils/iputils/pull/399#issuecomment-1068641619) and following comments, I'm not a fan.

I need to think about it twice But agree with you that not to fix my original implementation https://github.com/iputils/iputils/pull/399. And openly mention in man page also required kernel fix for -I and that -I does not work on older kernels without this fix.

gobenji:
> Are you sure you enabled ICMP datagram socket with `sysctl -w net.ipv4.ping_group_range='0 2147483647'`.

Yes.

I figured out the reason for the difference in our results. I was testing ping to a **link-local** address. The problem is with ping to a **localhost/loopback** address.

As you correctly identified, it was fixed in the kernel by
```
35a79e64de29 ping: fix the dif and sdif check in ping_lookup (v5.17-rc5)
```

It was not ppc64 specific.

```
root@vbullseye:~# uname -a
Linux vbullseye 5.10.0-11-amd64 #1 SMP Debian 5.10.92-2 (2022-02-28) x86_64 GNU/Linux
root@vbullseye:~# ip -6 addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 fd0e:224b:829:4050:5054:ff:fefd:900/64 scope global dynamic mngtmpaddr
       valid_lft 86392sec preferred_lft 14392sec
    inet6 fe80::5054:ff:fefd:900/64 scope link
       valid_lft forever preferred_lft forever

root@vbullseye:~# ping -I eth0 fe80::5054:ff:fefd:900 -c1
ping: Warning: source address might be selected on device other than: eth0
PING fe80::5054:ff:fefd:900(fe80::5054:ff:fefd:900) from :: eth0: 56 data bytes
64 bytes from fe80::5054:ff:fefd:900%eth0: icmp_seq=1 ttl=64 time=0.020 ms

--- fe80::5054:ff:fefd:900 ping statistics ---
rtt min/avg/max/mdev = 0.020/0.020/0.020/0.000 ms

root@vbullseye:~# sysctl net.ipv4.ping_group_range="0 10000"
net.ipv4.ping_group_range = 0 10000
root@vbullseye:~# ping -I eth0 fe80::5054:ff:fefd:900 -c1
ping: Warning: source address might be selected on device other than: eth0
PING fe80::5054:ff:fefd:900(fe80::5054:ff:fefd:900) from :: eth0: 56 data bytes

--- fe80::5054:ff:fefd:900 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms
root@vbullseye:~# ping -I eth0 fd0e:224b:829:4050:5054:ff:fefd:900 -c1
PING fd0e:224b:829:4050:5054:ff:fefd:900(fd0e:224b:829:4050:5054:ff:fefd:900) from fd0e:224b:829:4050:5054:ff:fefd:900 eth0: 56 data bytes

--- fd0e:224b:829:4050:5054:ff:fefd:900 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms
```

pevik:
> I figured out the reason for the difference in our results. I was testing ping to a **link-local** address. The problem is with ping to a **localhost/loopback** address.

I understand that `inet6 fe80::5054:ff:fefd:900/64 scope link` **is** link-local address. It's on `eth0`, not on `lo`. Do I miss something obvious in the terminology?

> As you correctly identified, it was fixed in the kernel by
> 
> ```
> 35a79e64de29 ping: fix the dif and sdif check in ping_lookup (v5.17-rc5)
> ```
> 
> It was not ppc64 specific.

Correct, it's not any arch specific. I was wrong. I got confused, because until I test `35a79e64de29` and commit before, I was able to reproduce it only on ppc64le.

gobenji:
I think you confused a link-local address with an address configured on a local interface (what I called localhost/loopback address).

A link-local address is not necessarily an address that's configured on one of the local host interfaces. An ipv6 link-local address is defined in [RFC4291 §2.5.6](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.6). It's an address within the fe80::/64 subnet. It is reachable without going through a router.

Secondly, even if a locally-configured address is assigned to an interface like "eth0", the route to it goes over the loopback interface:
```
root@vbullseye:~# ip rou get fe80::5054:ff:fefd:900
local fe80::5054:ff:fefd:900 from :: dev lo table local proto kernel src fe80::5054:ff:fefd:900 metric 0 pref medium
```
(notice `dev lo`)

That's why I called it a loopback address.

In my previous comment you can see the same ping problem occurs with fd0e:224b:829:4050:5054:ff:fefd:900 which is a localhost address but is not a link-local address.

pevik:
I have problems with github notifications, that's why I always miss your reply.

> I think you confused a link-local address with an address configured on a local interface (what I called localhost/loopback address).
> 
> A link-local address is not necessarily an address that's configured on one of the local host interfaces. An ipv6 link-local address is defined in [RFC4291 §2.5.6](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.6). It's an address within the fe80::/64 subnet. It is reachable without going through a router.

OK, we finally understand each other :). No, I was talking about link-local address where I reproduced the problem (on first system I met the problem it was configured also on `eth0`, on others it was on `lo`). I briefly read this RFC before to understand this problem, I just didn't notice it goes always over `lo`. That explains why I sometimes see `ping: Warning: source address might be selected on device other than: eth0` when ping link-local with `-I eth0`

> Secondly, even if a locally-configured address is assigned to an interface like "eth0", the route to it goes over the loopback interface:
> 
> ```
> root@vbullseye:~# ip rou get fe80::5054:ff:fefd:900
> local fe80::5054:ff:fefd:900 from :: dev lo table local proto kernel src fe80::5054:ff:fefd:900 metric 0 pref medium
> ```
> 
> (notice `dev lo`)
> 
> That's why I called it a loopback address.

But it's a link-local address.
 
> In my previous comment you can see the same ping problem occurs with fd0e:224b:829:4050:5054:ff:fefd:900 which is a localhost address but is not a link-local address.

IMHO this is a different issue, I noticed before that binding with `-I` on a global scope address is broken on v5.18 kernel. v5.17 and fixed LTS kernels (all but 4.9 LTS) which contains fix - these 3 commits, work:
2afc3b5a31f9 ("ping: fix the sk_bound_dev_if match in ping_lookup")
35a79e64de29 ("ping: fix the dif and sdif check in ping_lookup")
cd33bdcbead8 ("ping: remove pr_err from ping_lookup")

Either v5.18 introduced a regression or we have another bug in ping, I'll try look into it.
This also confused me and took some time (different IPv6 address types have different bugs).

FYI 4.9 LTS would need extra "feature" commits to be able to backport fixes, IMHO unlikely to get backported 
(I asked https://lore.kernel.org/netdev/YppqNtTmqjeR5cZV@pevik/).

pevik:
> In my previous comment you can see the same ping problem occurs with fd0e:224b:829:4050:5054:ff:fefd:900 which is a localhost address but is not a link-local address.

FYI It looks like [e6652a8ef3e6 ("net: ping6: Fix ping -6 with interface name")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=e6652a8ef3e64d953168a95878fe29b934ad78ac) from v5.19-rc1 fixes the problem introduced in [13651224c00b ("net: ping6: support setting basic SOL_IPV6 options via cmsg")](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=13651224c00b74a02c731b7840577f14e0311c83) in v5.18-rc1.

ping/ping.c
pevik: After long research, given that `-Iifname` is not going to be fixed for 4.9 LTS branch (https://lore.kernel.org/netdev/YppqNtTmqjeR5cZV@pevik/) or people may use unfixed kernels I'd keep suggesting only `address%<ifname|scope-id>`. But I'll put some info to the man page.

doc/ping.xml
```diff
@@ -107,17 +107,34 @@ xml:id="man.ping">
...
-    <para>
     <command>ping</command> can also send IPv6 Node Information
     Queries (RFC4620). Intermediate
     <emphasis remap="I">hop</emphasis>s may not be allowed, because
     IPv6 source routing was deprecated (RFC5095).</para>
   </refsection>
 
+  <refsection xml:id="destination">
+    <info>
+      <title>DESTINATION</title>
```
gobenji, line 118: TBH I feel the content is a bit awkward but I'm affraid I'm just bike shedding. You've been working on this for a long time so I'll just let the maintainers chime in.

ping/ping.c
```diff
@@ -557,6 +557,15 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
+	unsigned char buf[sizeof(struct in6_addr)];
+	if (!strchr(target, '%') && sock6.socktype == SOCK_DGRAM &&
+		inet_pton(AF_INET6, target, buf) > 0 &&
+		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
+			error(0, 0, _(
+				"Warning: IPv6 link-local address on ICMP datagram socket requires ifname or scope-id"
```
gobenji, line 565: Given the man page changes, "requires" -> "may require" ?

doc/ping.xml
gobenji: On a positive note, I agree that a separate section is worthwhile to discuss interface specification with ipv6 link local destinations.

pevik: I agree that the content of man page section I added in commit "man/ping: Move destination to own section, improve" is a bit awkward. I wanted to be clear about unfixed kernels and old iputils, but I don't like wording myself. Maybe just mentioned that it was fixed in 5.17 and in LTS kernels >= 4.14.

gobenji: Ok, here were my thoughts:

> Newer kernels (>= 5.17 and LTS: 5.15, 5.10, 5.4, 4.19, 4.14) contain

There will come a time in the future when those are no longer "newer kernels". It would also be more useful to mention the sublevel of the stable kernels, ie.: 5.15.19, 5.10.96, 5.4.176, 4.19.228, 4.14.265

> It also requires ping >= s20180629, therefore using

Generally, the man page is only relevant for the version of ping it is shipped with, right? I would skip that part.

How about:
```
IPV6 LINK-LOCAL DESTINATIONS

For IPv6, when the destination address has link-local scope and ping is using
ICMP datagram sockets, the output interface must be specified. When ping is
using raw sockets, it is not strictly necessary to specify the output
interface but it should be done to avoid ambiguity when there are multiple
possible output interfaces.

There are two ways to specify the output interface:
* using the % notation
	The destination address is postfixed with % and the output interface
	name or ifindex, for example:
		ping fe80::5054:ff:fe70:67bc%eth0
		ping fe80::5054:ff:fe70:67bc%2
* using the -I option
	When using ICMP datagram sockets, this method is only supported on
	some kernel versions (...).
```
The remaining issue IMO is that it is not easy for a user to know if ping is using ICMP sockets or raw sockets.

doc/arping.xml
```diff
@@ -60,6 +60,15 @@ xml:id="man.arping">
     <emphasis remap="I">source</emphasis>.</para>
   </refsection>
 
+  <refsection xml:id="destination">
```
gobenji, line 63: IMO it is not necessary to add a separate "destination" section just to be consistent with the ping manpage.

I would rather append the following to the Description section:
```
arping supports IPv4 addresses only. For IPv6, see ndisc6(8).
```
And add ndisc6 to the "See Also" section.

doc/tracepath.xml
```diff
@@ -75,6 +75,17 @@ xml:id="man.tracepath">
     history.</para>
   </refsection>
 
+  <refsection xml:id="destination">
```
gobenji, line 78: I think I would drop this patch. The first two options documented in the manpage are -4 and -6 so it is already quite clear that tracepath supports ipv4 and ipv6.

doc/ping.xml
pevik: Thanks, I'll use this, add you as author. I'd keep info in git commit message (also helps me to remember on changelog).

One more note to changelog: musl just does not support `-I iface` even with fixed kernel (I suppose `getaddrinfo()` is different). But OTOH it allows to ping even without `'%'-notation` (but works with the notation).

pevik: > The remaining issue IMO is that it is not easy for a user to know if ping is using ICMP sockets or raw sockets.

I've been thinking about print this when `-v` or even always.
https://github.com/iputils/iputils/issues/411

pevik: I added note about unsupported musl and test also uclibc-ng.

pevik: OK, it works also on uclibc-ng. Thus only musl not supported (ignoring bionic although aosp thinks about updating iputils.

doc/ping.xml
```diff
@@ -118,6 +118,46 @@ xml:id="man.ping">
...
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+          • using the
+          <emphasis remap="I">-I option</emphasis>
+        </term>
+        <listitem>
+          <para>When using <emphasis remap="I">ICMP datagram sockets</emphasis>,
+          this method is only supported on some kernel versions: 5.17,
```
gobenji, line 153: I think this should have been ">=5.17" or something like that, right?

pevik: Ah, yep. Is only one `>=` enough? This looks a bit strange to me, but better than without `>=`:

`this method is only supported on some kernel versions: >= 5.17, 5.15.19, 5.10.96, 5.4.176, 4.19.228, 4.14.265.`

gobenji: I like "since".

doc/ping.xml
```diff
@@ -118,6 +118,46 @@ xml:id="man.ping">
...
   </refsection>
 
+  <refsection>
+    <info>
+      <title>IPV6 LINK-LOCAL DESTINATIONS</title>
+    </info>
+    <para>For IPv6, when the destination address has link-local scope and
+    ping is using <emphasis remap="I">ICMP datagram sockets</emphasis>,
+    the output interface must be specified.
+    When <command>ping</command> is using raw sockets, it is not strictly
```
gobenji, line 128: For consitency, should "raw" also have emphasis?

doc/ping.xml
```diff
@@ -118,6 +118,46 @@ xml:id="man.ping">
     IPv6 source routing was deprecated (RFC5095).</para>
   </refsection>
 
+  <refsection>
```
gobenji, line 121: BTW, I would've placed the new section after OPTIONS; among the already present miscellaneous info sections like ICMP PACKET DETAILS and so on.

doc/ping.xml
pevik: yep

doc/ping.xml
```diff
@@ -118,6 +118,46 @@ xml:id="man.ping">
...
+    <info>
+      <title>IPV6 LINK-LOCAL DESTINATIONS</title>
+    </info>
+    <para>For IPv6, when the destination address has link-local scope and
+    ping is using <emphasis remap="I">ICMP datagram sockets</emphasis>,
```
gobenji, line 126: ping -> <command>ping</command> for consistency.

doc/ping.xml
gobenji: `this method is only supported since the following kernel versions:`

pevik:
Merged as dbb76fe144d0ac67af9b7fb2bbbe7634466e692e, ed7c27d32ec6498569b6715008643d03517a804f, c4b8b44a0403f52efc1c76adca9ee2181688ea9c, 6becbb3e1a34f957303ec7340b354c120ee59db0. Additional fixes in 60c064071896468ff4c54583acbffaba8374ba61 (https://github.com/iputils/iputils/pull/413).

PR #409, opened 2022-05-31
Fix the problem of potential memory leakage.

ggvl:
If user use '-p' opt multi-times, the previous pointer generated by
strdup functiong will be discarded.

Signed-off-by: lvgenggeng <lvgenggeng@uniontech.com>

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -449,6 +449,7 @@ main(int argc, char **argv)
 			break;
 		case 'p':
 			rts.opt_pingfilled = 1;
+			free(outpack_fill);
 			outpack_fill = strdup(optarg);
 			if (!outpack_fill)
 				error(2, errno, _("memory allocation failed"));


ggvl:
@hwoarang need a maintainer to approve running workflows

ggvl:
@zx2c4 @rossburton @sfionov review please, thanks.

ggvl:
@zx2c4 @rossburton @hwoarang @sfionov need review, thanks.

ggvl:
@zx2c4 @rossburton @hwoarang @sfionov need review, thanks.

rossburton:
I'm not a maintainer, so I can't approve workflows.

However, that patch changes behaviour: if I pass `-p` twice then the second value is ignored. Conventionally, the second value will replace the first.

ggvl:
> I'm not a maintainer, so I can't approve workflows.
> 
> However, that patch changes behaviour: if I pass `-p` twice then the second value is ignored. Conventionally, the second value will replace the first.

Yes. you are right.

ggvl:
@zx2c4 @hwoarang @sfionov

pevik:
@ggvl FYI not approved workflow is not blocking this, lack of my time block it. iputils have practically no functional tests, thus I need to look carefully into the code (there have been many regressions in the past in ping).
BTW I wonder why you don't have allowed GitHub Actions in your fork: https://github.com/ggvl/iputils/actions

pevik:
> However, that patch changes behaviour: if I pass `-p` twice then the second value is ignored. Conventionally, the second value will replace the first.

@[rossburton](https://github.com/rossburton) IMHO last value is always used (in master, and after this change). Or do I miss something?

rossburton:
An earlier revision ignored all but the first option, this is fixed now.

pevik:

ping/ping.c
```diff
@@ -449,9 +449,15 @@ main(int argc, char **argv)
 			break;
 		case 'p':
 			rts.opt_pingfilled = 1;
+			if (outpack_fill) {
+				free(outpack_fill);
+				outpack_fill = NULL;
```
metan-ucw, line 454: I do not think that we have to clear the `outpack_fill` here, we assign it new value right after this. Also `free(NULL)` is no-op so this whole part can be just:

```
        ...
        rts.opt_pingfilled = 1;
        free(outpack_fill);
        outpack_fill = strdup(optarg);
        if (!outpack_fill)
                ...
```

ggvl:
@pevik

pevik:
@ggvl Thinking it twice Could we merge just fixing the leak first:
```
--- ping/ping.c
+++ ping/ping.c
@@ -449,6 +449,7 @@ main(int argc, char **argv)
                        break;
                case 'p':
                        rts.opt_pingfilled = 1;
+                       free(outpack_fill);
                        outpack_fill = strdup(optarg);
                        if (!outpack_fill)
                                error(2, errno, _("memory allocation failed"));
```
1) check for max 16 bytes is a separate issue, which should be in a separate commit (you haven't even mention it in the commit message, but separate commit would be better).
2) I'm not sure if `strlen(outpack_fill) > 16` is a correct check. Because pattern is in hex, busybox implementation simply checks if number is in 0..255 range.

ggvl:
> 1. check for max 16 bytes is a separate issue, which should be in a separate commit (you haven't even mention it in the commit message, but separate commit would be better).

Fine. I will separate it.
> 2. I'm not sure if strlen(outpack_fill) > 16 is a correct check. Because pattern is in hex, busybox implementation simply checks if number is in 0..255 range.

I read fill function and find it use isxdigit to check each byte in the readbuf, so the maxlen should be 32.

ggvl:
@pevik This on is ok and I will commit the second patch later.

pevik:
> I read the fill() function and find it use isxdigit to check each byte in the readbuf, so the maxlen should be 32.

Agree, but it'd be worth to check why busybox implementation allows only 0..255 decimal value.
https://github.com/iputils/iputils/issues/412

PR #416, opened 2022-07-15
clockdiff: update capabilities documentation

michaelkarr:
Due to the following line, running `clockdiff` requires `CAP_SYS_NICE` on order to be able to execute fully. https://github.com/iputils/iputils/blob/master/clockdiff.c#L539


michaelkarr:
Is there a reason why this has been closed? The capability was required on my system and I’d like to update the documentation so other folks can hopefully save a bit of time as well.

michaelkarr:
Ah apologies, did not see the patch was merged elsewhere. Thank you for the help!

Issue #406, opened 2022-03-30
No error message on socket creation failure.


gobenji:
If `ping` has neither CAP_NET_RAW nor ICMP socket creation ability (most likely because it is incorrectly installed), it used to print an error message:
```
user@vbullseye:~$ ping 127.0.0.1
ping: socket: Operation not permitted
```
But it doesn't anymore:
```
user@vsid:~$ ping 127.0.0.1
user@vsid:~$ echo $?
2
user@vsid:~$ ping -v 127.0.0.1
ping: socket: Operation not permitted
ping: socket: Operation not permitted
```
This regression was introduced by commit 79d713eab618 ("ping: Remove 'unsupported IPv6' warning on disabled IPv6")

PR #418, opened 2022-08-10
ping: Fix socket error reporting

pevik:
There is actually no need for errno EAFNOSUPPORT special handling,
nor for any other errno. *All* errors needs to be reported on verbose
mode or when socket is required. If AF_INET6 socket is used with
disabled IPv6, error reporting is done by gai_strerror() (ipv6.disable=1)
or by connect() error handling in ping6_run() (EADDRNOTAVAIL on
net.ipv6.conf.all.disable_ipv6=1).

Bug was hidden because condition "errno == EAFNOSUPPORT && socktype ==
AF_INET6" introduced in d141cb6 was always false as AF_INET6 is is an address
family not a socket type, until otherwise correct fix 904cdb6. Attempt
to fix it in 79d713e introduced regression that other errors weren't
reported (e.g. EPERM on RAW socket on non-root, see #406).

Fixes: https://github.com/iputils/iputils/issues/406
Fixes: 79d713e ("ping: Remove 'unsupported IPv6' warning on disabled IPv6")
Fixes: d141cb6 ("ping: work with older kernels that don't support ping sockets")

```
Reported-by: Benjamin Poirier <benjamin.poirier@gmail.com>
```

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -149,11 +149,7 @@ static void create_socket(struct ping_rts *rts, socket_st *sock, int family,
 	}
 
 	if (sock->fd == -1) {
-		/* Report error related to disabled IPv6 only when IPv6 also failed or in
-		 * verbose mode. Report other errors always.
-		 */
-		if ((errno == EAFNOSUPPORT && family == AF_INET6 && requisite) ||
-		    rts->opt_verbose)
+		if (requisite || rts->opt_verbose)
 			error(0, errno, "socket");
 		if (requisite)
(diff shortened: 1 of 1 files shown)


pevik:
@gobenji @lekto Could you please have a look?

lekto:
I don't see this message any more on s20211215, so looks like it was fixed for me at some point.

gobenji:
Tested-by: Benjamin Poirier <benjamin.poirier@gmail.com>

PR #419, opened 2022-08-11
Simple maintenance improvements

a1346054:


build.sh
```diff
@@ -49,7 +49,7 @@ run()
 	ret=$?
 
 	if [ $ret -ne 0 ]; then
-		echo "ERROR: '$@' failed, exit code: $ret" >&2
+		echo "ERROR: '$*' failed, exit code: $ret" >&2
```
pevik, line 52: '$@' works well on dash. Is there problem on busybox sh or any other shell?

build.sh
```diff
@@ -115,35 +115,35 @@ case "$1" in
 	*) echo "ERROR: wrong command '$1'" >&2; exit 1;;
 esac
 
-if [ -z "$cmd" -o "$cmd" = "dependencies" ]; then
+if [ -z "$cmd"  ] || [ "$cmd" = "dependencies" ]; then
```
pevik, line 118: `-o` is working on dash and busybox sh. Is there a real problem on some shell? Because shellcheck general complains aren't a real reason to change something what works (yes, shell check complains about `Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.` but in this case it works).

ci/debian.sh
```diff
@@ -36,6 +36,6 @@ apt install -y --no-install-recommends \
 	pkg-config \
 	xsltproc
 
-if [ "$BACKPORT_REPO" ]; then
+if [ -n "$BACKPORT_REPO" ]; then
```
pevik, line 39: Again, this really works on all common shell being used on Linux distros. Being pedant like this does make sense on autotools configure script which is used on really different shells and OS, but here it's really not required.

pevik:
Thank you for your effort. Although I hesitate to take shell fixes unless are really fixing something, there are certainly useful commits in this PR. Could you please sign your commits, i.e. `Signed-off-by: Your Name <your@email>` (see examples in git history), or at least post it here, so that I can cherry-pick.

build.sh
a1346054: It's mixing arrays and strings, here's a good resource https://www.shellcheck.net/wiki/SC2145

build.sh
a1346054: Strictly speaking, it's undefined behavior because it hits the case of more than 4 arguments to [

From https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html "4 arguments: The results are unspecified. "

ci/debian.sh
a1346054: You are correct that it is not required, but most readers will be surprised by a behavior such as
```bash
var=0
if [ "$var" ]; then
    echo "This will still print"
fi
```

because that's where shell syntax deviates from expectations of many programming languages. Adhering to principle of least astonishment is a good practice.

pevik: I'm sorry, but this project has more important issues than discussing shell style.

a1346054:
I can drop the shell fixes from the PR, or you can only cherry-pick the commits you wish to merge.

I am currently away from the machine that has the gpg key to re-sign the individual commits, but you can use `Signed-off-by: a1346054 <36859588+a1346054@users.noreply.github.com>`

All these contributions would be considered trivial under any contributor license agreement.

build.sh
pevik: POSIX compliant shell syntax, which I use for a reason (i.e. working on dash, which is Debian default or busybox sh which is on many embedded systems) does not have arrays. This is artificial problem. While shellcheck is great to found hidden bugs in complex shell scripts, here is misleading.

ci/debian.sh
a1346054: Understandable.

ci/fedora.sh
```diff
@@ -15,7 +15,7 @@ yum -y install \
 yum -y install libidn2-devel docbook5-style-xsl || true
 
 # supported since CentOS 7 (CentOS 6 don't have python 3.5 meson dependency)
-if [ "$(basename $0)" = "centos.sh" ]; then
+if [ "$(basename "$0")" = "centos.sh" ]; then
```
pevik, line 18: Also this: protect space in PATH: again problem found by shellcheck, not a real problem. This script is used for CI. It can be used locally, but if anybody creates space in path, many things get broken. Again: this project has real issues, why wasting time with fixing artificial problems?

tools/update-project-keyring.sh
```diff
@@ -1,12 +1,12 @@
...
 
-keys="4807F894E0620EEA" # YOSHIFUJI Hideaki
-keys="9FC5E7B74A45D0DE 4A8D0987698B0274 $keys" # David Heidelberg
+keys="902265EB1ECF3AD10BDF80D84807F894E0620EEA" # YOSHIFUJI Hideaki
+keys="73540C3F796DC1C30ECACCAC9FC5E7B74A45D0DE 5192BA645ED64BC326D3E0114A8D0987698B0274 $keys" # David Heidelberg
 keys="25999B407FB6A9B701BBF1E40D46FEF7E61DBB46 8ED396E37E38D471A00530D3A9553245FDE9B739 $keys" # Sami Kerola
 keys="33C58482C402292D2E3C5C069709F90C3C96FFC8 $keys" # Thomas Deutschmann
 keys="2016FEA4858B1C36B32E833AC0DEC2EE72F33A5F $keys" # Petr Vorel
-keys="4AEE18F83AFDEB23 $keys" # GitHub (web-flow commit signing)
+keys="5DE3E0509C47EA3CF04A42D34AEE18F83AFDEB23 $keys" # GitHub (web-flow commit signing)
```
pevik, line 9: NOTE for myself: https://security.stackexchange.com/questions/84280/short-openpgp-key-ids-are-insecure-how-to-configure-gnupg-to-use-long-key-ids-i

pevik:
Of course, I have checked git metadata before thus I had noticed `a1346054 <36859588+a1346054@users.noreply.github.com>` , but hiding behind github is not welcome behavior. Is it really needed? This is not git repository of tor (anonymity software).

a1346054:
Added `Signed-off-by:` to all commits and dropped commits that fix "stylistic" issues in scripts.

pevik:
> Added `Signed-off-by:` to all commits and dropped commits that fix "stylistic" issues in scripts.

Thanks for taking care, but I'd much more prefer to know your real name and real email contact, it'd be enough to paste it here.

pevik:
@a1346054 ping: any chance you add your real name and email? Hiding behind github is not really appreciated.

pevik:
@a1346054 FYI we follow the Developer’s Certificate of Origin from Linux Kernel
https://www.kernel.org/doc/html/latest/process/submitting-patches.html?highlight=signed%20off#sign-your-work-the-developer-s-certificate-of-origin
> using your real name (sorry, no pseudonyms or anonymous contributions.)

a1346054:
I don't wish to go through all that legalese. Feel free to close this pull request and make the changes in your own name instead.

PR #426, opened 2022-10-13
ping: Add support for Identifier field for icmp ECHO_REQUEST

rumvadim:
Sometimes I need to manually specify identifier for series of ping requests, so I decided to add this feature to the ping.
I think that not only me will find it useful.

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -24,7 +24,7 @@ xml:id="man.ping">
     <cmdsynopsis sepchar=" ">
       <command>ping</command>
       <arg choice="opt" rep="norepeat">
-        <option>-aAbBdDfhLnOqrRUvV46</option>
+        <option>-aAbBdDefhLnOqrRUvV46</option>
       </arg>
       <arg choice="opt" rep="norepeat">
         <option>-c
@@ -202,6 +202,22 @@ xml:id="man.ping">
           <para>Set the SO_DEBUG option on the socket being used.
           Essentially, this socket option is not used by Linux
           kernel.</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
+          <option>-e</option>
+        </term>
+        <listitem>
+          <para>Set identification field of the
+          <emphasis remap="I">ECHO_REQUEST</emphasis>.
+          Setting identification to 0 implies
+          <emphasis remap="I">raw socket</emphasis>
+          option since setting zero identification on
+          <emphasis remap="I">ICMP datagram sockets</emphasis> isn't
+          supported right now. The value of the field may be printed with
+          <option>-v</option> option.
+          </para>
         </listitem>
       </varlistentry>
       <varlistentry>
@@ -962,10 +978,12 @@ xml:id="man.ping">
     <para>
     <command>ping</command> requires CAP_NET_RAW capability to be
     executed 1) if the program is used for non-echo queries (See
-    <option>-N</option> option), or 2) if kernel does not support
+    <option>-N</option> option) or set zero identification field of
+    the <emphasis remap="I">ECHO_REQUEST</emphasis> (See
+    <option>-e</option>), or 2) if kernel does not support
     ICMP datagram sockets, or 3) if the user is not allowed to
-    create an ICMP echo socket. The program may be used as set-uid
-    root.</para>
+    create an ICMP echo socket. The program may be used as set-uid root.
+    </para>
   </refsection>
 
   <refsection xml:id="availability">
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -275,6 +275,7 @@ main(int argc, char **argv)
 		.tmin = LONG_MAX,
 		.pipesize = -1,
 		.datalen = DEFDATALEN,
+		.ident = -1,
 		.screen_width = INT_MAX,
 #ifdef HAVE_LIBCAP
 		.cap_raw = CAP_NET_RAW,
@@ -311,7 +312,7 @@ main(int argc, char **argv)
 		hints.ai_family = AF_INET6;
 
 	/* Parse command line options */
-	while ((ch = getopt(argc, argv, "h?" "4bRT:" "6F:N:" "aABc:dDfi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:")) != EOF) {
+	while ((ch = getopt(argc, argv, "h?" "4be:RT:" "6F:N:" "aABc:dDfi:I:l:Lm:M:nOp:qQ:rs:S:t:UvVw:W:")) != EOF) {
 		switch(ch) {
 		/* IPv4 specific options */
 		case '4':
@@ -321,6 +322,13 @@ main(int argc, char **argv)
 			break;
 		case 'b':
 			rts.broadcast_pings = 1;
+			break;
+		case 'e':
+			rts.ident = htons(strtoul_or_err(optarg, _("bad value for identifier"), 0, IDENTIFIER_MAX));
+			if (rts.ident == 0) {
+				/* Right now linux doesn't support setting ident == 0 in SOCK_DGRAM mode */
+				hints.ai_socktype = SOCK_RAW;
+			}
 			break;
 		case 'R':
 			if (rts.opt_timestamp)
@@ -805,9 +813,14 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 			error(2, errno, "IP_MTU_DISCOVER");
 	}
 
-	if (rts->opt_strictsource &&
-	    bind(sock->fd, (struct sockaddr *)&rts->source, sizeof rts->source) == -1)
-		error(2, errno, "bind");
+	int set_ident = rts->ident > 0 && sock->socktype == SOCK_DGRAM;
+	if (set_ident)
+		rts->source.sin_port = rts->ident;
+
+	if (rts->opt_strictsource || set_ident) {
+		if (bind(sock->fd, (struct sockaddr *)&rts->source, sizeof rts->source) == -1)
+			error(2, errno, "bind");
+	}
 
 	if (sock->socktype == SOCK_RAW) {
 		struct icmp_filter filt;
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -214,8 +214,9 @@ static int parseflow(char *str)
 	if (str[0] == '0' && (str[1] == 'x' || str[1] == 'X')) {
 		cp = str + 2;
 		val = (int)strtoul(cp, &ep, 16);
-	} else
+	} else {
 		val = (int)strtoul(str, &ep, 10);
+	}
 
 	/* doesn't look like decimal or hex, eh? */
 	if (*ep != '\0')
@@ -237,8 +238,9 @@ static int parsetos(char *str)
 	if (str[0] == '0' && (str[1] == 'x' || str[1] == 'X')) {
 		cp = str + 2;
 		tos = (int)strtol(cp, &ep, 16);
-	} else
+	} else {
 		tos = (int)strtol(str, &ep, 10);
+	}
 
 	/* doesn't look like decimal or hex, eh? */
 	if (*ep != '\0')
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -82,6 +82,8 @@ void usage(void)
 		"  -b                 allow pinging broadcast\n"
 		"  -R                 record route\n"
 		"  -T <timestamp>     define timestamp, can be one of <tsonly|tsandaddr|tsprespec>\n"
+		"  -e <identifier>    define identifier for ping session, default is random for\n"
+		"                     SOCK_RAW and kernel defined for SOCK_DGRAM\n"
 		"\nIPv6 options:\n"
 		"  -6                 use IPv6\n"
 		"  -F <flowlabel>     define flow label, default is random\n"
@@ -530,7 +532,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 			*p++ = i;
 	}
 
-	if (sock->socktype == SOCK_RAW)
+	if (sock->socktype == SOCK_RAW && rts->ident == -1)
 		rts->ident = rand() & IDENTIFIER_MAX;
 
 	set_signal(SIGINT, sigexit);
@@ -793,6 +795,9 @@ restamp:
 
 		if (pr_reply)
 			pr_reply(icmph, cc);
+
+		if (rts->opt_verbose && rts->ident != -1)
+			printf(_(" ident=%d"), ntohs(rts->ident));
 
 		if (hops >= 0)
 			printf(_(" ttl=%d"), hops);
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -531,7 +531,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 	}
 
 	if (sock->socktype == SOCK_RAW)
-		rts->ident = rand() & 0xFFFF;
+		rts->ident = rand() & IDENTIFIER_MAX;
 
 	set_signal(SIGINT, sigexit);
 	set_signal(SIGALRM, sigexit);
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -66,6 +66,7 @@
 #define	MAXWAIT		10		/* max seconds to wait for response */
 #define MININTERVAL	10		/* Minimal interpacket gap */
 #define MINUSERINTERVAL	2		/* Minimal allowed interval for non-root */
+#define IDENTIFIER_MAX	0xFFFF		/* max unsigned 2-byte value */
 
 #define SCHINT(a)	(((a) <= MININTERVAL) ? MININTERVAL : (a))
 


pevik:
Thanks for your effort. Interesting idea, code generally LGTM (I have minor comments), but it'd be good if anybody else review whole idea (security perspective, the need to use raw socket, ...).

ping/ping_common.c
```diff
@@ -530,7 +531,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 			*p++ = i;
 	}
 
-	if (sock->socktype == SOCK_RAW)
+	if (sock->socktype == SOCK_RAW && rts->ident == -1)
 		rts->ident = rand() & 0xFFFF;
```
pevik, line 535: Maybe use the constant when you defined it?
`rts->ident = rand() & IDENTIFIER_MAX;`

ping/ping_common.c
```diff
@@ -794,6 +795,9 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
 		if (pr_reply)
 			pr_reply(icmph, cc);
 
+		if (rts->opt_verbose && rts->ident != -1)
+			printf(_(" ident=%d"), ntohs(rts->ident));
```
pevik, line 799: I thought it'd be interesting to print the ident always on `-v`. But that would change the output formatting, thus better to keep it as it as you suggested (only when ident specified by `-e`).

ping/ping.c
```diff
@@ -249,6 +250,30 @@ static int parsetos(char *str)
...
+	const char *cp;
+	unsigned long val;
+	char *ep;
+
+	/* handle both hex and decimal values */
+	if (str[0] == '0' && (str[1] == 'x' || str[1] == 'X')) {
+		cp = str + 2;
+		val = (int)strtoul(cp, &ep, 16);
+	} else
+		val = (int)strtoul(str, &ep, 10);
```
pevik, line 265: I guess handling also hexadecimal is the reason why you haven't used `strtoul_or_err()`. I don't like single option which has numeric parameter behaves differently than the rest (and you haven't mention it in man and help). I wonder if it's safe to add hex handling to `strtoul_or_err()`. But you don't have to bother with it, I can add it myself later.

Also nit: please add brackets also around `else`. `ping` code uses really ancient code style, I'd prefer for new code to use linux kernel code style https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces:

> This does not apply if only one branch of a conditional statement is a single statement; in the latter case use braces in both branches:

```
if (condition) {
        do_this();
        do_that();
} else {
        otherwise();
}
```

doc/ping.xml
```diff
@@ -204,6 +204,19 @@ xml:id="man.ping">
...
+      <varlistentry>
+        <term>
+          <option>-e</option>
+        </term>
+        <listitem>
+          <para>Set identification field of the ECHO_REQUEST.
+          This implies SOCK_RAW option since setting identification
+          on SOCK_DGRAM doesn't work right now. The value of
+          the field may be printed with
+          <option>-v</option> option.
```
pevik, line 216: This is the main concern I have, because on ICMP datagram socket this means requiring root. But there is no other way, it would not work on distros which switched to ICMP datagram socket.

Also, please use `<emphasis remap="I">raw socket</emphasis>` and `<emphasis remap="I">ICMP datagram sockets</emphasis>` - keep used terminology and formatting

doc/ping.xml
```diff
@@ -964,8 +977,8 @@ xml:id="man.ping">
     executed 1) if the program is used for non-echo queries (See
     <option>-N</option> option), or 2) if kernel does not support
     ICMP datagram sockets, or 3) if the user is not allowed to
-    create an ICMP echo socket. The program may be used as set-uid
-    root.</para>
+    create an ICMP echo socket, or 4) uses raw sockets for any reason.
+    The program may be used as set-uid root.</para>
```
pevik, line 981: Instead of `4) uses raw sockets for any reason.` you should change 1) to:

```
... 1) if the program is used for non-echo queries (See
<option>-N</option> option) or set identification field of the 
<emphasis remap="I">ECHO_REQUEST</emphasis> (See
 <option>-e</option>), ...
```

doc/ping.xml
```diff
@@ -24,7 +24,7 @@ xml:id="man.ping">
     <cmdsynopsis sepchar=" ">
       <command>ping</command>
       <arg choice="opt" rep="norepeat">
-        <option>-aAbBdDfhLnOqrRUvV46</option>
+        <option>-aAbBdDefhLnOqrRUvV46</option>
```
pevik, line 27: I also wondered, if `-e` is the best option to be used. Looking that [`fping`](https://fping.org/) uses `-e` to `show elapsed time on return packets`:
```
$ fping ::1 -e
::1 is alive (0.042 ms)
```
which is really `fping` specific, we can use `-e` for this purpose.

ping/ping_common.c
```diff
@@ -82,6 +82,7 @@ void usage(void)
 		"  -b                 allow pinging broadcast\n"
 		"  -R                 record route\n"
 		"  -T <timestamp>     define timestamp, can be one of <tsonly|tsandaddr|tsprespec>\n"
+		"  -e <identifier>    define identifier for ping session, default is random for SOCK_RAW and kernel defined for SOCK_DGRAM\n"
 		"\nIPv6 options:\n"
```
pevik, line 86: IMHO line is too long, maybe split it?

```
"  -e <identifier>    define identifier for ping session, default is random for\n"
"                     SOCK_RAW and kernel defined for SOCK_DGRAM\n"
```

doc/ping.xml
```diff
@@ -204,6 +204,19 @@ xml:id="man.ping">
...
         </listitem>
       </varlistentry>
+      <varlistentry>
+        <term>
+          <option>-e</option>
+        </term>
+        <listitem>
+          <para>Set identification field of the ECHO_REQUEST.
+          This implies SOCK_RAW option since setting identification
+          on SOCK_DGRAM doesn't work right now. The value of
```
pevik, line 214: BTW looking at kernel sources, do you know how is `iph->id` (the identifier) set for ICMP datagram socket (it's not touched in `net/ipv[46]/icmp.c`).

rumvadim:
> Thanks for your effort. Interesting idea, code generally LGTM (I have minor comments), but it'd be good if anybody else review whole idea (security perspective, the need to use raw socket, ...).
> 
> Also, could you share why this is useful to you?

Thanks for considering this change!

Basically how I found out about Identification field is the ping was either working without losses, or lost 100% of packages when run several times. Turned out one of BGP routes was faulty and dropped all packages. The BGP route was chosen by hashing some info that included Identification field in icmp echo, but not seq id. Since then I sometimes specify manually Identification field to fully reproduce ping run or to get to test different routes. I used hping3 for this, but it turned out their rtt calculation was very bad, so I decided to suggest this change to the ping utility.

ping/ping_common.c
rumvadim: done

ping/ping_common.c
rumvadim: My thoughts exactly

ping/ping.c
rumvadim: TBH I just copy-pasted parse_tos function :sweat_smile: 
I can use strtoul_or_err if you like, but I feel that hex input may be useful.

doc/ping.xml
rumvadim: Nowadays you don't need root, you just need CAP_NET_RAW on the binary. I'm not sure if it is set on ping utility by package mainteiners though. And I don't feel I have enough knowledge about security to suggest it to maintainers if it's not set.
I read man on SOCK_DGRAM and I didn't find a way to set Identification field, and it gets overriden by kernel. So I used SOCK_RAW.

rumvadim: formatting fixed

doc/ping.xml
rumvadim: done

doc/ping.xml
rumvadim: I'm down to whatever name you'd like. I've chosen `-e` because `-i` seemed too much for this option, `-d` was already used and the next letter in word `identification` is `e` :)

rumvadim:
So, I've read kernel source and found out one can set identification field through binding to port set to the value of the idenficatioin field one wants. However setting ident to 0 still requires RAW_SOCKET, so I changed the patch accordingly.

doc/ping.xml
pevik: I'll keep  `-e` unless anybody has a reason why not use it (if anybody reviews it).

ping/ping.c
```diff
@@ -803,9 +836,13 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
...
-	if (rts->opt_strictsource &&
-	    bind(sock->fd, (struct sockaddr *)&rts->source, sizeof rts->source) == -1)
-		error(2, errno, "bind");
+	int do_need_to_set_ident = rts->ident > 0 && sock->socktype == SOCK_DGRAM;
+	if (rts->opt_strictsource || do_need_to_set_ident) {
+		if (do_need_to_set_ident)
+			rts->source.sin_port = rts->ident;
+		if (bind(sock->fd, (struct sockaddr *)&rts->source, sizeof rts->source) == -1)
+			error(2, errno, "bind");
+  }
```
pevik, line 845: This part was not in original propose. I don't like the code style when variables are set before using, but it just follows the bad style from the past (including long `if/else` parts.
* `do_need_to_set_ident` is too long, how about just `set_ident`?
* final `}` has wrong indent.
IMHO this is a bit better:
```
int set_ident = rts->ident > 0 && sock->socktype == SOCK_DGRAM;
if (set_ident)
	rts->source.sin_port = rts->ident;

if (rts->opt_strictsource || set_ident) {
	if (bind(sock->fd, (struct sockaddr *)&rts->source, sizeof rts->source) == -1)
		error(2, errno, "bind");
}
```

pevik:
Thank you for updating.
+1 for splitting cleanup into separate commit. But I'd prefer if you do cleanup in first commit and then squash changes your changes (squash 1st and 3rd commit into single commit). You can overwrite with `git push -f` (some people don't know it).

ping/ping.c
```diff
@@ -273,6 +298,7 @@ main(int argc, char **argv)
 		.tmin = LONG_MAX,
 		.pipesize = -1,
 		.datalen = DEFDATALEN,
+		.ident = -1, /* ident is not set */
```
pevik, line 301: I'd remove the comment, it's obvious.

ping/ping.c
pevik: Right, could you please use `strtoul_or_err()` now? I'd be for adding hex to `strtoul_or_err()` after the release (trying to minimize changes before release).

pevik:
> So, I've read kernel source and found out one can set identification field through binding to port set to the value of the idenficatioin field one wants. However setting ident to 0 still requires RAW_SOCKET, so I changed the patch accordingly.

Could you share which part of kernel sources (file, function) handles it? Also it might be worth to share some info in git commit message.

rumvadim:
> Thank you for updating.
> +1 for splitting cleanup into separate commit. But I'd prefer if you do cleanup in first commit and then squash changes your changes (squash 1st and 3rd commit into single commit). You can overwrite with git push -f (some people don't know it).

I'm a little bit confused. 2nd commit fixes the comments you made in this review. I can't put it before the 1st commit, because there would nothing to fix. So I would put everything in one commit then. If you meant something else, please correct me.

ping/ping.c
rumvadim: done

rumvadim:
> I'm a little bit confused. 2nd commit fixes the comments you made in this review. I can't put it before the 1st commit, because there would be nothing to fix. So I would put everything in one commit then. If you meant something else, please correct me.

Redid it as you suggested.

rumvadim:
@pevik I think I corrected all your comments. Could you, please, check this look at the change again?

pevik:
@rumvadim I pushed your cleanup commits (separated them with two and add one with a bit more cleanup). That's the preparation.

I was going to merge slightly adjusted your commit: https://github.com/pevik/iputils/commits/identifier.fixes but this prevented me:

1) Not only `id` is not IPv4 specific, it's also in IPv6.
2) `-e` actually allows to modify `id` on IPv6, but only on raw socket (regardless the value).

Tested on kernel 6.0.1:

## Raw socket
```
$ sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 1	0
# tcpdump -i any -y LINUX_SLL2 icmp6 and dst ::1 -vvv
tcpdump: data link type LINUX_SLL2
tcpdump: listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes

$ ./builddir/ping/ping -c1 ::1
...
=> tcpdump output
13:20:12.911964 lo    In  IP6 (flowlabel 0xb39de, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo request, id 8170, seq 1
13:20:12.911981 lo    In  IP6 (flowlabel 0xe1a6f, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo reply, id 8170, seq 1
```

Modify `id` with `-e` on IPv6 raw socket (`id` is changed to `77`):

```
$ ./builddir/ping/ping -e 77 -c2 ::1
13:33:34.689258 lo    In  IP6 (flowlabel 0xb39de, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo request, id 77, seq 1
13:33:34.689266 lo    In  IP6 (flowlabel 0xe1a6f, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo reply, id 77, seq 1
13:33:35.712879 lo    In  IP6 (flowlabel 0xb39de, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo request, id 77, seq 2
13:33:35.712904 lo    In  IP6 (flowlabel 0xe1a6f, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo reply, id 77, seq 2
```

## ICMP datagram socket
Obviously this does not work on ICMP datagram socket (`id`is `1`):
```
# sysctl -w net.ipv4.ping_group_range='0 2147483647' # ICMP (SOCK_DGRAM) => ICMP datagram socket
$ ./builddir/ping/ping -e 77 -c2 ::1
13:30:03.598915 lo    In  IP6 (flowlabel 0xb39de, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo request, id 1, seq 1
13:30:03.598923 lo    In  IP6 (flowlabel 0xe1a6f, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo reply, id 1, seq 1
13:30:04.608776 lo    In  IP6 (flowlabel 0xb39de, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo request, id 1, seq 2
13:30:04.608797 lo    In  IP6 (flowlabel 0xe1a6f, hlim 64, next-header ICMPv6 (58) payload length: 64) localhost > localhost: [icmp6 sum ok] ICMP6, echo reply, id 1, seq 2
```


Did you intend to support only IPv4? IMHO it should be supported on both IPv6.
But how to decide whether to use raw socket by default? Without `-4` or `-6` or numeric address there is `AF_UNSPEC`, therefore following change would be too strict: `-e` would require raw socket unless `-4`:

```
+++ ping/ping.c
@@ -330,8 +330,13 @@ main(int argc, char **argv)
                        rts.ident = htons(strtoul_or_err(optarg, _("invalid argument"),
                                                         0, IDENTIFIER_MAX));
 
-                       /* Current Linux kernel doesn't support setting ident == 0 in SOCK_DGRAM mode */
-                       if (!rts.ident)
+                       printf("rts.ident: %d\n", ntohs(rts.ident));
+                       /*
+                        * Current Linux kernel doesn't support on SOCK_DGRAM setting
+                        * 1) ident == 0 on IPv4 2) setting any ident on IPv6.
+                        */
+                       if ((hints.ai_family == AF_INET && !rts.ident) ||
+                               (hints.ai_family != AF_INET && rts.ident >= 0))
                                hints.ai_socktype = SOCK_RAW;
                        break;
```
Therefore my fixes in https://github.com/pevik/iputils/tree/identifier.fixes.ipv6 would work better as it enables raw socket only for sock4 when `rts.ident == 0` (I prefer to write it in C style as `!rts.ident` and for sock6 on `sock4.fd == -1 && rts.ident >= 0` (sock4 not valid => IPv6 address):

```
+++ ping/ping.c
@@ -329,10 +329,6 @@ main(int argc, char **argv)
                case 'e':
                        rts.ident = htons(strtoul_or_err(optarg, _("invalid argument"),
                                                         0, IDENTIFIER_MAX));
-
-                       /* Current Linux kernel doesn't support setting ident == 0 in SOCK_DGRAM mode */
-                       if (!rts.ident)
-                               hints.ai_socktype = SOCK_RAW;
                        break;
                case 'R':
                        if (rts.opt_timestamp)
@@ -533,10 +529,29 @@ main(int argc, char **argv)
 
        /* Create sockets */
        enable_capability_raw();
-       if (hints.ai_family != AF_INET6)
+
+       if (hints.ai_family != AF_INET6) {
+
+               /*
+                * Current Linux kernel doesn't support on SOCK_DGRAM setting ident == 0
+                * on IPv4.
+                */
+               if (!rts.ident)
+                       hints.ai_socktype = SOCK_RAW;
+
                create_socket(&rts, &sock4, AF_INET, hints.ai_socktype, IPPROTO_ICMP,
                              hints.ai_family == AF_INET);
+       }
+
        if (hints.ai_family != AF_INET) {
+
+               /*
+                * Current Linux kernel doesn't support on SOCK_DGRAM any ident setting
+                * on IPv6.
+                */
+               if (sock4.fd == -1 && rts.ident >= 0)
+                       hints.ai_socktype = SOCK_RAW;
+
                create_socket(&rts, &sock6, AF_INET6, hints.ai_socktype, IPPROTO_ICMPV6, sock4.fd == -1);
                /* This may not be needed if both protocol versions always had the same value, but
                 * since I don't know that, it's better to be safe than sorry. */
@@ -544,6 +559,7 @@ main(int argc, char **argv)
                               rts.pmtudisc == IP_PMTUDISC_DONT ? IPV6_PMTUDISC_DONT :
                               rts.pmtudisc == IP_PMTUDISC_WANT ? IPV6_PMTUDISC_WANT : rts.pmtudisc;
        }
+
        disable_capability_raw();
 
        /* Limit address family on single-protocol systems */
```
## TODO
If you agree with these changes I need to update man page and help.
I also I need to study kernel sources and ICMP and ICMPv6 RFC a bit to get idea if/how `id` differs in both versions.
NOTE: iputils is not my primary work, I have limited time for it, but try to be quick.

rumvadim:
I actually did not think about IPv6 when I suggested this fix. I'll take a look at your changes and get back in a couple of days.

pevik:
@rumvadim FYI I updated https://github.com/pevik/iputils/tree/identifier.fixes.ipv6. I'd like to merge it soon, in 1 or 2 weeks I'd like to freeze git and ask people to test and translate.

pevik:
@rumvadim Fixed raw socket need detection, updated man and help and merged.
Thanks a lot for your contribution!

rumvadim:
@pevik Thank you for pushing my commit!
Sorry for absence - I currently very busy as work.
However I did find a way to use idintification field on IPv6 with SOCK_DGRAM (actually it's the same way as for IPv4). So I did another pull request: https://github.com/iputils/iputils/pull/433

pevik:
Two doc related fixes were needed, probably introduced by me: 19c9adbd5fe026e31cf0b3bbe4a5130c89c20693, 43e38f2215a16e8a2c245e3176bea2adb3ede958.

Issue #411, opened 2022-06-17
ping: print used socket type on verbose mode


pevik:
`ping` should print whether it uses RAW socket or ICMP datagram socket.
It could be on `-v` or even always (but that would change expected output).

pevik:
Fixed in https://github.com/iputils/iputils/commit/a6a905be594488b51a9b1fb3488a839ada8b6aa3 (https://github.com/iputils/iputils/pull/431)

PR #431, opened 2022-11-07
ping: Print socket type and address family on -v

pevik:
Implements: https://github.com/iputils/iputils/issues/411


ping/ping.c
```diff
@@ -82,6 +82,36 @@ ping_func_set_st ping4_func_set = {
...
+#define CASE_TYPE(x) case x: return #x;
+
+static char *str_family(int family)
+{
+	switch (family) {
+		CASE_TYPE(AF_UNSPEC)
+		CASE_TYPE(AF_INET)
+		CASE_TYPE(AF_INET6)
+	default:
+		error(2, 0, _("unknown protocol family: %d"), family);
```
richiejp, line 94: Ignorant question, but where does the memory come from for this?

Probably needs to be a static rotating buffer with at least 3 or 4 entries looking at the usage below.

ping/ping.c
```diff
@@ -599,6 +636,10 @@ main(int argc, char **argv)
 		error(2, 0, "%s: %s", target, gai_strerror(ret_val));
 
 	for (ai = result; ai; ai = ai->ai_next) {
+		if (rts.opt_verbose)
+			printf("ai->ai_family: %s, ai->ai_canonname: %s\n",
+				   str_family(ai->ai_family), ai->ai_canonname);
```
richiejp, line 641: Possibly ai_canonname can be NULL too. I'm not sure what POSIX specifies on this and whether all the libc's that iputils are likely to be compiled with check for NULL.

ping/ping.c
pevik: I'm missing something, are you talking about `error()` macro or `int family`?
`valgrind` is happy with this code.

ping/ping.c
```diff
@@ -82,6 +82,36 @@ ping_func_set_st ping4_func_set = {
...
+{
+	switch (family) {
+		CASE_TYPE(AF_UNSPEC)
+		CASE_TYPE(AF_INET)
+		CASE_TYPE(AF_INET6)
+	default:
+		error(2, 0, _("unknown protocol family: %d"), family);
+	}
+
+	return NULL;
```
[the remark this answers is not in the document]
pevik: Well, `NULL` is just for compiler, because `error(2, ...)` exits the program.

ping/ping.c
pevik: I haven't had a problem with `NULL`, but maybe I'm wrong. So would you change it to:
`, ai->ai_canonname ?: "");`

ping/ping.c
```diff
@@ -599,6 +636,11 @@ main(int argc, char **argv)
 		error(2, 0, "%s: %s", target, gai_strerror(ret_val));
 
 	for (ai = result; ai; ai = ai->ai_next) {
+		if (rts.opt_verbose)
+			printf("ai->ai_family: %s, ai->ai_canonname: %s\n",
+				   str_family(ai->ai_family),
+				   ai->ai_canonname ? ai->ai_canonname : _("not found"));
+
```
pevik, line 643: @richiejp NOTE: I put here `"not found" in case of `NULL`. The rest of strings should be handled well in `str_family()` and `str_socktype()`.

BTW I cannot use ` ai->ai_canonname ?: _("not found"));` due:

```
../ping/ping.c:642:54: warning: ISO C forbids omitting the middle term of a ‘?:’ expression [-Wpedantic]
  642 |                                    ai->ai_canonname ?: _("not found"));

```
A bit surprising, I suppose it's not by C99 but by other options. Flags configured with meson:
```
 default_options : [
        'c_std=c99',
        'warning_level=3',
        'localstatedir=var',
],
```

richiejp: yeah it's not in any C standard IIRC.

More importantly is "not found" a valid name? It would be nice to have some symbols in there which make it clear. Or print the name in quotes, then just use an empty string as default.

pevik: @richiejp I put just empty string (`""`), because it'd have to be some deeper error. Normally `getaddrinfo()` returns `-1`, thus executing gets quit by `error(2, 0, "%s: %s", target, gai_strerror(ret_val));`:

https://github.com/iputils/iputils/blob/43e38f2215a16e8a2c245e3176bea2adb3ede958/ping/ping.c#L597-L599

ping/ping.c
```diff
@@ -82,6 +82,36 @@ ping_func_set_st ping4_func_set = {
...
+		return "0";
+
+	switch (socktype) {
+		CASE_TYPE(SOCK_DGRAM)
+		CASE_TYPE(SOCK_RAW)
+	default:
+		error(2, 0, _("unknown sock type: %d"), socktype);
+	}
+
+	return NULL;
```
richiejp, line 112: We still return NULL here. How about just ""?

pevik: Thanks, makes sense, will update that before merge.

PR #432, opened 2022-11-08
ping: Print config options on -V

pevik:
Add to all programs (arping, clockdiff, ping, tracepath). Example:

```
$ ./ping -V
ping from iputils 20211215-58-gd746509
libcap: yes, IDN: yes, NLS: no, error.h: yes, getrandom(): yes, __fpending(): yes
```

Some info does not make sense for certain program, e.g. clockdiff uses
only libcap, but not IDN and NLS. OTOH certain headers are included for
all programs, thus kept for all.

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -918,6 +918,7 @@ int main(int argc, char **argv)
 			break;
 		case 'V':
 			printf(IPUTILS_VERSION("arping"));
+			print_config();
 			exit(0);
 		case 'h':
 		case '?':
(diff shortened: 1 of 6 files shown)


richiejp:
Could you just stringify and concatenate the macros instead?

Looks correct though.

pevik:
> Could you just stringify and concatenate the macros instead?

That was my first approach. But the decision is based on defined values and IMHO it's not possible to combine `#define ` and `#ifdef` (or `#if`). Or am I wrong?

richiejp:
You're probably right

pevik:
Maybe you mean so use single `printf()` call, right?

richiejp:
Yeah, but it is not important because it's only the version string. It doesn't work how I was thinking.

PR #448, opened 2023-01-05
Improve MTU probing support in tracepath and ping

csalinezh:
A regression in MTU probing feature of tracepath since 81d0f03d53138b3e100f60a50da50ec946ca866c is fixed, so that previous PMTU info cached in kernel FIB won't affect the result of tracepath. The symptom has been observed but regretfully I don't know how to reproduce it locally.

The possibility of sending such over-PMTU-sized probing packets in ping is added so users can play with different MTU manually with `ping -M probe`.

Better commit description is appreciated for both commits.

---

For the tracepath commit: Discrepancy already exists between `AF_INET6` and `AF_INET`, where the code for v6 will fallback to `IPV6_PMTUDISC_DO` if `IPV6_PMTUDISC_PROBE` is not available (probably due to old kernel versions), while code for v4 will fail if `IP_PMTUDISC_PROBE` is not available. Both options were introduced in Linux 2.6.22 (https://github.com/torvalds/linux/commit/628a5c561890a9a9a74dea017873530584aab06e). I wonder if the v4 version should do the fallback the same as v6 does.

For introducing the option into ping, I personally think the current option names to be confusing. But since they simply correspond to kernel API definition with the same name, `probe` is used in my patch following the convention. I considered using words like "ignore", "no", "off" or "force-ignore".

Documentation for `ping -M` is reworded a bit. Suggestions are appreciated.

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -351,10 +351,13 @@ xml:id="man.ping">
           <para>Select Path MTU Discovery strategy.
           <emphasis remap="I">pmtudisc_option</emphasis> may be
           either
-          <emphasis remap="I">do</emphasis> (prohibit fragmentation,
-          even local one),
+          <emphasis remap="I">do</emphasis> (set DF flag but
+          subject to PMTU checks by kernel, packets too large will
+          be rejected),
           <emphasis remap="I">want</emphasis> (do PMTU discovery,
-          fragment locally when packet size is large), or
+          fragment locally when packet size is large),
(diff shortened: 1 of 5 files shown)


pevik:
@csalinezh Thanks! LGTM, but I need a deeper look. It'd be nice to manage to reproduce it locally.
For kernel related question it'd be best to ask in netdev ML (http://vger.kernel.org/vger-lists.html#netdev).

@yoshfuji FYI

PR #450, opened 2023-02-06
ping: Fix the errno handling for strtod

skazi0:
The `setlocale(LC_ALL, "")` following the `strtod()` for the option 'i' can fail if the `LC_CTYPE` is invalid.
Hence the `errno` check following the `setlocale(LC_ALL, "")` thinks wrongly that `strtod()` failed with the `errno` and prints a warning. The `errno` got from the execution of `strtod()` is saved and restored after `setlocale()` to be checked for any errors.

Co-Developed-by: Sriram Rajagopalan <sriramr@arista.com>
Signed-off-by: Sriram Rajagopalan <sriramr@arista.com>
Signed-off-by: Jacek Tomasiak <jtomasiak@arista.com>

**To reproduce run e.g.**:
```
% LC_ALL=XXX sudo ping 8.8.8.8 -i 1.9 -c1
ping: option argument contains garbage: 
ping: this will become fatal error in the future
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=58 time=1.34 ms

--- 8.8.8.8 ping statistics ---
rtt min/avg/max/mdev = 1.339/1.339/1.339/0.000 ms
```

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -214,6 +214,7 @@ static double ping_strtod(const char *str, const char *err_msg)
 {
 	double num;
 	char *end = NULL;
+	int strtod_errno = 0;
 
 	if (str == NULL || *str == '\0')
 		goto err;
@@ -225,7 +226,10 @@ static double ping_strtod(const char *str, const char *err_msg)
 	 */
 	setlocale(LC_ALL, "C");
 	num = strtod(str, &end);
+	strtod_errno = errno;
 	setlocale(LC_ALL, "");
+	/* Ignore setlocale() errno (e.g. invalid locale in env). */
+	errno = strtod_errno;
 
 	if (errno || str == end || (end && *end)) {
 		error(0, 0, _("option argument contains garbage: %s"), end);


pevik (approved):
```
```
Good catch, LGTM.

pevik:
nit: I guess Sriram would be better set as `Co-developed-by: <Sriram Rajagopalan sriramr@arista.com>` (you can fix it yourself and force push it or I can do it before merging as I'll have to add my GPG sign and RBT, thus I need to amend the commit anyway).

skazi0:
I added `Co-Developed-by:` and left `Signed-off-by:` as suggested by https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by
Feel free to edit this as needed when merging.

pevik:
@skazi0 While the code makes sense, I'm not able to reproduce the warning with my setup (tested even with user `nobody` to get the default setup). Would you please post the output of following?

```
locale
echo $SHELL
cat /etc/os-release
```
I suppose manage to reproduce the problem on current master.

skazi0:
Ah, right, forgot to mention this. I hit this problem on RH system and couldn't reproduce it on Ubuntu. Spent some hours tracking it down. There is an old bug in `glibc` which was not fixed upstream yet. The fix is here: https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc-rh827510.patch Without it, only first call to `setlocale()` will try to load the file and set `errno` after failure.

To answer your question directly, here's the output:
```
/ @jtomasiak-r9test.sjc% LC_ALL=XXX sudo ping 8.8.8.8 -i 1.9 -c1
ping: option argument contains garbage: 
ping: this will become fatal error in the future
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=58 time=1.24 ms

--- 8.8.8.8 ping statistics ---
rtt min/avg/max/mdev = 1.237/1.237/1.237/0.000 ms
/ @jtomasiak-r9test.sjc% locale
LANG=C
LC_CTYPE="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_COLLATE="C"
LC_MONETARY="C"
LC_MESSAGES="C"
LC_PAPER="C"
LC_NAME="C"
LC_ADDRESS="C"
LC_TELEPHONE="C"
LC_MEASUREMENT="C"
LC_IDENTIFICATION="C"
LC_ALL=C
/ @jtomasiak-r9test.sjc% echo $SHELL
/bin/bash
/ @jtomasiak-r9test.sjc% cat /etc/os-release 
NAME="CentOS Stream"
VERSION="9"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
PRETTY_NAME="CentOS Stream 9"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:centos:9"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

PR #451, opened 2023-02-26
fix potential memory leak

eaglegai:
arping:
valgrind --leak-check=full  --show-leak-kinds=all arping 99.99.99.99  -c 2 ==2757117== Memcheck, a memory error detector
==2757117== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al. ==2757117== Using Valgrind-3.16.0 and LibVEX; rerun with -h for copyright info ==2757117== Command: arping 99.99.99.99 -c 2
==2757117==
Interface "lo" is not ARPable
==2757117==
==2757117== HEAP SUMMARY:
==2757117==     in use at exit: 2,224 bytes in 1 blocks
==2757117==   total heap usage: 78 allocs, 77 frees, 28,608 bytes allocated
==2757117==：
==2757117== 2,224 bytes in 1 blocks are still reachable in loss record 1 of 1
==2757117==    at 0x4841971: calloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==2757117==    by 0x49896E8: getifaddrs_internal (ifaddrs.c:417)
==2757117==    by 0x498A1A7: getifaddrs (ifaddrs.c:837)
==2757117==    by 0x10BD64: check_device (arping.c:631)
==2757117==    by 0x10CD5B: main (arping.c:971)
==2757117==
==2757117== LEAK SUMMARY:
==2757117==    definitely lost: 0 bytes in 0 blocks
==2757117==    indirectly lost: 0 bytes in 0 blocks
==2757117==      possibly lost: 0 bytes in 0 blocks
==2757117==    still reachable: 2,224 bytes in 1 blocks
==2757117==         suppressed: 0 bytes in 0 blocks
==2757117==
==2757117== For lists of detected and suppressed errors, rerun with: -s
==2757117== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)

tracepath:
==2943644== 65,535 bytes in 1 blocks are still reachable in loss record 9 of 9
==2943644==    at 0x483F751: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==2943644==    by 0x10A88E: main (in /usr/bin/tracepath)

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -609,6 +609,10 @@ int main(int argc, char **argv)
 
  done:
 	freeaddrinfo(result);
+	if(ctl.pktbuf) {
+		free(ctl.pktbuf);
+		ctl.pktbuf = NULL;
+	}
 
 	printf(_("     Resume: pmtu %d "), ctl.mtu);
 	if (ctl.hops_to >= 0)
--- a/arping.c
+++ b/arping.c
@@ -587,6 +587,7 @@ static int check_ifflags(struct run_state const *const ctl, unsigned int ifflags
 		if (ctl->device.name != NULL) {
 			if (!ctl->quiet)
 				printf(_("Interface \"%s\" is down\n"), ctl->device.name);
+			freeifaddrs(ctl->ifa0);
 			exit(2);
 		}
 		return -1;
@@ -595,6 +596,7 @@ static int check_ifflags(struct run_state const *const ctl, unsigned int ifflags
 		if (ctl->device.name != NULL) {
 			if (!ctl->quiet)
 				printf(_("Interface \"%s\" is not ARPable\n"), ctl->device.name);
+			freeifaddrs(ctl->ifa0);
 			exit(ctl->dad ? 0 : 2);
 		}
 		return -1;


eaglegai:
@pevik @kerolasa could you review these commits?

pevik:
@metan-ucw TL;DR any hint how to solve `freeifaddrs(ctl->ifa0)` before any exit on error handling?

arping needs another `freeifaddrs()` before any exit after `check_device()` call which calls `getifaddrs(&ctl->ifa0)`, e.g before quit in case of zero `sll_halen` in `main()`:
https://github.com/iputils/iputils/blob/33e78be2e60ed9ac918dec13271d1bd9dce6e94e/arping.c#L1037-L1040

Then later there is exit without previous `freeifaddrs()` in case of source address in not-DAD mode:
https://github.com/iputils/iputils/blob/33e78be2e60ed9ac918dec13271d1bd9dce6e94e/arping.c#L1052-L1054

Also `event_loop()` has several returns before final `freeifaddrs(ctl->ifa0)`:
https://github.com/iputils/iputils/blob/33e78be2e60ed9ac918dec13271d1bd9dce6e94e/arping.c#L737-L846

I wonder if moving `freeifaddrs(ctl->ifa0)` to main loop would solve at least the returns before `freeifaddrs(ctl->ifa0)`:

```
int rc;
...
rc = event_loop(&ctl);
freeifaddrs(ctl->ifa0);

return rc;
```
But not sure how to solve exit via `error()` in `drop_capabilities()`, there would have to be a wrapper over `error()`, `drop_capabilities()` would need to get `struct run_state *ctl`.

tracepath.c
```diff
@@ -609,6 +609,10 @@ int main(int argc, char **argv)
 
  done:
 	freeaddrinfo(result);
+	if(ctl.pktbuf) {
+		free(ctl.pktbuf);
+		ctl.pktbuf = NULL;
```
pevik, line 614: What is the reason to set `ctl.pktbuf` `NULL`? It's not being used later.
Also, there is another `exit()` via `error()`, which does not call `free(ctl.pktbuf)`:

https://github.com/iputils/iputils/blob/33e78be2e60ed9ac918dec13271d1bd9dce6e94e/tracepath.c#L570-L589

eaglegai: no special reason，just used to it. feel free to delete "set ctl.pktbuf NULL".

about exit by errors but no free(ctl.pktbuf), i think we could change to "error(0, errno, "IPV6_UNICAST_HOPS");  free(ctl.pktbuf); exit(1);"

I will modify the PR later

eaglegai:
about drop_capabilities()，why not modify "static void drop_capabilities(void)" to "static int drop_capabilities(void)"？
then exit and free  in main() ,like:
static int drop_capabilities(void)
{
        int ret = 0;
        cap_t cap_p = cap_init();

        if (!cap_p) {
                error(0, errno, "cap_init");
                ret = -1;
        }

        if (cap_set_proc(cap_p) < 0) {
                error(0, errno, "cap_set_proc");
                ret = -1;
        }

        cap_free(cap_p);
        return ret;
}
......
        if (drop_capabilities() < 0) {
                freeifaddrs(ctl.ifa0);
                exit(-1);
        }

pevik:
@eaglegai that could work. I'll wait for @metan-ucw input in case there is more elegant solution.

eaglegai:
> @eaglegai that could work. I'll wait for @metan-ucw input in case there is more elegant solution.

OK

metan-ucw:
So first of all, why do we even care about leaving buffers if we call `exit()` right after we print the error message? The whole program memory will be teared down by the kernel anyways and this will just slow down the program exit by a small bit.

So in my opinion it does not make that much sense to invest time into silencing the valgrind output here.

pevik:
@metan-ucw Well, I thought without calling `freeifaddrs()` there will be a memory leak, quote [`man freeifaddrs(3)`](https://linux.die.net/man/3/freeifaddrs): The data returned by getifaddrs() is dynamically allocated and should be freed using freeifaddrs() when no longer needed.

Looking at simple [musl implementation](https://git.musl-libc.org/cgit/musl/tree/src/network/getifaddrs.c), memory first [get's allocated](https://git.musl-libc.org/cgit/musl/tree/src/network/getifaddrs.c#n127) for `struct ifaddrs_storage` and `freeifaddrs()` does [call `free()` for `struct ifaddrs`). Am I missing something obvious?

pevik:
@eaglegai FYI correct way is to rebased code, not merge master (we prefer linear git history). But I'll need to amend your commits in order to add my gpg key anyway (rule for the project), thus I'll fix it.

metan-ucw:
@pevik No there will be no leak. Have a look at a simpler example

```
int main(void)
{
    void *ptr = malloc(1);
    exit(0);
}
```

Does the code free the memory allocated by malloc? No.
Does the program trigger valgrind memory leak detection? Yes.
Does the program leak memory? No.

The point is that when program calls `exit()` all program memory is freed.

eaglegai:
> @eaglegai FYI correct way is to rebased code, not merge master (we prefer linear git history). But I'll need to amend your commits in order to add my gpg key anyway (rule for the project), thus I'll fix it.

feel free to change

eaglegai:
> @pevik No there will be no leak. Have a look at a simpler example
> 
> ```
> int main(void)
> {
>     void *ptr = malloc(1);
>     exit(0);
> }
> ```
> 
> Does the code free the memory allocated by malloc? No. Does the program trigger valgrind memory leak detection? Yes. Does the program leak memory? No.
> 
> The point is that when program calls `exit()` all program memory is freed.

yes, but users  malloc memory and free by itself is better, instead of relying on the kernel.

metan-ucw:
@eaglegai You are relying on the kernel to map and unmap all of the program memory, if that stops working, the whole operating system is doomed... And actually freeing memory before exit() is even considered harmful by some developers, because the only thing that it achieves is to burn CPU time pointlessly.

nmeyerhans:
+1 to what @metan-ucw said. There is never a need to explicitly free heap memory before exiting a program.  That just involves extra userspace code execution that is completely redundant.  The kernel's behavior in this regard is defined by POSIX and has been established for decades.

eaglegai:
@metan-ucw，@nmeyerhans，what you said is very reasonable.
do we really not need to refer to this guide https://linux.die.net/man/3/getifaddrs ? it says “The data returned by getifaddrs() is dynamically allocated and should be freed using freeifaddrs() when no longer needed”…

nmeyerhans:
>  do we really not need to refer to this guide https://linux.die.net/man/3/getifaddrs ? it says “The data returned by getifaddrs() is dynamically allocated and should be freed using freeifaddrs() when no longer needed”…

Correct. Implicit in that statement is "or exit the program", because the implicit behavior any time a program exits is that its memory is reclaimed by the OS.

eaglegai:
> > do we really not need to refer to this guide https://linux.die.net/man/3/getifaddrs ? it says “The data returned by getifaddrs() is dynamically allocated and should be freed using freeifaddrs() when no longer needed”…
> 
> Correct. Implicit in that statement is "or exit the program", because the implicit behavior any time a program exits is that its memory is reclaimed by the OS.

but in it's example,it  uses freeifaddrs(ifaddr)  befor  exit(EXIT_SUCCESS),may it doesn't think so, about “ Implicit in that statement is "or exit the program"”

nmeyerhans:
> > > do we really not need to refer to this guide https://linux.die.net/man/3/getifaddrs ? it says “The data returned by getifaddrs() is dynamically allocated and should be freed using freeifaddrs() when no longer needed”…
> > 
> > 
> > Correct. Implicit in that statement is "or exit the program", because the implicit behavior any time a program exits is that its memory is reclaimed by the OS.
> 
> but in it's example,it uses freeifaddrs(ifaddr) before exit(EXIT_SUCCESS),may it doesn't think so, about “ Implicit in that statement is "or exit the program"”

It's an example showing how to free the memory explicitly, so it needs to do that even if it's redundant.  But it's still redundant.

pevik:
OK, thanks to you all, closing PR.

PR #457, opened 2023-04-18
ci: Add CentOS Stream 9 and Rocky Linux 9

robert-scheck:
CentOS Stream 9 is the current development branch of CentOS and also the upstream of Red Hat Enterprise Linux 9, where Rocky Linux 9 is a 1:1 RHEL clone/rebuild of RHEL 9 (like CentOS Linux was previously).

This is unfortunately a mess: The CI for CentOS Linux 8 got removed from iputils with commit 931504a096788289d40aafadeeb7a844e223e567, because Red Hat terminated CentOS Linux 8 (EOL: 2021-12-31) in favor of CentOS Stream 8. However, there are no official CentOS Stream images on Docker Hub at all, only at Red Hat's Quay.io.

While I suggest hereby to add CentOS Stream 9 and Rocky Linux 9 (latter to cover RHEL 9), you might maybe dislike the quirks needed for CentOS Stream 9 via Quay.io. Adding CentOS Stream 8 does not make sense, because it reaches EOL at 2024‑05‑31. If at all, Rocky Linux 8 could make sense, because it reaches EOL at 2029-05-31.

Just let me know if you would like to see changes to this pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,14 @@ jobs:
             env:
               CC: clang
 
+          - container: "quay.io/centos/centos:stream9"
+            env:
+              CC: gcc
+
+          - container: "rockylinux:9"
+            env:
+              CC: gcc
+
           - container: "centos:7"
             env:
               CC: gcc
@@ -94,6 +102,7 @@ jobs:
       run: |
         export INSTALL=${{ matrix.container }}
         export DISTRO_VERSION="${INSTALL#*:}"
+        INSTALL="${INSTALL#quay.io/centos/}"
         INSTALL="${INSTALL%%:*}"
         INSTALL="${INSTALL%%/*}"
         ./ci/$INSTALL.sh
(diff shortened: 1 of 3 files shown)


robert-scheck:
Okay, then I'm adding Rocky Linux 8 (to cover RHEL 8), too.  And indeed, feel free to assign future issues related to building on CentOS and/or Rocky Linux to me. Note that I'm not a C programmer, I'm a packager in my free time for Fedora (and implicitly for its downstreams like CentOS Stream).

pevik:
Rocky Linux 8 failure: `Error: Unable to find a match: docbook5-style-xsl`.

robert-scheck:
Yes, I should have knewn it. Sorry! There is only `docbook-style-xsl` in CentOS/RHEL/Rocky 8. I'm reverting this for now and will add `docbook5-style-xsl` to EPEL 8 (and then come back with another pull request).

PR #458, opened 2023-04-19
ci/alpine.sh: Add libcap-utils

robert-scheck:
Just adding `libcap-utils` (provides `setcap`; to match other distributions), because adding `libidn2-dev` as well lets the tests fail…

### DIFF
--- a/ci/alpine.sh
+++ b/ci/alpine.sh
@@ -13,6 +13,7 @@ apk add \
 	iproute2 \
 	gettext-dev \
 	libcap-dev \
+	libcap-utils \
 	libxslt \
 	make \
 	meson \


pevik:
> provides setcap; to match other distributions)

I'm not sure if the goal is to install everything which _could_ be needed for the development, but it's not needed for CI. Because in CI it was disabled in 9cec7363, I just added single test for Debian.

Therefore I would tend to do the opposite: to remove `libcap-progs` from opensuse.sh (i.e. remove explicit installation of unneeded tools). But I agree that it can be useful for documentation purposes (my main approach for documentation purposes was CentOS, where I found really difficult to install correct packages due EPEL, but I expect anybody can find `setcap` in his distro).

robert-scheck:
Oh, sorry…let's close this then simply.

PR #462, opened 2023-05-05
Update Debian CI environments

nmeyerhans:
This PR refreshes the Debian environments used in the CI configuration, as they've been out of date with respect to actual Debian releases for some time.

Use static release names corresponding to specific releases, rather than moving targets.  In other words, replace names like "stable" and "testing" with specific distros, "bullseye", "bookworm", etc.

Remove configuration that assumed that "oldstable" referred to Debian 9, as this hasn't been the case since the Debian 11 release in mid-2021 and Debian 9 is fully EOL at this point anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,19 +12,19 @@ jobs:
       matrix:
         include:
           # cross compilation builds
-          - container: "debian:testing"
+          - container: "debian:bookworm"
             env:
               ARCH: arm64
               CC: aarch64-linux-gnu-gcc
               VARIANT: cross-compile
 
-          - container: "debian:testing"
+          - container: "debian:bookworm"
             env:
               ARCH: ppc64el
               CC: powerpc64le-linux-gnu-gcc
               VARIANT: cross-compile
 
-          - container: "debian:testing"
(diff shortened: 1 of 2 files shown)


nmeyerhans:
Note that this will address the test failure seen in #459

pevik:
Well, I intended it to be moving target, so that I didn't have to keep updating it (I'd easily forget). Yes, sometimes Debian release requires adjustments, but very rarely. I'd actually prefer have way to have moving targets for other distros, but they have IMHO only "unstable" equivalent. Would you mind to amend your commit?

pevik:
@nmeyerhans would you please update the PR to keep the generic names? Or shell I do it on your behalf? I need to fix CI for https://github.com/iputils/iputils/issues/465 fix.

pevik:
@nmeyerhans would you please update the PR to keep the generic names? Or shell I do it on your behalf? I need to fix CI for https://github.com/iputils/iputils/issues/465 fix.

pevik:
@nmeyerhans If you don't mind, I'd prefer to merge just this change:
https://github.com/pevik/iputils/commit/c9ad9e155bbc2f23721e57d87f3acf66c51e0400

Issue #465, opened 2023-05-15
ping negative interval


cronolio:
looks like ping allows to use negative interval like `ping -i -0.5 foo.bar`. With this ping produce a lot of packages...

example output
```
ping -ni -0.5 -OD 8.8.8.8
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
[1684137776.503110] no answer yet for icmp_seq=1
[1684137776.503132] no answer yet for icmp_seq=2
[1684137776.503141] no answer yet for icmp_seq=3
[1684137776.503150] no answer yet for icmp_seq=4
[1684137776.503158] no answer yet for icmp_seq=5
[1684137776.503167] no answer yet for icmp_seq=6
[1684137776.503175] no answer yet for icmp_seq=7
[1684137776.503209] no answer yet for icmp_seq=8
[1684137776.503218] no answer yet for icmp_seq=9
[1684137776.503226] no answer yet for icmp_seq=10
[1684137776.503235] no answer yet for icmp_seq=11
[1684137776.503243] no answer yet for icmp_seq=12
[1684137776.503252] no answer yet for icmp_seq=13
[1684137776.519204] no answer yet for icmp_seq=13
[1684137776.519265] no answer yet for icmp_seq=14
```

```
ping from iputils 20221126
libcap: yes, IDN: yes, NLS: no, error.h: yes, getrandom(): yes, __fpending(): yes
```
```
Debian GNU/Linux 12 (bookworm)
Linux <edited> 6.1.0-7-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.20-2 (2023-04-08) x86_64 GNU/Linux
```

pevik:
@cronolio Would you please provide your real name and your email? I'd like to add your credit as `Reviewed-by: Your Name <your_email@example.com>`. You find quite old regression - from 918e824dc13a39e4d68fcd82fd2d248c9fba6bbd (released in s20190324).

cronolio:
actually you can't check a name and an email is really belong to me.
so nickname is enough. thank you.

pevik:
We follow the Linux kernel policy, which expect real name (I'd be ok with the nickname) _and_ real email:
https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

But sure, you don't have to provide it.

cronolio:
[...]

PR #466, opened 2023-05-16
ping: Fix overflow on negative -i

pevik:
Restore back check for -i <= 0.

Fixes: 918e824 ("ping: add support for sub-second timeouts")
Closes: https://github.com/iputils/iputils/issues/465


pevik:
BTW we need to first check CI failure: https://github.com/iputils/iputils/pull/462.

PR #468, opened 2023-05-18
ping6: Improve handling of option '-Q'.

gault:
Using ping6 -Q on a system using ip-rules can fail because the probe_fd
doesn't set IPV6_TCLASS. Therefore, probe_fd might make its route
lookup in a different table than the one that will be used to really
send the packet. This is fixed by patch 1.

Patch 2 then cleans up some dead code also related to IPV6_TCLASS.

### DIFF
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -372,15 +372,6 @@ int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 	   )
 		error(2, errno, _("can't receive hop limit"));
 
-	if (rts->opt_tclass) {
-#ifdef IPV6_TCLASS
-		if (setsockopt(sock->fd, IPPROTO_IPV6, IPV6_TCLASS, &rts->tclass, sizeof rts->tclass) == -1)
-			error(2, errno, _("setsockopt(IPV6_TCLASS)"));
-#else
-		error(0, 0, _("traffic class is not supported"));
-#endif
-	}
-
 	if (rts->opt_flowinfo) {
 		char freq_buf[CMSG_ALIGN(sizeof(struct in6_flowlabel_req)) + rts->cmsglen];
 		struct in6_flowlabel_req *freq = (struct in6_flowlabel_req *)freq_buf;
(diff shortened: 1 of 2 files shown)


pevik:
Thank you. LGTM, I just need time to do some testing.

gault:
> Thank you. LGTM, I just need time to do some testing.

Thanks @pevik. Do you have any feedback for these patches?

pevik:
@gault I'm sorry to keep you waiting so long. I'll try to have look this week.

pevik:
@gault I'm ok to require kernel fixes, because it was merged in 2023-02-22 to even v4.14.y.

And yes, following code really works with your changes:

```
ip route add table main unreachable 2001:db8::10/124
ip route add table 100 2001:db8::10/124 dev eth0
ip -6 rule add dsfield 0x04 table 100
ping -Q 0x04 2001:db8::11
```

Would you mind to share other setup required for ping to be successfully get reply? Ideally using network namespaces. It could be used in the testing suite I'm planning to write.

ping/ping6_common.c
```diff
@@ -372,15 +372,6 @@ int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
...
 
-	if (rts->opt_tclass) {
-#ifdef IPV6_TCLASS
-		if (setsockopt(sock->fd, IPPROTO_IPV6, IPV6_TCLASS, &rts->tclass, sizeof rts->tclass) == -1)
-			error(2, errno, _("setsockopt(IPV6_TCLASS)"));
-#else
-		error(0, 0, _("traffic class is not supported"));
-#endif
-	}
-
```
pevik, line 379: @gault I wonder if 78aa5c1516d6f299331b4cd0efac5f0f53a28c0f brought the regression or if it was later by some Pavel Simerda changes.

pevik:
To both commits:
```
```

ping/ping6_common.c
gault: I think it's commit ebad35fee3de ("ping: merge `ping6` command into `ping`"). It removed the "options |= F_TCLASS;" statement that was in ping6_main(), but didn't remove the "if (options & F_TCLASS) {" test in ping6_run().

I didn't put any Fixes: tag because that was just dead code elimination. But if we want to point to a commit, I think that should be ebad35fee3de.

gault:
Thanks pevik!

Issue #469, opened 2023-05-22
tracepath / Documentation: 


marek22k:
In the man pages of Tracepath it says `async` under [output](https://manned.org/tracepath.8#head5), but this never appears in the output. Could it be that `asymm` is meant here?

Furthermore it says `the difference between number of hops in forward and return direction is`. As far as I could observe, it is **not** the **difference** between the hops on the outbound and return path, but only the **number of hops on the return path** without `difference`.

PR #470, opened 2023-05-22
update tracepath documentation

marek22k:
Fixes https://github.com/iputils/iputils/issues/469


pevik:
LGTM, given the code is in https://github.com/iputils/iputils/blob/2a63b94313352ed62f47adbc1a0d33bfea7ca188/tracepath.c#L323-L328, thus merged with adjusted commit message.
Thank you!

PR #471, opened 2023-05-24
Tracepath cleanup

pevik:
Simple update in po message.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: iputils\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-03-06 21:32+0100\n"
+"POT-Creation-Date: 2023-05-24 10:45+0200\n"
 "PO-Revision-Date: 2022-10-19 23:20+0000\n"
 "Last-Translator: Robert Scheck <redhat@linuxnetz.de>\n"
 "Language-Team: German <https://translate.fedoraproject.org/projects/iputils/"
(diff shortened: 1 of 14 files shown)


pevik:
@marek22k Could you please review this?

po/de.po
```diff
@@ -1382,31 +1382,31 @@ msgstr ""
...
+#: tracepath.c:615
 #, c-format
 msgid "back %d "
 msgstr ""
 
-#: tracepath.c:622
+#: tracepath.c:620
 #, c-format
 msgid "pktlen must be within: %d < value <= %d"
 msgstr ""
```
marek22k, line 1412: I think the best here would be `pktlen muss innerhalb von %d < Wert <= %d liegen`.

Or alternatively, if that doesn't work: `pktlen muss innerhalb der Grenzen liegen: %d < value <= %d` (`pktlen must be within the limits: %d < value <= %d`) or `pktlen muss innerhalb der Werte liegen: %d < Wert <= %d`

marek22k (commented):
In the some translation, some passages are also missing. Is that then simply ignored if it is empty? I myself, can only translate German. (Is git the right place for that?)

The code is also correct as far as I can tell.

Otherwise, I honestly don't know what to say about it / how to review it.

pevik:
@marek22k Thanks for your comments about German translation (comments are now not visible, not sure why). Translation is done via https://translate.fedoraproject.org/projects/iputils/iputils/. Although it's the preferred way to translate, but if you have time to improve the translation and don't want to register on the web, I'd be ok if you open a pull request.

pevik:
Merged in 608c62b3967882d31ff9edde97e38de0fcff154b and 986859ac1d83ae985d3467a79bf9daa1385ed146.

marek22k:
I removed the translations again and put them into Weblate.

marek22k:
I would be done with the translation. For many strings, correct translations also already be suggested. I hope I have translated everything correctly :-)

Issue #421, opened 2022-08-28
ping: do a reverse lookup for numeric addresses


rnhmjoj:
It [seems](https://github.com/iputils/iputils/commit/f6bc8f6f01498527307e8c9580c1588adbd0c818) that if you specify a "literal" IP address ping won't do a reverse name lookup. I think this would be good to have in general and it's particularly nice for IPv6: if you have set up a reverse zone for the link local you can then do
```
$ ping ff02::1%eth0 -c2
PING ff02::1%eth0(ff02::1%eth0 (ff02::1%eth0)) 56 data bytes
64 bytes from me.home.arpa (fe80::46f5:36ff:fe71:2081%eth0): icmp_seq=1 ttl=64 time=0.080 ms
64 bytes from router.home.arpa (fe80::3e48:edff:fe5a:50ab%eth0): icmp_seq=1 ttl=64 time=0.450 ms
64 bytes from ap-1.home.arpa (fe80::864c:80ff:fe0a:3058%eth0): icmp_seq=1 ttl=64 time=0.798 ms
64 bytes from ap-2.home.arpa (fe80::6231:4eff:feb2:50d3%eth0): icmp_seq=1 ttl=64 time=6.92 ms
64 bytes from me.home.arpa (fe80::46f5:36ff:fe71:2081%eth0): icmp_seq=2 ttl=64 time=0.083 ms

--- ff02::1%eth0 ping statistics ---
2 packets transmitted, 2 received, +3 duplicates, 0% packet loss, time 1001ms
rtt min/avg/max/mdev = 0.080/1.665/6.916/2.638 ms
```
and get a nice scan of your network.

It's enough to remove this check to get this behavior, not sure if there are other side effects.
```patch
diff --git a/ping/ping6_common.c b/ping/ping6_common.c
index 13bf45a..39f7d59 100644
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -132,9 +132,6 @@ int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 	memcpy(&rts->whereto6, ai->ai_addr, sizeof(rts->whereto6));
 	rts->whereto6.sin6_port = htons(IPPROTO_ICMPV6);
 
-	if (memchr(target, ':', strlen(target)))
-		rts->opt_numeric = 1;
-
 	if (IN6_IS_ADDR_UNSPECIFIED(&rts->firsthop.sin6_addr)) {
[...]

PR #485, opened 2023-08-17
ping: Use print target when empty ai_canonname

pevik:
On certain network setup getaddrinfo() can return empty ai_canonname. Instead of printing nothing in "PING" line use the target.

This fixes the output of PING line:
```
$ ping seznam.cz -c1 -4 -v
...
ai->ai_family: AF_INET6, ai->ai_canonname: 'seznam.cz' ai->ai_family: AF_INET6, ai->ai_canonname: ''
ai->ai_family: AF_INET, ai->ai_canonname: ''
PING  (77.75.79.222) 56(84) bytes of data.
64 bytes from www.seznam.cz (77.75.79.222): icmp_seq=1 ttl=55 time=4.82 ms
```

to
```
$ ping seznam.cz -c1 -4 -v
...
PING seznam.cz (77.75.79.222) 56(84) bytes of data. 64 bytes from www.seznam.cz (77.75.79.222): icmp_seq=1 ttl=55 time=4.88 ms
```

Link: https://github.com/iputils/iputils/pull/478#issuecomment-1676119425
Reported-by: Robert Scheck <robert@fedoraproject.org>


pevik:
@nmeyerhans @robert-scheck Could you please have a look? Shouldn't we add something after `target` (e.g. `seznam.cz (guess)`)  to show there is `getaddrinfo()` issue?

robert-scheck (commented):
Personally I think, `ping` should not use `getaddrinfo()` for the target name at all, because I do not see any value in reverse DNS lookups.

Example:

```
$ ping <a>
PING <a> (<b>) 56(84) bytes of data.
64 bytes from <a> (<b>): icmp_seq=1 ttl=55 time=4.82 ms
```

- `<a>` is always the same string as passed on the command line
- If `<a>` is already an IP address: `<a>` equals `<b>`
- Else: `<b>` is the result of the forward DNS lookup of `<a>`

But this is my point of view. Your previously shown output is somehow similar in the optical result in various cases and thus seems to be fine to me.

pevik:
There can be `<c>` (here `prodcs.lwn.net`) which is different from `<a>` as a result of reverse DNS lookup:
```
$ ping lwn.net -c2
PING lwn.net (173.255.236.65) 56(84) bytes of data.
64 bytes from prodcs.lwn.net (173.255.236.65): icmp_seq=1 ttl=47 time=109 ms
64 bytes from prodcs.lwn.net (173.255.236.65): icmp_seq=2 ttl=47 time=109 ms
```

I suppose this is a primary reason why the second `getnameinfo()` is called (unless `-f` or `-n` or IP address specified):

https://github.com/iputils/iputils/blob/bc3f2e328c79c75a9e369bd228a75383f2e71bdf/ping/ping.c#L1770-L1772

I'm still not convinced, that `-n` behavior should be the default (although somebody has suggested it recently in comment in issue or pull request, but I cannot find it).

pevik:
BTW `getnameinfo()` code was added in 7ddd66a5e9a6d42359ae56282e26daed63421ce5, but the behavior is the same as the previous code, which used legacy `gethostbyaddr()`.

pevik:
@robert-scheck FYI I implemented the opposite of `-n`: https://github.com/iputils/iputils/issues/421

metan-ucw:
Uff, this is hard to decide. Reverse lookup is how ping behaved for decades, if anyone changes that there will be a lot of surprises. I suppose that even copying the target as this patch does could be seen as controversial by someone...

ping/ping.c
```diff
@@ -787,8 +787,15 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
...
-				strncpy(hnamebuf, result->ai_canonname, sizeof hnamebuf - 1);
+
+			/*
+			 * On certain network setup getaddrinfo() can return empty
+			 * ai_canonname. Instead of printing nothing in "PING"
+			 * line use the target.
+			 */
+			strncpy(hnamebuf, result->ai_canonname ? result->ai_canonname :
+					target, sizeof hnamebuf - 1);
+
```
metan-ucw, line 798: Why can't we do this in the else branch?

```
/*
 * ....
 */
if (result->ai_cannoname)
        strcpy(..., result->ai_cannonname, ...);
else
       strcpy(..., target, ...);
```

pevik:
> Uff, this is hard to decide. Reverse lookup is how ping behaved for decades, if anyone changes that there will be a lot of surprises. I suppose that even copying the target as this patch does could be seen as controversial by someone...

Yes, that's exactly the reason I'd prefer just to fix the output here. Unfortunately we maintain a legacy product.

ping/ping.c
pevik: Fixed, thanks!

Issue #484, opened 2023-07-28
ping thinks it's sent more packets than it thinks with -i 0.002 -c 1 -w 1


JasonWoof:
I've got iputils 20221106 on nixos on an xps 13.

When I pass a small value for -i, and also pass -c 1 -w 1, then ping does the math as though it has sent more packets than it actually has (or at least more than it reports, and more than it should.)  I don't always get the same number of extra packets in the math.

Example:

```
> ping -i 0.002 8.8.8.8 -c 1 -w 1
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=117 time=76.4 ms

--- 8.8.8.8 ping statistics ---
8 packets transmitted, 1 received, 87.5% packet loss, time 71ms
rtt min/avg/max/mdev = 76.424/76.424/76.424/0.000 ms, pipe 8
```

Note that on the third line ping shows that it's sent one packet (as it should) but then on the 6th line it says "8 packets transmitted" where it should say "1 packets transmitted" and the math on the rest of that line is wrong (as though it had actually sent 8 packets).

I have not done any network sniffing to detect the actual number of packets sent.

Another more example:

```
> ping -i 0.01 8.8.8.8 -c 1 -w 1
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=117 time=820 ms

--- 8.8.8.8 ping statistics ---
39 packets transmitted, 1 received, 97.4359% packet loss, time 814ms
rtt min/avg/max/mdev = 819.761/819.761/819.761/0.000 ms, pipe 39
```

39?!? :)

The variability in the number of imagined packets is variable, and only occurs when the value passed with -i is sufficiently small.

I wonder if this is influenced by system load or network congestion.

This is consistently reproduceable for me with -i values from 0.002 to 0.01. at 0.05 it works correctly sometimes and fails others. at 0.1 it works correctly most of the time (80-90%) and at 1.0 I haven't seen it bug yet.

pevik:
Thank you for the report, it's likely a bug. Unfortunately I will not have time to look into it before the end of August.

pevik:
> I've got iputils 20221106 on nixos on an xps 13.

@JasonWoof I suppose this is not nixos specific at all, right? And you probably mean `20221126`.
It'd be great if you could bisect to find a problematic commit when the behavior started.

[...]

PR #486, opened 2023-08-24
man/ping: Mention broadcast and multicast limit for non-root

pevik:
Mention broadcast and multicast limit for non-root.

Link: https://github.com/iputils/iputils/issues/484#issuecomment-1691025991


pevik:
@robert-scheck could you please have a look?

PR #487, opened 2023-08-25
ping: Make interval error messages more readable

pevik:
Improve too low interval for user on broadcast, multicast and flood
error message.  Instead of printing "raw" interval (i.e. -i value *
1000) it's more user friendly to print requested -i value.

```
$ ./builddir/ping/ping -b 192.168.122.255 -i0.1
WARNING: pinging broadcast address
./builddir/ping/ping: minimal interval for broadcast ping for user must be >= 1000 ms, use -i 1 (or higher)

$ ./builddir/ping/ping -i0.1 ff02::1%tun0
./builddir/ping/ping: minimal interval for multicast ping for user must be >= 1000 ms, use -i 1 (or higher)
```
Too low -i value and -f for user without -i use the same message. Not
sure if this should be further improved:

```
$ ./builddir/ping/ping ::1 -f
PING ::1 (::1) 56 data bytes
./builddir/ping/ping: cannot flood, minimal interval for user must be >= 2 ms, use -i 0.002 (or higher)

$ ./builddir/ping/ping -c2 -i 0.0019 ::1
PING ::1 (::1) 56 data bytes
```
./builddir/ping/ping: cannot flood, minimal interval for user must be >= 2 ms, use -i 0.002 (or higher)

Modified strings are localized (translation update before release will be needed).

Based on discussion in: https://github.com/iputils/iputils/issues/484#issuecomment-1691025991
Related PR (changes in man page): https://github.com/iputils/iputils/pull/486

### DIFF
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -389,6 +389,7 @@ extern void drop_capabilities(void);
 
 char *pr_addr(struct ping_rts *rts, void *sa, socklen_t salen);
 char *pr_raw_addr(struct ping_rts *rts, void *sa, socklen_t salen);
+char *str_interval(int interval);
 
 int is_ours(struct ping_rts *rts, socket_st *sock, uint16_t id);
 extern int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock);
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -63,11 +63,11 @@
 #define	DEFDATALEN	(64 - 8)	/* default data length */
 
 #define	MAXWAIT		10		/* max seconds to wait for response */
-#define MININTERVAL	10		/* Minimal interpacket gap */
-#define MINUSERINTERVAL	2		/* Minimal allowed interval for non-root */
+#define MIN_INTERVAL_MS	10		/* Minimal interpacket gap */
+#define MIN_USER_INTERVAL_MS	2		/* Minimal allowed interval for non-root */
 #define IDENTIFIER_MAX	0xFFFF		/* max unsigned 2-byte value */
 
-#define SCHINT(a)	(((a) <= MININTERVAL) ? MININTERVAL : (a))
+#define SCHINT(a)	(((a) <= MIN_INTERVAL_MS) ? MIN_INTERVAL_MS : (a))
 
 
 #ifndef MSG_CONFIRM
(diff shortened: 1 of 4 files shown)


pevik:
@robert-scheck @yurchor can you please have a look? Does the string make sense to you?

yurchor:
> 
Looks good for me. No objections.

ping/ping_common.c
```diff
@@ -963,3 +965,18 @@ inline int is_ours(struct ping_rts *rts, socket_st * sock, uint16_t id)
...
+	static char buf[7];
+
+	/*
+	 * Here we always want to use locale regardless USE_IDN or ENABLE_NLS,
+	 * because we need do to print dot as decimal point of a number in -i option.
+	 */
+	setlocale(LC_ALL, "C");
+	sprintf(buf, "%.g", (double)interval / 1000);
+	setlocale(LC_ALL, "");
+
```
metan-ucw, line 980: If we wanted avoid messing with locales and floating point we can do something as:

```
        if (interval % 1000)
                printf("%1i.%03i", interval/1000, interval%1000);
        else
                printf("%i", interval/1000);
```

ping/ping.h
```diff
@@ -63,11 +63,11 @@
 #define	DEFDATALEN	(64 - 8)	/* default data length */
 
 #define	MAXWAIT		10		/* max seconds to wait for response */
-#define MININTERVAL	10		/* Minimal interpacket gap */
-#define MINUSERINTERVAL	2		/* Minimal allowed interval for non-root */
+#define MIN_INTERVAL	10		/* Minimal interpacket gap */
+#define MIN_USER_INTERVAL	2		/* Minimal allowed interval for non-root */
```
metan-ucw: If we are doing the rename we may as well add units to the name, e.g. `MIN_INTERVAL_MS` and `MIN_USER_INTERVAL_MS`

ping/ping_common.c
pevik: I got the idea. We need `char *`, due `error()`, thus:

```C
static char buf[14];
if (interval % 1000)
         sprintf(buf, "%1i.%03i", interval/1000, interval%1000);
else
         sprintf(buf, "%i", interval/1000);

return buf;
```

metan-ucw: Maybe snprintf(buf, sizeof(buf), ...) would be safer, on the other hand as long as interval is 32bit int it cannot overflow 14 characters...

pevik: Yep, it should be safe, but let's use `snprintf()` for extra security. Thanks!

Issue #489, opened 2023-09-19
DIFFERENT ADDRESS - problem with duplicit ident field


milo:
Hi. I'm aware that this is probably not an issue of this package, but I don't really know who to ask. So if you find it inappropriate please close the issue, I'll understand.

I'm running Debian 12 (Bookworm) with bundled kernel 6.1.0-11-amd64. I'm running Nagios 4.4.6-4 and as a `check-host-alive` I use your `ping` in version 20221126-1. I tried bundled kernel 5.10.0-25-amd64 with the same result.

Before upgrading to this setup, I used Debian 11 with kernel v4. There where no issues.

As a `check-host-alive` I use command:
```
/bin/ping4 -n -v -D -W 1 -i 1 -c 5 -M 'do' -s 56 -O "$Host" 2>&1
```
and when everything OK, I'll get (IP obfuscated):
```
/bin/ping4: sock4.fd: 3 (socktype: SOCK_RAW), sock6.fd: -1 (socktype: 0), hints.ai_family: AF_INET

ai-ai_family: AF_INET, ai-ai_canonname: 10.66.a.b
PING 10.66.a.b (10.66.a.b) 56(84) bytes of data.
[1695120545.794919] 64 bytes from 10.66.a.b: icmp_seq=1 ident=59225 ttl=63 time=0.563 ms
[1695120546.796904] 64 bytes from 10.66.a.b: icmp_seq=2 ident=59225 ttl=63 time=0.557 ms
[1695120547.826579] 64 bytes from 10.66.a.b: icmp_seq=3 ident=59225 ttl=63 time=0.532 ms
[1695120548.854606] 64 bytes from 10.66.a.b: icmp_seq=4 ident=59225 ttl=63 time=0.543 ms
[1695120549.874664] 64 bytes from 10.66.a.b: icmp_seq=5 ident=59225 ttl=63 time=0.540 ms

--- 10.66.a.b ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4080ms
rtt min/avg/max/mdev = 0.532/0.547/0.563/0.011 ms

Date/Time: Tue Sep 19 12:49:09 CEST 2023

```

But time to time, and that's the issue, I get `DIFFERENT ADDRESS` responses:
```
/bin/ping4: sock4.fd: 3 (socktype: SOCK_RAW), sock6.fd: -1 (socktype: 0), hints.ai_family: AF_INET

[...]

PR #490, opened 2023-09-21
ping: Print also the default ident

pevik:
There is no reason why not printing the default ident, which is -1, but printed via ntohs() as 65535.

Reported-by: Miloslav Hůla
Link: https://github.com/iputils/iputils/issues/489#issuecomment-1728967782
Fixes: 43e38f2 ("ping: Add option -e to set Identifier field for ICMP ECHO_REQUEST")


pevik:
@rumvadim Could you please review?
@milo FYI

rumvadim:
Hi!
This will not print the right ident, though. When the ident is -1 the ident for the group of requests is chosen by the kernel and most likely it's not 65535.

pevik:
Thanks for a hint, I need to verify that with `tcpdump`.

pevik:
@nmav Could you please also have look?

rumvadim:
The kernel code is here: https://elixir.bootlin.com/linux/latest/source/net/ipv4/ping.c#L78
ident=unknown seems fine to me, but I'm no expert in good interfaces though :sweat_smile:

Issue #479, opened 2023-06-25
Create distribution tarballs


thesamesam:
Would it be possible for iputils releases to be accompanied by distribution tarballs? Meson can generate these with `meson dist`.

The reason I'm asking is to have pre-generated documentation (man pages) in the tarball without needing additional dependencies to build them, but there's some other advantages like being able to sign them & not relying on github's auto-generated source tarballs which are unstable (they've changed a few times over the years).

I can do the wiring up for the man pages if you're up for it.

pevik:
Thanks a lot! This makes sense, but I'd start from the new release.

I'll update docs and create release script (there is already a lot of copy pasting from the docs).

thesamesam:
Thank you, this will be a big help!

pevik:
@thesamesam can you please have a look on https://github.com/iputils/iputils/pull/492?

pevik:
Fixed in 83f011ab2a1e8ebf6f6e9b00494ceb069ac42a63, 72a36873cc909abe51d3d57053fb786402ef734a.

pevik:
@thesamesam I'm sorry, I overlooked the main reason for this, to get pregenerated man pages.

pevik:
Finally found that simple `CC` pointing to cross compile toolchain was the problem, fixed and merged. Thanks!

PR #492, opened 2023-10-07
tools: Add scripts for makinga new release

pevik:
Adding two scripts: tools/tag-release.sh and tools/create-tarballs.sh.
See: https://github.com/iputils/iputils/issues/479

### DIFF
--- a/tools/create-tarballs.sh
+++ b/tools/create-tarballs.sh
@@ -1,0 +1,62 @@
+#!/bin/sh -eu
+# Copyright (c) 2023 Petr Vorel <pvorel@suse.cz>
+# Create tarballs and checksums for uploading after tagging a new release.
+
+basedir="$(dirname "$0")"
+. "$basedir/lib.sh"
+require_cmd date git gpg meson sha256sum
+
+tag="$(date +%Y%m%d)"
+tarball_dir="iputils-full-$tag"
+git_dir=$(cd "$basedir/.."; pwd)
+dir="$(cd "$git_dir/../"; pwd)/iputils-release-$tag"
+dist=builddir
+formats="xztar,gztar,zip"
+
+# check output directory
+if [ -d "$dir" ]; then
+	ask "Directory '$dir' exists, will be deleted"
+	rm -rf "$dir"
+fi
+rod mkdir "$dir"
+cd "$dir"
+dir="$PWD"
+
+# git clone (local)
+title "git clone"
+rod git clone "$git_dir" "$tarball_dir"
+rod cd "$tarball_dir"
+
+# check for tag
+if ! git show "$tag" 2> /dev/null; then
+	ask "Tag '$tag' does not exist, you should run tag-release.sh before, otherwise meson creates archive for $(git describe --abbrev=0)"
+fi
+
+# tarballs
+title "Generate tarballs"
+rod meson "$dist"
+rod meson dist -C "$dist" --formats "$formats"
+
+cd "$dist/meson-dist/"
+
+title "Generate tarballs and checksums"
+# meson puts '*' before filename in checksums (unusable for distros => need to
+# recreate them)
+rm -f ./*.sha256sum
+
+for file in *.*; do
+	gpg --sign --armor --detach-sign "$file"
+	sha256sum "$file" >> sha256sums
+done
+
+# sha256sums.asc (GPG signed checksums)
+gpg --clearsign sha256sums
+rm -f sha256sums
+
+mv -v ./* "$dir"
+
+echo
+title "Generated tarballs and checksums"
+cd "$dir" && ls -p | grep -v /
+echo
+echo "Files are in '$dir', upload them to github"
--- a/tools/lib.sh
+++ b/tools/lib.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2023 Petr Vorel <pvorel@suse.cz>
+
+# Confirm before proceeding, otherwise exit.
+ask()
+{
+	local msg="$1"
+	local answer
+
+	printf "\n%s. Proceed? [N/y]: " "$msg"
+	read -r answer
+	case "$answer" in
+		[Yy]*) : ;;
+		*) exit 2
+	esac
+}
+
+# Print error message and exit.
+quit()
+{
+	printf "\n%s\n" "$*" >&2
+	exit 1
+}
+
+# Check if commands are available, exit on first not found.
+require_cmd()
+{
+	local cmd
+
+	for cmd in "$@"; do
+		if ! command -v "$cmd" >/dev/null 2>&1; then
+			quit "'$cmd' not found"
+		fi
+	done
+}
+
+# "Run or die". Run command and print failing command and exit on failure.
+rod()
+{
+	"$@" || quit "$* failed"
+}
+
+# Print a header.
+title()
+{
+	echo "===== $1 ====="
+}
--- a/tools/tag-release.sh
+++ b/tools/tag-release.sh
@@ -1,0 +1,86 @@
+#!/bin/sh -eu
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2023 Petr Vorel <pvorel@suse.cz>
+# Tag iputils release.
+
+basedir="$(dirname "$0")"
+cd "$basedir/.."
+. "$basedir/lib.sh"
+
+require_cmd date git sed sort uniq
+
+upstream_git="iputils/iputils"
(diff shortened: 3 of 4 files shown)


pevik:
@nmeyerhans @vapier @robert-scheck Can you please have a look?

tools/lib.sh
```diff
@@ -0,0 +1,31 @@
...
+
+quit()
+{
+	printf "\n$@\n" >&2
+	exit 1
+}
+
+rod()
+{
+	eval "$@" || quit "$@ failed"
```
vapier, line 25: yikes, don't use `eval`.  it's not even necessary here -- just delete it.

also, use `$*`, not `$@` when calling `quit`

tools/lib.sh
```diff
@@ -0,0 +1,31 @@
...
+	esac
+}
+
+quit()
+{
+	printf "\n$@\n" >&2
+	exit 1
+}
+
+rod()
```
vapier, line 23: i have no idea what this is short for.  func (all of them really) could use a short comment header.

tools/create-tarballs.sh
```diff
@@ -0,0 +1,54 @@
...
+# Create tarballs and checksums for uploading after tagging a new release.
+
+tag="$(date +%Y%m%d)"
+tarball_dir="iputils-full-$tag"
+git_dir=$(cd $(dirname "$0")/..; pwd)
+dir="$(cd $git_dir/../; pwd)/iputils-release-$tag"
+dist=builddir
+formats="xztar,gztar,zip"
+
+. $(dirname "$0")/lib.sh
```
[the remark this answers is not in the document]
pevik: I usually check just with checkbashisms, which catches a real problems.
Some of the `shellcheck` errors does not make sense:
`SC3043 (warning): In POSIX sh, 'local' is undefined.`

or does not make sense in the context:
`^--^ SC2162 (info): read without -r will mangle backslashes.` (when `read answer` compares only against `[Yy]`

This is IMHO invalid (will not work on dash or busybox sh - these might be used on alpine):
`SC2010 (warning): Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.`

Also not sure if `prevent word splitting` is feasible to fix - with `rod` it needs to be escaped.
But sure, some warnings and errors are valid, thus fixing at least some of them (more will likely follow, just not sure if I fix all of them).

tools/create-tarballs.sh
```diff
@@ -0,0 +1,54 @@
...
+
+cd $dist/meson-dist/
+
+title "Generate tarballs and checksums"
+# meson puts '*' before filename in checksums (unusable for distros => need to
+# recreate them)
+rm -f *.sha256sum
+
+for file in *.*; do
+	sha256sum $file >> sha256sums
```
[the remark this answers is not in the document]
pevik: No, 1) list of the files is evaluated before `sha256sums` is created 2) `*.*` pattern would not select it anyway.

tools/lib.sh
pevik: > yikes, don't use eval. it's not even necessary here -- just delete it.

This is inspired by LTP, which uses `tst_rod` binary. I'm not sure if everything would work without `eval`, but maybe you're right.

tools/lib.sh
pevik: +1. Inspired by LTP, I wonder myself.

nmeyerhans: The best I could come up with when reading this was that it's short for "run-or-die".  But yeah, some more descriptive names and/or descriptive comments would be nice.

nmeyerhans (commented):
I had some comments queued up, but @vapier beat me to them.  +1 to everything he said.  Especially around:
1. Ensure variables referencing path names are quoted
2. Get rid of `eval`
3. `shellcheck` looks like it reports a few legitimate concerns; It would be good to fix them.

Looks pretty good in general, though.  I like the release notes autogeneration.

tools/create-tarballs.sh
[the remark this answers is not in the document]
vapier: while `local` is not defined in POSIX, every shell in Linux supports it, so it's not something i would bother with

tools/create-tarballs.sh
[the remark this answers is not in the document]
vapier: sorry, i thought it was `"*"`, not `*.*`

at this point, why is there a loop at all ?  just run the command directly once.  it produces the same output.

pevik:
> I had some comments queued up, but @vapier beat me to them. +1 to everything he said. Especially around:
> 
> 1. Ensure variables referencing path names are quoted
I've personally haven't heart of anybody who would clone git repository to the path which contains a space, which is IMHO the only reason why quoting is needed (or am I missing something?), but sure, I'll fix it.

> 2. Get rid of `eval`
Done.

> 3. `shellcheck` looks like it reports a few legitimate concerns; It would be good to fix them.
Sure.

> Looks pretty good in general, though. I like the release notes autogeneration.

Thanks for your feedback.

pevik:
@vapier @nmeyerhans thanks for your review. I updated the docs + slightly changed the scripts.

I dared to keep these shellcheck info/warnings:
```
In create-tarballs.sh line 44:
rm -f *.sha256sum
      ^-- SC2035 (info): Use ./*glob* or -- *glob* so names with dashes won't become options.

In create-tarballs.sh line 55:
mv -v * "$dir"
      ^-- SC2035 (info): Use ./*glob* or -- *glob* so names with dashes won't become options.

In create-tarballs.sh line 59:
cd "$dir" && ls -p | grep -v /
             ^-- SC2010 (warning): Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.

In tag-release.sh line 51:
git shortlog -s -n "$old_tag".. >> "$credit"
^-- SC2129 (style): Consider using { cmd1; cmd2; } >> file instead of individual redirects.


In lib.sh line 8:
	local msg="$1"
        ^-------^ SC3043 (warning): In POSIX sh, 'local' is undefined.


In lib.sh line 9:
	local answer
        ^----------^ SC3043 (warning): In POSIX sh, 'local' is undefined.
```

tools/lib.sh
pevik: Yep, it is "run or die". I added descriptive comments, although it's a bit overkill for 1 or 2 lines script :).

vapier: it's easy to write bad shell.  but bad shell is also bad ;).

tools/tag-release.sh
```diff
@@ -0,0 +1,79 @@
...
+old_tag="$(git describe --abbrev=0)"
+tag_msg="iputils-$tag"
+credit="/tmp/$(basename $upstream_git)-release-$tag"
+
+if ! git ls-remote --get-url origin | grep -q $upstream_git; then
+	quit "Not an upstream project"
+fi
+
+if ! git --no-pager diff --exit-code; then
+	quit "Please commit your changes before making new release"
```
vapier, line 21: this doesn't handle changes staged in the index, or unknown files

pevik: Would you add something else? Maybe
```
if ! git --no-pager diff --exit-code || ! git --no-pager diff --staged  --exit-code; then
```

Fell free to post a command. BTW I didn't meant this check to be perfect, it's just a sanity check.

Maybe not being on master branch could be a problem worth to detect (we probably don't want to release from some random branch, release is a single commit on already widely tested master branch, but merging from other branch could merge more commits to master or fail during push to master):

```
if [ "$(git branch --show-current)" != "master" ]; then
    quit "not on master branch"
fi
```
```

vapier: personally i don't have release scripts operate in the current checkout at all.  git makes it cheap to create a new shared checkout using the current repo.  so check that out to a temp location, do the generation/commit/tag steps in there, and then throw it away.

pevik: That's already in `tools/create-tarballs.sh`, because that's really safer for creating tarballs (to have just source files). But I thought the tag creation should be ok work in the current checkout. Sure, I can clone with `-b master`, but then I might need to ask user to verify 1) if the latest commit on master is the expected one to tag, 2) preferably then pull the release commit and tag in the working directory (not strictly needed for the release).

pevik:
I fixed 2 more warnings (`Use ./*glob* or -- *glob* so names with dashes won't become options.`) and added above mentioned git checks.

I also added checks for non-basic commands being installed (git, meson, ...), so that script fails at the beginning, not in the middle of the work.

@vapier @nmeyerhans I dared to add your `Reviewed-by` tags for first and second commit, because you spent time with reviewing. Feel free to have final check on them + please check for the doc change (the grammar, as I'm not a native speaker). I suppose we spent really a lot of time on this, let's get it merged.

PR #494, opened 2023-10-19
ping: Add option -H to force reverse DNS resolution

pevik:
Implements: https://github.com/iputils/iputils/issues/421
Suggested-by: Michele Guerini Rocco


pevik:
I was thinking whether `-n` and `-H` should be mutually exclusive. WDYT?

vinc17fr:
> I was thinking whether `-n` and `-H` should be mutually exclusive. WDYT?

IMHO, the last one should override the previous ones. This can be useful when one has an alias to `ping -n` or `ping -H`.

The documentation of option `-n` should be updated too, and say when this is the default. This should solve my remark on the documentation in [Debian bug 650479](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=650479) I had reported in 2011.

And concerning the `-H` documentation, "Force DNS name resolution" could be changed to "Force DNS name resolution for the output".

pevik:
> > I was thinking whether `-n` and `-H` should be mutually exclusive. WDYT?
> 
> IMHO, the last one should override the previous ones. This can be useful when one has an alias to `ping -n` or `ping -H`.

+1, I'll keep the implementation.

> The documentation of option `-n` should be updated too, and say when this is the default.

+1 I added this, with your credit in the commit message ( `Reported-by:`)

> This should solve my remark on the documentation in [Debian bug 650479](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=650479) I had reported in 2011.

I added your credit (`Reported-by:` and `Reviewed-by:`) and link to Debian bug tracker to the commit message.

> And concerning the `-H` documentation, "Force DNS name resolution" could be changed to "Force DNS name resolution for the output".

+1, done.

Could you please have a final look?

metan-ucw:
Looking at the changes, this is option to force **reverse** DNS lookup, right? So I suppose that all the documentation strings should say "reverse DNS" rather than just "DNS". Also it may make sense to map the functionality to '-R' flag instead.

gault:
> Looking at the changes, this is option to force **reverse** DNS lookup, right? So I suppose that all the documentation strings should say "reverse DNS" rather than just "DNS".

I agree that using "reverse DNS", rather than just "DNS", would probably make the documentation clearer.

> Also it may make sense to map the functionality to '-R' flag instead.

-R is already used for the record route option.

Or did you mean to do a reverse DNS lookup on the IP addresses returned by the record route option? This should be already be covered by this patch as IPOPT_RR uses pr_addr() to display its addresses (in the IPOPT_RR switch case of the pr_options() function, in ping.c).

gault (approved):

gault (approved):

pevik:
> > Looking at the changes, this is option to force **reverse** DNS lookup, right? So I suppose that all the documentation strings should say "reverse DNS" rather than just "DNS".

+1, thanks for noticing this.

> I agree that using "reverse DNS", rather than just "DNS", would probably make the documentation clearer.

I updated that (in 2 separate commits -separating man and ping changes).
 
> > Also it may make sense to map the functionality to '-R' flag instead.
> 
> -R is already used for the record route option.

@metan-ucw yes, `-R` is taken. Unfortunately there is not much letters left.

> Or did you mean to do a reverse DNS lookup on the IP addresses returned by the record route option? This should be already covered by this patch, as IPOPT_RR uses pr_addr() to display its addresses (in the IPOPT_RR switch case of the pr_options() function, in ping.c).

@metan-ucw @gault Can you please have a final look so that I can merge with your `Reviewed-by:` tag? Thank you for your time.

gault (approved):

gault (commented):
Since the rest of the patchset now says "reverse DNS", it'd make sense to adapt the commit message and the doc/ping.xml chunk too. However that doesn't look too important to me, and I'm fine with the patch as it is.

vinc17fr:
> > > I was thinking whether `-n` and `-H` should be mutually exclusive. WDYT?
> > 
> > IMHO, the last one should override the previous ones. This can be useful when one has an alias to `ping -n` or `ping -H`.
> 
> +1, I'll keep the implementation.

I meant that `-H` should override a previous `-n` and that `-n` should override a previous `-H`. It seems that this is not what commit 881cb90223e6df66f1ecae6a625f82e53822790c does. I suppose that `-n` should also set `rts.opt_force_lookup` to 0 in order to override a previous `-H`.

pevik:
> I meant that `-H` should override a previous `-n` and that `-n` should override a previous `-H`. It seems that this is not what commit [881cb90](https://github.com/iputils/iputils/commit/881cb90223e6df66f1ecae6a625f82e53822790c) does. I suppose that `-n` should also set `rts.opt_force_lookup` to 0 in order to override a previous `-H`.

@vinc17fr Fixed this before merge.

@vinc17fr @gault @metan-ucw Thanks a lot for your reviews, suggestions!

PR #498, opened 2023-11-15
tracepath: Fix possible nullptr dereference.

chestnykh:
`ctl.ai` 'potentially' can be null when we reach `switch()` statement because of loop above,
so we should check that `ctl.ai` is not NULL.

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -11,6 +11,7 @@
 
 #define _GNU_SOURCE
 
+#include <assert.h>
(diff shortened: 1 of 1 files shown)


metan-ucw:
@pevik technically speaking if the result contains only responses that match the if conditions that cause the loop to continue we can end up with `ctl.ai` being` NULL` it's probably very unlikely or close to impossible to happen but I guess that `abort()` wouldn't hurt.

PR #499, opened 2023-11-21
ping: Handle interval correctly in the second after booting

joshtriplett:
ping assumes that if a timespec has `tv_sec == 0`, it hasn't been initialized
yet. However, in the second after booting up, `tv_sec` will legitimately be 0.
This causes ping to send pings one after another without waiting.

Check that `tv_nsec` is 0 as well, which fixes the interval handling to wait
between pings.

Fix a similar issue in tracepath, as well.

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -319,7 +319,7 @@ int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock)
 		return 1000;
 
 	/* Check that packets < rate*time + preload */
-	if (rts->cur_time.tv_sec == 0) {
+	if (rts->cur_time.tv_sec == 0 && rts->cur_time.tv_nsec == 0) {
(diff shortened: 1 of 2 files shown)


pevik (approved):
```
```

pevik:
Out of curiosity, do you use some dedicated single specific distribution, or did you run with `init=/usr/bin/ping`? Or you were really able to reproduce the issue with real VM? I'm just curious who needs ping in less than a first second after boot :).

joshtriplett:
@pevik I was running user code in isolated VMs, which boot, initialize in milliseconds, and immediately run a user workload. I was debugging networking issues with those VMs, and in the course of that debugging I handed them a "user workload" of `ping -4 -n -c 10 8.8.8.8`, and found that the 10 pings all happened immediately rather than over the course of 10 seconds.

PR #502, opened 2023-12-04
ping: More precise check for our packet

pevik:
Ping run with Nagios via Debian specific check-host-alive Nagios plugin (e.g. ping -n -v -D -W 1 -i 1 -c 5 -M 'do' -s 56 -O "$Host") can run many ping in parallel (e.g. 75-100 in the reported issue).

This leads to the collisions, because ICMP identification field is only 16 bits (65535). The problem is not new, but it was hidden until 5f6bec5 ("ping: Print reply with wrong source with warning")

We consider this warning useful, therefore instead of reverting another check for sequence was added.

Other possible solutions to this:

* Revert 5026c22 ("ping: use random value for the identifier field") Ping in the past used PID as ICMP identification field. This was considered insecure, thus it was changed to use pseudo random number in 5026c22 ("ping: use random value for the identifier field") although there is no CVE and the same approach is used in fping and busybox's ping, thus it's questionable if it was really needed.

* Add another check in the ICMP optional data We could add 128 bit random value to check. But we already use struct timeval if packet size is big enough for it (>= 16 bits), therefore we could use it for comparing for most of the packet sizes (the default is 56 bits).

Fixes: https://github.com/iputils/iputils/issues/489
Reported-by: Miloslav Hůla <miloslav.hula@gmail.com>

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1014,6 +1014,7 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 
 	if (rts->datalen >= (int)sizeof(struct timeval))	/* can we time transfer */
 		rts->timing = 1;
+
 	packlen = rts->datalen + MAXIPLEN + MAXICMPLEN;
 	if (!(packet = (unsigned char *)malloc((unsigned int)packlen)))
 		error(2, errno, _("memory allocation failed"));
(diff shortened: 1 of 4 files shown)


metan-ucw:
I do not think that the sequence number is a good identifier of a packet. I suppose that if we set -i small enough, we may start getting packets out of order, for me running 'ping -i 0.01 1.1.1.1' produces quite a bit out-of-order replies.

I would propose closing this since we decided to revert the id randomization. @pevik what do you think?

pevik:
Yes, you're right, this is not the solution. I also saw lost packets in the statics, when I ping `1.1.1.1` or `8.8.8.8` (i.e. IP over net) with higher `-i`, I before tested only pinging localhost or VM on the laptop, which is ok.

We really need to revert back to PID (revert 5026c2221a15bf13e601eade015c971bf07a27e9) as you suggested on https://github.com/iputils/iputils/issues/489. I'll prepare PR.

PR #507, opened 2023-12-13
man/ping: Document collisions and pid_max

pevik:
This is a follow up of d466aab ("Revert "ping: use random value for the identifier field"").

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -851,6 +851,29 @@ xml:id="man.ping">
     hosts).</para>
   </refsection>
 
+  <refsection xml:id="id_collisions">
+    <info>
+      <title>ID COLLISIONS</title>
+    </info>
+    <para>
+    Unlike TCP and UDP, which use port to uniquely identify the recipient to
+    deliver data, ICMP uses identifier field (ID) for identification.
+    Therefore, if on the same machine, at the same time, two ping processes
+    use the same ID, echo reply can be delivered to a wrong recipient.
+    This is a known problem due to the limited size of the 16-bit ID field.
+    That is a historical limitation of the protocol that cannot be fixed
+    at the moment unless we encode an ID into the ping packet payload.
+    <command>ping</command> prints <emphasis remap="I">DIFFERENT ADDRESS</emphasis>
+    error and packet loss is negative.
+    </para>
+    <para>
+    <command>ping</command> uses PID to get unique number.  The default value of
+    <emphasis remap="I">/proc/sys/kernel/pid_max</emphasis> is 32768.
+    On the systems that use ping heavily and with <emphasis remap="I">pid_max</emphasis>
+    greater than 65535 collisions are bound to happen.
+    </para>
+  </refsection>
+
   <refsection xml:id="trying_different_data_patterns">
     <info>
       <title>TRYING DIFFERENT DATA PATTERNS</title>


doc/ping.xml
```diff
@@ -851,6 +851,28 @@ xml:id="man.ping">
     hosts).</para>
   </refsection>
 
+  <refsection xml:id="id_collisions">
+    <info>
+      <title>ID COLLISIONS</title>
+    </info>
+    <para>
+    Unlike TCP and UDP, which use port to uniquely identify the socket to
+    deliver data, ICMP use identifier field (ID) to identify the socket.
```
metan-ucw, line 860: I'm not sure it's supposed to be called socket for ICMP. Maybe recipient?

doc/ping.xml
```diff
@@ -851,6 +851,28 @@ xml:id="man.ping">
...
+    Therefore if on the same machine, at the same time, two ping processes
+    use the same ID, echo reply can be delivered to the wrong socket.
```
metan-ucw, line 862: Here as well 'wrong recipient'?

doc/ping.xml
```diff
@@ -851,6 +851,28 @@ xml:id="man.ping">
...
+    This is known problem due 16 bit ID field (65535).
```
metan-ucw, line 863: 'due to limited size of the 16ibt ID field'

doc/ping.xml
```diff
@@ -851,6 +851,28 @@ xml:id="man.ping">
...
+    <command>ping</command> prints <emphasis remap="I">DIFFERENT ADDRESS</emphasis>
+    error and packet loss is negative.
+    </para>
+    <para>
+    <command>ping</command> uses PID to get unique number.  The default value of
+    <emphasis remap="I">/proc/sys/kernel/pid_max</emphasis> is 32768.
+    On the systems that use ping heavily it's important to have
+    <emphasis remap="I">/proc/sys/kernel/pid_max </emphasis>
+    value less or equal to 65535.
```
metan-ucw, line 872: Maybe something along the lines 'on the systems that use ping heavily and with pid_max greater than 65535 collisions are bound to happen' and maybe add that this is a historical limitation of the protocol that we cannot really fix at this point. Also if we have an option to encode an ID into the ping packet payload we should probably mention it here as well.

doc/ping.xml
```diff
@@ -851,6 +851,28 @@ xml:id="man.ping">
     hosts).</para>
   </refsection>
 
+  <refsection xml:id="id_collisions">
+    <info>
+      <title>ID COLLISIONS</title>
+    </info>
+    <para>
+    Unlike TCP and UDP, which use port to uniquely identify the recipient to
+    deliver data, ICMP use identifier field (ID) for identification.
```
Avinesh, line 860: nit:
s/ICMP use/ICMP uses the/

Avinesh (approved):

doc/ping.xml
```diff
@@ -851,6 +851,29 @@ xml:id="man.ping">
...
   </refsection>
 
+  <refsection xml:id="id_collisions">
+    <info>
+      <title>ID COLLISIONS</title>
+    </info>
+    <para>
+    Unlike TCP and UDP, which use port to uniquely identify the recipient to
+    deliver data, ICMP uses identifier field (ID) for identification.
+    Therefore if on the same machine, at the same time, two ping processes
```
Avinesh, line 861: s/Therefore if/Therefore, if/

doc/ping.xml
```diff
@@ -851,6 +851,29 @@ xml:id="man.ping">
...
+    use the same ID, echo reply can be delivered to a wrong recipient.
+    This is known problem due to limited size of the 16 bit ID field.
```
Avinesh, line 863: again minor nits:

This is a known problem due to the limited size of the 16-bit ID field

Avinesh (approved):
with those grammar fixes :)

pevik:
Thanks a lot both!

Issue #504, opened 2023-12-11
Add an errno return


xanoxes:
In some special networking environments, like the beginning of bond4, sendto may fail. In that way, for example in arping, it will print Sent 0 probes output, but in most scenarios, we just judge the return value. In this cases it just returns -1, we wonder if it is necessary to add some more detailed error reportings in output ?

xanoxes:
--> like the beginning of bond4, sendto may fail
  --> here is a issue about a usage of arping in this scene
    https://github.com/fedora-sysv/initscripts/issues/465

PR #508, opened 2023-12-14
add ret check of send_pack in arping

xanoxes:
In some special networking environments, like the beginning of bond4, sendto may fail. In that way, for example in arping, it will print Sent 0 probes output, but in most scenarios, we just judge the return value. In this cases it just returns -1, we wonder if it is necessary to add some more detailed error reportings in output ?

here is a issue about a usage of arping in this scene
https://github.com/fedora-sysv/initscripts/issues/465

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -693,7 +693,7 @@ static void find_broadcast_address(struct run_state *ctl)
 
 static int event_loop(struct run_state *ctl)
 {
-	int exit_loop = 0, rc = 0;
+	int exit_loop = 0, rc = 0, err = 0;
 	ssize_t s;
 	enum {
 		POLLFD_SIGNAL = 0,
@@ -775,7 +775,7 @@ static int event_loop(struct run_state *ctl)
 	/* socket */
 	pfds[POLLFD_SOCKET].fd = ctl->socketfd;
 	pfds[POLLFD_SOCKET].events = POLLIN | POLLERR | POLLHUP;
-	send_pack(ctl);
+	err = send_pack(ctl);
 
 	while (!exit_loop) {
 		int ret;
@@ -818,7 +818,7 @@ static int event_loop(struct run_state *ctl)
 					exit_loop = 1;
 					continue;
 				}
-				send_pack(ctl);
+				err = send_pack(ctl);
 				break;
 			case POLLFD_TIMEOUT:
 				exit_loop = 1;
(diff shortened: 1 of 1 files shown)


xanoxes:
Associate this issue:
https://github.com/iputils/iputils/issues/504

arping.c
```diff
@@ -844,6 +844,8 @@ static int event_loop(struct run_state *ctl)
 	close(sfd);
 	close(tfd);
 	freeifaddrs(ctl->ifa0);
+	if (err < 0)
+                return err;
```
pevik, line 848: Unlike ping, which exits on error with `1` or `2` (depending whether exit code is related by ping error or other error), arping specifies just `1`. IMHO the exit code should be also `2`.

`-1` will be translated to `255` due `status & 0xFF` (see [`man exit(3)`](https://man7.org/linux/man-pages/man3/exit.3.html).

Also, we would need to print error message, use [`strerror(3)`](https://man7.org/linux/man-pages/man3/strerror.3.html).

pevik:
Although this is useful, it's unfortunately not ready for upcoming release.

Are you able to reproduce the issue without using bond4? That would help testing the functionality.

Also, you should add https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin (it was in the description you deleted, it will not be merged under your name without it).

arping.c
xanoxes: The 'err' here is the return value of sendto(). According to https://man7.org/linux/man-pages/man3/sendto.3p.html, the return value of sendto() returned from send_pack() when failed should be -1, shall we consider other scenarios of failure?I add a print of error message using strerror, should we just do it in send_pack() or out?

xanoxes:
> Although this is useful, it's unfortunately not ready for upcoming release.
> 
> Are you able to reproduce the issue without using bond4? That would help testing the functionality.
> 

Of course I will add the way to reproduce this issue using bond4.

> Also, you should add https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin (it was in the description you deleted, it will not be merged under your name without it).

How should i add this, directly added to commitmsg?

pevik:
> > Although this is useful, it's unfortunately not ready for upcoming release.
> > Are you able to reproduce the issue without using bond4? That would help testing the functionality.
> 
> Of course I will add the way to reproduce this issue using bond4.

Could you just post commands which you are using for testing?

> 
> > Also, you should add https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin (it was in the description you deleted, it will not be merged under your name without it).
> 
> How should i add this, directly added to commitmsg?

Yes.

Issue #515, opened 2024-01-11
Capability issue with ping -m


JJ-Meng:
Currently, `ping` is configured with `CAP_NET_RAW`, which generally suffices for most operations. However, we encountered a permission issue when using the `-m` option to mark outgoing packets.

I’ve tested it on the newest release(20231222).

```shell
$ /usr/sbin/getcap ping
ping = cap_net_raw+p

$ ./ping www.debian.org -m1
./ping: WARNING: failed to set mark: 1: Operation not permitted
PING www.debian.org (130.89.148.77) 56(84) bytes of data.
./ping: WARNING: failed to set mark: 1: Operation not permitted
64 bytes from klecker-misc.debian.org (130.89.148.77): icmp_seq=1 ttl=48 time=288 ms
64 bytes from klecker-misc.debian.org (130.89.148.77): icmp_seq=2 ttl=48 time=288 ms
``` 

Accroding to the capability [man-page](https://www.man7.org/linux/man-pages/man7/capabilities.7.html),  setting `SO_MARK` needs `CAP_NET_ADMIN`. So I would like to ask which is better, using root to execute the `ping` or adding `CAP_NET_ADMIN` to the `ping`?

pevik:
> Accroding to the capability [man-page](https://www.man7.org/linux/man-pages/man7/capabilities.7.html), setting `SO_MARK` needs `CAP_NET_ADMIN`. So I would like to ask which is better, using root to execute the `ping` or adding `CAP_NET_ADMIN` to the `ping`?

Safer is to add `CAP_NET_ADMIN` (only a specific privilege, than using root, which is full privilege).

JJ-Meng:
Thanks for the explanation. I also found it kind of weird to use root for a capability-aware program like `ping`. So would you consider adding `CAP_NET_ADMIN` to it? It'd make things more consistent and user-friendly.

pevik:
It looks to be a regression when the feature was added in ce7be0e0f91d70b5af8fb6fc3c108c7cd880d7eb (released in s20100214).

pevik:
Hm, but we already do set `CAP_NET_ADMIN` via `enable_capability_admin()`:
https://github.com/iputils/iputils/blob/c04828ab8cc671d2a383f544f49a4d721bdec8fc/ping/ping_common.c#L458-L471

[...]

PR #517, opened 2024-01-12
Add `cap_net_admin`, refactor socket mark failure handling

pevik:
ping: Refactor socket mark failure handling

Ping warned 2x about setsockopt SO_MARK failure, because since 08f0e17 is called also for probe socket (before it was only on real socket):

	$ ping -m 2 ::1
	ping: WARNING: failed to set mark: 2: Operation not permitted
	PING ::1 (::1) 56 data bytes
	ping: WARNING: failed to set mark: 2: Operation not permitted
	64 bytes from ::1: icmp_seq=1 ttl=64 time=0.039 ms

Therefore pass rts struct to:

1) Disable SO_MARK call on real socket when it failed on probe socket,
therefore only single warning is printed.
2) Print hint about possible missing cap_net_admin+p capability

```
$ ./builddir/ping/ping -m 2 ::1
./builddir/ping/ping: WARNING: failed to set mark: 2: Operation not permitted
./builddir/ping/ping: => missing cap_net_admin+p capability?
PING ::1 (::1) 56 data bytes
64 bytes from ::1: icmp_seq=1 ttl=64 time=0.053 ms
```

Fixes: 08f0e17 ("ping: Set mark (SO_MARK) on probe socket")
Fixes: https://github.com/iputils/iputils/issues/515

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -357,7 +357,12 @@ xml:id="man.ping">
           <emphasis remap="I">mark</emphasis> to tag the packets
           going out. This is useful for variety of reasons within
           the kernel such as using policy routing to select
-          specific outbound processing.</para>
+          specific outbound processing. CAP_NET_ADMIN or CAP_NET_RAW
+          (since Linux 5.17) capability is required, see
+          <citerefentry>
+            <refentrytitle>socket</refentrytitle>
+            <manvolnum>7</manvolnum>
+          </citerefentry>.</para>
         </listitem>
       </varlistentry>
       <varlistentry>
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -455,20 +455,28 @@ void sock_setbufs(struct ping_rts *rts, socket_st *sock, int alloc)
 	}
 }
 
-void sock_setmark(unsigned int mark, int fd)
+void sock_setmark(struct ping_rts *rts, int fd)
 {
 #ifdef SO_MARK
 	int ret;
 	int errno_save;
 
+	if (!rts->opt_mark)
+		return;
+
 	enable_capability_admin();
-	ret = setsockopt(fd, SOL_SOCKET, SO_MARK, &mark, sizeof(mark));
+	ret = setsockopt(fd, SOL_SOCKET, SO_MARK, &(rts->mark), sizeof(rts->mark));
 	errno_save = errno;
 	disable_capability_admin();
 
-	/* Do not exit, old kernels do not support mark. */
-	if (ret == -1)
-		error(0, errno_save, _("WARNING: failed to set mark: %u"), mark);
+	if (ret == -1) {
+		error(0, errno_save, _("WARNING: failed to set mark: %u"), rts->mark);
+
+		if (errno_save == EPERM)
+			error(0, 0, _("=> missing cap_net_admin+p or cap_net_raw+p (since Linux 5.17) capability?"));
+
+		rts->opt_mark = 0;
+	}
 #else
 		error(0, errno_save, _("WARNING: SO_MARK not supported"));
 #endif
@@ -506,8 +514,7 @@ void setup(struct ping_rts *rts, socket_st *sock)
 	}
 #endif
 
-	if (rts->opt_mark)
-		sock_setmark(rts->mark, sock->fd);
+	sock_setmark(rts, sock->fd);
 
 	/* Set some SNDTIMEO to prevent blocking forever
 	 * on sends, when device is too slow or stalls. Just put limit
--- a/build-aux/setcap-setuid.sh
+++ b/build-aux/setcap-setuid.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright (c) Iputils Project, 2018-2021
+# Copyright (c) Iputils Project, 2018-2024
 # Meson install script to setcap or setuid to an executable.
 
 exec_path="$1/$2"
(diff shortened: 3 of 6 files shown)


pevik:
@PaulSD Could you please have look on this? Touches `SO_MARK`.

ping/ping_common.c
```diff
@@ -455,20 +455,25 @@ void sock_setbufs(struct ping_rts *rts, socket_st *sock, int alloc)
...
+	ret = setsockopt(fd, SOL_SOCKET, SO_MARK, &(rts->mark), sizeof(rts->mark));
 	errno_save = errno;
 	disable_capability_admin();
 
-	/* Do not exit, old kernels do not support mark. */
-	if (ret == -1)
-		error(0, errno_save, _("WARNING: failed to set mark: %u"), mark);
+	if (ret == -1) {
+		error(0, errno_save, _("WARNING: failed to set mark: %u"), rts->mark);
+		error(0, 0, _("=> missing cap_net_admin+p capability?"));
```
mgerstner, line 474: You could limit this hint to cases when errno actually is `EPERM` (or `EACCES`?)

doc/ping.xml
```diff
@@ -357,7 +357,8 @@ xml:id="man.ping">
           <emphasis remap="I">mark</emphasis> to tag the packets
           going out. This is useful for variety of reasons within
           the kernel such as using policy routing to select
-          specific outbound processing.</para>
+          specific outbound processing.
+          CAP_NET_ADMIN capability is required.</para>
```
mgerstner, line 361: You could also refer to `man 7 socket`:

```
        SO_MARK (since Linux 2.6.25)
               Set  the  mark  for  each  packet sent through this socket (similar to the netfilter MARK target but socket-based).  Changing the mark can be used for mark-based routing without netfilter or for packet filtering.
               Setting this option requires the CAP_NET_ADMIN capability.
```
as an authoritative source for this, here, or in another spot where this is suitable.

pevik:
> The change looks clean to me, although it is unfortunate that this capability needs to be added for a rather special purpose feature.

Yes, I'm not happy with it. I guess that was the reason, why it haven't been added so far. I also wonder if it's good distros would add it.

ping/ping_common.c
pevik: Makes sense actually, thanks! I also add `(or run ping under root)`, although it might not be a good idea due security to run ping under root (but it will be needed if user needs it and distro does not add it or he does not modify the capability himself).

pevik: Current kernel set `EPERM`, therefore I'll use just that. 
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/core/sock.c?h=v6.7#n1415

Also, since v5.17 the check added also `CAP_NET_RAW` (so now it's  `CAP_NET_RAW`  or `CAP_NET_ADMIN`: [079925cce1d0 ("net: allow SO_MARK with CAP_NET_RAW")
](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=079925cce1d06fe6614b1559b0020497b5b83d6a). Therefore I sent a patch to linux man pages https://lore.kernel.org/linux-man/20240116110418.2577798-1-pvorel@suse.cz/ and note it in ping man page as well.

doc/ping.xml
```diff
@@ -357,7 +357,12 @@ xml:id="man.ping">
           <emphasis remap="I">mark</emphasis> to tag the packets
           going out. This is useful for variety of reasons within
           the kernel such as using policy routing to select
-          specific outbound processing.</para>
+          specific outbound processing.
+          CAP_NET_ADMIN capability is required, see
```
PaulSD, line 361: This should probably also mention CAP_NET_RAW since 5.17?

pevik: Originally I did not want, but because I updated man page (https://www.alejandro-colomar.es/src/alx/linux/man-pages/man-pages.git/commit/?id=527f31b9f762d0b61aa6aa5da8b8f4e63520b1fd) I'll update it.

ping/ping_common.c
```diff
@@ -455,20 +455,26 @@ void sock_setbufs(struct ping_rts *rts, socket_st *sock, int alloc)
...
 	errno_save = errno;
 	disable_capability_admin();
 
-	/* Do not exit, old kernels do not support mark. */
-	if (ret == -1)
-		error(0, errno_save, _("WARNING: failed to set mark: %u"), mark);
+	if (ret == -1) {
+		error(0, errno_save, _("WARNING: failed to set mark: %u"), rts->mark);
+		if (errno_save == EPERM)
+			error(0, 0, _("=> missing cap_net_admin+p capability?"));
```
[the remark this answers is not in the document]
pevik: I guess I should check kernel version and print `cap_net_raw+p` only when relevant. I also wonder if print capability constants (`CAP_NET_RAW`, `CAP_NET_ADMIN`) instead of capability `setcap` capability (`cap_net_raw+p`, `cap_net_admin+p`).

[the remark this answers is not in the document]
pevik: I could take LTP functions `tst_parse_kver()` and `parse_digit()` which parse `char release[]` member of `struct utsname`, to detect if kernel is new enough for `cap_net_raw+p` to be relevant. But maybe it's not worth to do that. So far simple `missing cap_net_admin+p or cap_net_raw+p (since Linux 5.17) capability?` without checking kernel version added.

https://github.com/linux-test-project/ltp/blob/db0082b45cc6f8f7c8c1c5f1cffb3ed5522d812e/lib/tst_kvercmp.c#L31-L82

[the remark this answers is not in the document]
PaulSD: > I also wonder if print capability constants (CAP_NET_RAW, CAP_NET_ADMIN) instead of capability setcap capability (cap_net_raw+p, cap_net_admin+p).

I actually think the setcap is probably better here, since this is going to be seen by end users, and setcap is how the user would fix this.

I agree it probably isn't worth the effort to do version detection.  This LGTM.

PR #519, opened 2024-01-16
ping: Restore `-i0`

pevik:
`-i 0` has been supported for a long time, restore it back. There is no point to force user to set very minimal value to emulate 0.

Found by LTP tests (nft01.sh, iptables01.sh) which use it.

Fixes: 2a63b94 ("ping: Fix overflow on negative -i")

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -441,7 +441,7 @@ main(int argc, char **argv)
 			double optval;
 
 			optval = ping_strtod(optarg, _("bad timing interval"));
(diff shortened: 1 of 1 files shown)


pevik:
@iputils/maintainers Could you please have look? I probably release this week to get the regression fixed quickly.

okias:
LGTM,

```
Reviewed-by: David Heidelberg <david@ixit.cz>
```

PR #520, opened 2024-01-25
meson: include pre-generated man pages & HTML docs in dist tarballs

thesamesam:
Bug: https://github.com/iputils/iputils/issues/479
Bug: https://bugs.gentoo.org/908817
Bug: https://bugs.gentoo.org/920901

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -220,6 +220,10 @@ endif
 
 if build_mans == true or build_html_mans == true
 	subdir ('doc')
+
+	if meson.version().version_compare('>=0.48.0')
+		meson.add_dist_script('tools/meson-build-dist-man.sh')
+	endif
 endif
 
 if run_tests
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -20,6 +20,30 @@ if build_tracepath == true
 	manpages += ['tracepath']
 endif
 
+if meson.version().version_compare('>=0.58.0')
+	# Use the docs included in a dist tarball if available
+	fs = import('fs')
+	out_man_pages = []
+	out_html_pages = []
+	generated_docs_exist = true
+	foreach page : manpages
+		man_page_name = page.replace('.xml', '.8')
+		html_page_name = page.replace('.xml', '.html')
+		out_man_pages += man_page_name
+		out_html_pages += html_page_name
+		if not fs.exists(man_page_name) or not fs.exists(html_page_name)
+			generated_docs_exist = false
+		endif
+	endforeach
+	if generated_docs_exist
+		install_man(out_man_pages)
+	        install_data(out_html_pages, install_dir : join_paths(get_option('datadir'), meson.project_name()))
+		subdir_done()
+	endif
+else
+	warning('Will not attempt to use pre-generated man pages with <0.58.0')
+endif
+
 xsltproc = find_program('xsltproc', required : build_mans or build_html_mans)
 xsltproc_args = [
 	'--nonet',
--- a/tools/meson-build-dist-man.sh
+++ b/tools/meson-build-dist-man.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+# This script should be invoked by meson itself (via 'meson dist')
+# See https://github.com/mesonbuild/meson/issues/2166 and more specifically,
+# https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911.
+set -eu
+
+cd "$MESON_DIST_ROOT"
+DIR=$(mktemp -d)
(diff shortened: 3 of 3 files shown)


thesamesam:
This is a followup to https://github.com/iputils/iputils/issues/479. I'm sorry I didn't end up reviewing the work there, it's been a very hectic couple of months :(

meson.build
```diff
@@ -218,8 +218,11 @@ if build_arping == true
 	endif
 endif
 
-if build_mans == true or build_html_mans == true
-	subdir ('doc')
+if meson.version().version_compare('>=0.58.0')
+	if build_mans == true or build_html_mans == true
+		subdir ('doc')
+		meson.add_dist_script('meson-build-dist-man.sh')
+	endif
```
eli-schwartz, line 225: Typo for 0.48 (when dist scripts were added)?

Anyway I think you don't want to skip the entire doc/ directory on too old versions of meson.

eli-schwartz (changes requested):

meson.build
thesamesam: Yeah, I think I was copying the 0.58.0 from `doc`. Good catch on the skipping...

pevik:
@nmeyerhans any opinion about this?

I would really prefer people either download build man page dependencies or disable building it. But understand the use case.
The implementation needs minor cleanup (script in `tools/` etc.), but that's a minor detail (I can fix these tiny issues if we agree we want this approach).

thesamesam:
FWIW, I wouldn't ask if it weren't a core system package, and I'm also happy to maintain and handle any problems with this side of things if they crop up. But I also understand if you want to leave it too. Thank you for considering it either way!

meson-build-dist-man.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+# This script should be invoked by meson itself (via 'meson dist')
+# See https://github.com/mesonbuild/meson/issues/2166 and more specifically,
+# https://github.com/mesonbuild/meson/issues/2166#issuecomment-629696911.
+set -eu
+
+cd "${MESON_DIST_ROOT}"
```
pevik, line 7: nit: could you please change to   `cd "$MESON_DIST_ROOT"`

meson-build-dist-man.sh
```diff
@@ -0,0 +1,12 @@
...
+mkdir build
```
pevik, line 8: I wonder if we should use `mktemp`, that would avoid to clash (name is too generic).

meson-build-dist-man.sh
pevik: Also, please move the script to the `tools/` directory.

pevik:
@thesamesam I understand all distros would benefit from it (less dependencies shorten build + require less resources), so I'm for this (I'm sorry I haven't managed to handle it before the release). Could you please rebase your pull request and handle minor issues I reported?

pevik:
@thesamesam FYI I have ready slightly modified branch https://github.com/pevik/iputils/tree/meson-dist.fixes, I just wait if we get any input from https://github.com/mesonbuild/meson/issues/2166#issuecomment-1938765316 to get this running on cross compilation, or if I need to disable it with help of [`meson.is_cross_build()`](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonis_cross_build) as you suggested.

pevik:
I wonder why `FAILED unittests/linuxliketests.py::LinuxlikeTests::test_generate_gir_with_address_sanitizer` failed - it timed out due:

```
stdout_seq = [b"ninja explain: deps for 'resources/simple-resources-test.p/meson-generated_.._simple-resources.c.o' are missing\n",...ces.c.o is dirty\n', b"ninja explain: deps for 'resources/simple-resources-test.p/simple-main.c.o' are missing\n", ...]

```

thesamesam:
Thank you!

pabs3:
FTR, other distros like Debian strongly prefer to not have generated files in the VCS and tarballs. Instead everything should always be built from source. More info on the Debian [AutoGeneratedFiles](https://wiki.debian.org/AutoGeneratedFiles) wiki page.

thesamesam:
You're free to not use them in that case.

nmeyerhans:
> FTR, other distros like Debian strongly prefer to not have generated files in the VCS and tarballs. Instead everything should always be built from source. More info on the Debian [AutoGeneratedFiles](https://wiki.debian.org/AutoGeneratedFiles) wiki page.

Debian doesn't use the upstream release tarballs at all, but builds from the git tags directly. So none of this is relevant here.

PR #532, opened 2024-04-02
feat(ping): ping support bind to device and addr at the same time

pushan01:
Currently, ping allows send packet from a specific address or a specificed interface but not both. With this PR,  one can use -I directive to  send ICMP packet through specific address and IP Address as below:

root@yatta:~/iputils/builddir/ping# ping 100.96.5.6 -I ipsecd9ykxy -I 100.96.5.5 -c 4
PING 100.96.5.6 (100.96.5.6) from 100.96.5.5 ipsecd9ykxy: 56(84) bytes of data.
64 bytes from 100.96.5.6: icmp_seq=1 ttl=64 time=0.017 ms
64 bytes from 100.96.5.6: icmp_seq=2 ttl=64 time=0.020 ms
64 bytes from 100.96.5.6: icmp_seq=3 ttl=64 time=0.027 ms
64 bytes from 100.96.5.6: icmp_seq=4 ttl=64 time=0.018 ms

--- 100.96.5.6 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 48ms
rtt min/avg/max/mdev = 0.017/0.020/0.027/0.006 ms

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -811,6 +811,9 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 		argv++;
 	}
 
+	if (rts->device)
+		bind_to_device(rts, sock->fd, dst.sin_addr.s_addr);
+
 	if (rts->source.sin_addr.s_addr == 0) {
 		socklen_t alen;
 		int probe_fd = socket(AF_INET, SOCK_DGRAM, 0);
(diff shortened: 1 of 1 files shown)


pevik:
Could you describe your use case, why you need both? Ideally with proof of concept using network namespaces (so that this can be retested).

Change like this would need to be carefully tested (there were some regressions on IPv6, also since  9e08707d743b29e853df81bd7def1729e3afe55d ping supports VRF).

ping/ping.c
```diff
@@ -811,6 +811,9 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 		argv++;
 	}
 
+	if (rts->device) {
+		bind_to_device(rts, sock->fd, dst.sin_addr.s_addr);
+	}
```
pevik, line 816: nit: `{ }` are not needed, add space below (readability).

PR #533, opened 2024-04-04
Fix print time_t problem, when it runs on 32 bits platform.

Chozmao:
- We have an image compiled with musl libc, which have configured time_t as 64-bit.
_musl 1.2.0 changes the definition of time_t, and thereby the definitions of all derived types, to be 64-bit across all archs._

- We have this image run on a 32-bit platform, then we encountered a problem. The consumed time for ping is always shown as 0ms in the output.
<pre>
ping 172.16.0.9
PING 172.16.0.9 (172.16.0.9) 56(84) bytes of data.
64 bytes from 172.16.0.9: icmp_seq=1 ttl=64 time=0.745 ms
64 bytes from 172.16.0.9: icmp_seq=2 ttl=64 time=0.914 ms
64 bytes from 172.16.0.9: icmp_seq=3 ttl=64 time=0.930 ms
--- 172.16.0.9 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.745/0.863/0.930/0.083 ms
</pre>
- I think a casting is necessary here.  From musl time64 Release Notes
<pre>
Undefined behavior and unwarranted assumptions
    Use of %ld format to print time_t or suseconds_t.
</pre>

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -903,7 +903,7 @@ int finish(struct ping_rts *rts)
 #endif
 		printf(_(", %g%% packet loss"),
 		       (float)((((long long)(rts->ntransmitted - rts->nreceived)) * 100.0) / rts->ntransmitted));
-		printf(_(", time %ldms"), 1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000);
+		printf(_(", time %llums"), (unsigned long long)(1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000));
 	}
 
 	putchar('\n');


ping/ping_common.c
```diff
@@ -903,7 +903,12 @@ int finish(struct ping_rts *rts)
 #endif
 		printf(_(", %g%% packet loss"),
 		       (float)((((long long)(rts->ntransmitted - rts->nreceived)) * 100.0) / rts->ntransmitted));
-		printf(_(", time %ldms"), 1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000);
+
+        if (sizeof(time_t) == 4) { // 32 bits time_t
+		    printf(_(", time %lums"), (unsigned long)(1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000));
+        } else { // 64 bits time_t
+		    printf(_(", time %llums"), (unsigned long long)(1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000));
+        }
```
metan-ucw, line 911: I do not think that we need the condition. I think that we can just cast everything into unsigned long long and be done with it, i.e. we only need the:

```
printf(_(", time %llums"), (unsigned long long)(1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000));
```

pevik: @Chozmao could you please test it? I should also add some 32bit build into CI.

Chozmao: @pevik @metan-ucw , you are correct. I have tested the casting to unsigned long long on 32bit device with 32bit time_t. It also works as expected. I just updated my pull request. Thanks

ping/ping_common.c
```diff
@@ -903,7 +903,8 @@ int finish(struct ping_rts *rts)
 #endif
 		printf(_(", %g%% packet loss"),
 		       (float)((((long long)(rts->ntransmitted - rts->nreceived)) * 100.0) / rts->ntransmitted));
-		printf(_(", time %ldms"), 1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000);
+
+        printf(_(", time %llums"), (unsigned long long)(1000 * tv.tv_sec + (tv.tv_nsec + 500000) / 1000000));
```
pevik, line 907: Also, could you please keep the tab where it was?

pevik:
@Chozmao Could you please

1) Squash your changes into single commit (`git rebase -i master` and in the editor add `s` to the second commit)
2) Add `Signed-off-by: Your Name <your@email>` into the commit message (e.g. `Hangmao Liu <Hangmao.Liu@garderos.com>`), you do it via `git commit --amend`.

Or, if you don't do it, I can do it myself before merge.

With these changes:
```
```
@metan-ucw Can I add your `Reviewed-by:`?

PR #534, opened 2024-04-04
ping: fix IPv4 checksum check always succeeding once again

a3f:
This issue was fixed once already in commit bff65fbb6f73 ("fix checksum always success in IPv4 ping."), but was reverted shortly after, likely due to a botched rebase.

Fix this issue again, so ping correctly reports checksum mismatches in ICMP ECHO replies.

Fixes: 8b8686794f69 ("warnings: remove variable shadowing")

---

To reproduce: Boot [barebox](https://www.barebox.org/doc/latest/boards/bcm2835.html#raspberry-pi) v2024.03.0 or older on a Raspberry Pi 3b and ping its network interface. Without [recent fixes](https://lore.barebox.org/barebox/20240404184001.1532897-1-a.fatoum@pengutronix.de/T/#mc182a3fe4eb2053a9af13e5a5e15c830a5ef468d), this results in checksum mismatch as reported by Wireshark. With this patch applied, `ping(1)` will also report this checksum mismatch:

```
64 bytes from 192.168.10.82: icmp_seq=1 ident=1066 ttl=64 time=9.69 ms (BAD CHECKSUM!)
```

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1654,7 +1654,7 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 			wrong_source = 1;
 		if (gather_statistics(rts, (uint8_t *)icp, sizeof(*icp), cc,
 				      ntohs(icp->un.echo.sequence),
-				      reply_ttl, 0, tv, pr_addr(rts, from, sizeof *from),
+				      reply_ttl, csfailed, tv, pr_addr(rts, from, sizeof *from),
 				      pr_echo_reply, rts->multicast, wrong_source)) {
 			fflush(stdout);
 			return 0;


pevik:
```
```
It looks correct, but since there is interesting reproducer, I'd like to test it. Hope to get time to do it soon.

pevik:
I'm sorry for the noise, I see, you liked the Barebox fix: https://lore.barebox.org/barebox/20240404184001.1532897-1-a.fatoum@pengutronix.de/T/#mc182a3fe4eb2053a9af13e5a5e15c830a5ef468d, it got merged: https://git.pengutronix.de/cgit/barebox/commit/?id=a38be63831aa3c5696581a58c8afb5bfc11e45d5.

a3f:
Hello @pevik,

The version of barebox in DistroKit is still affected by the issue. I didn't deem it critical enough to backport fixes and will just wait until barebox v2024.04.0 is released and update to that. The kernel version is not important. Pinging the kernel works fine, but pinging barebox _sometimes_ was the problem (more on that below).

If you wish to reproduce, I can help with either a pcap capture showing the broken barebox replies or upload you my FAT partition including firmware and barebox binary. (A rootfs is not required).

The problem in barebox was a combination of two bugs:

  - The [ping reply code in barebox](https://github.com/barebox/barebox/blob/v2024.03.0/net/net.c#L649) is quite simple: It takes the received packet, does some fix ups, recalculates the checksums and sends it back. To decide how much bytes to do the ICMP checksum over, the code uses the frame size reported by the network driver. This assumes there are no trailing bytes in the ICMP echo request packet.

  - The pings sent by `ping(1)` are not so short as to need padding, so this was no problem, except on the Raspberry Pi 3b: barebox configured the smsc95xx driver erroneously for checksum offloading. The checksums are written as trailing bytes after the frame. This had the result that this particular network driver reported longer frames than what was actually transmitted over the wire

The combination of these two led to ICMP checksums being wrong, but only when barebox is pinged, not when barebox itself does the pinging. Given that `ping(1)` didn't verify checksums, this issue went unnoticed for years until I tried to ping a device sitting behind a conntrack-enabled router. This router seems to verify the ICMP checksums and dropped the ping packets, which led me to look at packet dumps. Wireshark correctly flagged the broken checksums in ICMP packets and so I started investigating how that came about and why it went unnoticed. The result is the barebox series linked above, which fixes both the issue in the network stack and that in the driver as well as this PR here that should ensure such issues are more easily detected in future. :)

Issue #410, opened 2022-06-15
ping: cleanup data types


pevik:
https://github.com/iputils/iputils/pull/394 fixed `-m` parameter (mark) to be unsigned int instead of plain int and added `strtoul_or_err()`. That was a good start, there should be more tree wide cleanup:
* check expected data types in kernel and use fixed data types (e.g. `uint32_t`)
* compile in CI also for 32bit and write test: `ping -m INT_MAX+1`
* remove `strtol_or_err()` if not needed (likely)
* add macro (DRY): `#define SAFE_STRTOUL(num, min, max) strtol_or_err(num, _("invalid argument"), min, max)`

PR #535, opened 2024-04-16
More time_t fixes

pevik:
@metan-ucw @Chozmao FYI (trivial but better to check).

Proper fix would be to use fixed data types https://github.com/iputils/iputils/issues/410.

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -284,8 +284,9 @@ static int recverr(struct run_state *const ctl)
 		struct timespec res;
 
 		timespecsub(&ts, retts, &res);
-		printf(_("%3ld.%03ldms "), res.tv_sec * 1000 + res.tv_nsec / 1000000,
-					   (res.tv_nsec % 1000000) / 1000);
+		printf(_("%3lld.%03ldms "), (long long int)res.tv_sec * 1000
+			   + res.tv_nsec / 1000000, (res.tv_nsec % 1000000) / 1000);
+
 		if (broken_router)
 			printf(_("(This broken router returned corrupted payload) "));
 	}
--- a/clockdiff.c
+++ b/clockdiff.c
@@ -623,7 +623,7 @@ int main(int argc, char **argv)
 				ctl.hisname, ctl.rtt, ctl.rtt_sigma, ctl.min_rtt,
 				ctl.measure_delta, ctl.measure_delta1, s);
 		} else {
-			printf("%ld %d %d\n", now, ctl.measure_delta, ctl.measure_delta1);
+			printf("%lld %d %d\n", (long long)now, ctl.measure_delta, ctl.measure_delta1);
 		}
 	}
 	exit(0);


tracepath.c
```diff
@@ -284,8 +284,9 @@ static int recverr(struct run_state *const ctl)
 		struct timespec res;
 
 		timespecsub(&ts, retts, &res);
-		printf(_("%3ld.%03ldms "), res.tv_sec * 1000 + res.tv_nsec / 1000000,
-					   (res.tv_nsec % 1000000) / 1000);
+		printf(_("%3lld.%03ldms "), (long long int)res.tv_sec * 1000
```
pevik, line 287: @nmeyerhans @okias Any opinion if we need to translate this? Does `ms` (as millisecond) really need to be translated? Some languages (ka, ja) translate it. We don't translate `"%ld %d %d\n"` in `clockdiff.c`

nmeyerhans: My inclination is that we don't need to translate this. I expect that "ms" is widely recognized and understood.  However, please note that I'm a native english speaker and don't consider myself much of an authority when it comes to localization matters.  The "C" locale is sufficient for me. :)

pevik: @yurchor @rffontenelle @marek22k @Jibec @simmon-nplob @oersen @NorwayFun @robert-scheck any oppinion about this, please?

Similar would be `"host=%s rtt=%ld(%ld)ms/%ldms delta=%dms/%dms %s\n"`

robert-scheck: In the German language, we use "ms". Not sure if I would translate "host", "rtt" and "delta" into German, because "host" would likely end up with "server" (well, replacing one English technical word with another), "rtt" has no German abbreviation (I'm personally even not used to the German long form) and "delta" is at least in German known from the Greeks related to mathematics and science.

However, from my understanding, "ms" might be translated in languages that don't use latin characters (e.g. Macedonian uses "мс" according to Wikipedia), but I'm not experienced with this.

metan-ucw: The "мс" is ms in Cyrilic script, so all countries that use Cyrilic use that.

Also ping on windows seems to be fully localized at least in Czech https://idoc.vsb.cz/xwiki/bin/view/tuonet/sit-nastaveni/ping/?language=cs not sure if doing that for iputils is worth the effort. At least I can see all the bug reports that we broke shell script that are parsing the output from ping/tracepath/etc.

metan-ucw:
The patches looks good to me. Reviewed-by: Cyril Hrubis <chrubis@suse.cz>

tracepath.c
yurchor: No objections from me.

rffontenelle: LGTM.

oersen: LGTM

pevik: Did I understand properly that we agree _not_ to translate `ms`? Because it is important to have common output. Also, some translations just add space between number and `ms`. I can create a ticket for solution we agree on I did not understood whether we agree on.

I wonder if I force myself to do ping output in json. It would not be translated and it would be a proper solution for scripts.

PR #545, opened 2024-05-28
ping: check return value of write() to avoid integer overflow

jamacku:
The problem was discovered by the Coverity static analysis tool.

```c
Error: INTEGER_OVERFLOW (CWE-190):
iputils-20240117/ping/ping.h:291: tainted_data_return: Called function "write(1, str + o, len - o)", and a possible return value may be less than zero. 
iputils-20240117/ping/ping.h:291: assign: Assigning: "cc" = "write(1, str + o, len - o)". 
iputils-20240117/ping/ping.h:292: overflow: The expression "o += cc" might be negative, but is used in a context that treats it as unsigned. 
iputils-20240117/ping/ping.h:291: overflow: The expression "len - o" is deemed underflowed because at least one of its arguments has underflowed. 
iputils-20240117/ping/ping.h:291: overflow_sink: "len - o", which might have underflowed, is passed to "write(1, str + o, len - o)".
#  289|           ssize_t cc;
#  290|           do {
#  291|->                 cc = write(STDOUT_FILENO, str + o, len - o);
#  292|                   o += cc;
#  293|           } while (len > o || cc < 0);
```

### DIFF
--- a/ping/ping.h
+++ b/ping/ping.h
@@ -290,8 +290,12 @@ static inline void write_stdout(const char *str, size_t len)
 	ssize_t cc;
 	do {
 		cc = write(STDOUT_FILENO, str + o, len - o);
-		o += cc;
-	} while (len > o || cc < 0);
+
+		if (cc < 0)
+			break;
+
+		o += (size_t) cc;
+	} while (len > o);
 }
 
 /*


pevik:
On the first look LGTM, but I need to have a deeper look (hopefully next week or after).

Please add your SOB (`Signed-off-by: Jan Macku <jamacku@redhat.com>`) to the commit message and force push (`git push -f`).

jamacku:
Thanks, it's not urgent.

ping/ping.h
```diff
@@ -289,8 +289,15 @@ static inline void write_stdout(const char *str, size_t len)
 	ssize_t cc;
 	do {
 		cc = write(STDOUT_FILENO, str + o, len - o);
-		o += cc;
-	} while (len > o || cc < 0);
+
+		if (cc < 0) {
+			if (errno == EAGAIN)
+				continue;
```
pevik, line 295: You don't mention `EAGAIN` at all in the commit message and it's unrelated to the integer overflow fix. Did you encounter any blocking in writing output? Is it really needed?

Also [`man write(2)`](https://man7.org/linux/man-pages/man2/write.2.html#ERRORS) mentions for `EAGAIN`: 

> `EAGAIN` The file descriptor fd refers to a file other than a socket and has been marked nonblocking (`O_NONBLOCK`), and the write would block.  See [`open(2)`](https://man7.org/linux/man-pages/man2/open.2.html) for further details on the `O_NONBLOCK` flag.

But we don't use `O_NONBLOCK` anywhere in the code.

metan-ucw: I do not think that write to stdout can produce `EAGAIN` here unless someone deliberately set the standard streams to `O_NONBLOCK` which would be a very stupid thing to do. And I suppose that this may even backfire and trigger an infinite loop if one is crafty enough. I had a look on what errors are handled in standard command line tools in bussybox and it seems that the inner loop that writes the data checks for `EINTR` which is probably the only sensible error to retry on when writing into stdout.

For instance in bussy box the tools write into stdout with [full_write()](https://github.com/brgl/busybox/blob/master/libbb/full_write.c) that calls [safe_write()](https://github.com/brgl/busybox/blob/master/libbb/safe_write.c) and the safe_write() retries only on `EINTR`.

jamacku: Thank you both for reviewing. I added the `EAGAIN` part because I thought it was a good practice. I didn't realize that it might cause some issues.

pevik:
I'm sorry I forget on this.
```
```

metan-ucw:

metan-ucw (approved):

Issue #538, opened 2024-05-13
Cannot run multiple arping simultaneously


shuangye:
Our system used an old iputils. It can run multiple `arping` simultaneously:

```console
./arping -I ethlmp0 -c 2 -w 10 192.168.255.126 &
./arping -I ethlmp0 -c 2 -w 10 192.168.255.126 &
[1]-  Done                    ./arping -I ethlmp0 -c 2 -w 10 192.168.255.126
[2]+  Done                    ./arping -I ethlmp0 -c 2 -w 10 192.168.255.126
Version: s20160308
```

We updated it to a new version recently. I found that the behavior changed. Only the last arping succeeded:

```console
arping -I ethlmp0 -c 2 -w 10 192.168.255.126 &
arping -I ethlmp0 -c 2 -w 10 192.168.255.126 &
[1]-  Exit 1                  arping -I ethlmp0 -c 2 -w 10 192.168.255.126
[2]+  Done                    arping -I ethlmp0 -c 2 -w 10 192.168.255.126
Version: 20211215
```

Then I compiled the latest release 20240117. Now both failed:

```console
/usr/local/bin/arping -V
arping from iputils 20240117
libcap: no, IDN: no, NLS: yes, error.h: yes, getrandom(): yes, __fpending(): yes

sudo /usr/local/bin/arping -I eth0 -c 2 -w 10 10.0.0.1 &
sudo /usr/local/bin/arping -I eth0 -c 2 -w 10 10.0.0.1 &
[1] 3008
[2] 3009
ARPING 10.0.0.1 from 10.0.0.201 eth0
ARPING 10.0.0.1 from 10.0.0.201 eth0
Unicast reply from 10.0.0.1 [10:0C:6B:D5:ED:9A]  2.436ms
Unicast reply from 10.0.0.1 [10:0C:6B:D5:ED:9A]  2.465ms
Unicast reply from 10.0.0.1 [10:0C:6B:D5:ED:9A]  2.765ms
Sent 1 probes (1 broadcast(s))
Received 2 response(s)
Unicast reply from 10.0.0.1 [10:0C:6B:D5:ED:9A]  2.800ms
Sent 1 probes (1 broadcast(s))
Received 2 response(s)

[1]-  Exit 1                  sudo /usr/local/bin/arping -I eth0 -c 2 -w 10 10.0.0.1
[2]+  Exit 1                  sudo /usr/local/bin/arping -I eth0 -c 2 -w 10 10.0.0.1
```

[...]

PR #546, opened 2024-06-03
arping: Fix exit code if receive more replies than sent

pevik:
ARP protocol, unlike ICMP protocol, has no way to link REQUEST and REPLY together (detect to which sender belongs the response). E.g. running more arping instances currently causes failure due receiving more replies than sent probes:

    # ./builddir/arping -c2 -I eth0 192.168.255.1 -w10 &
    # ./builddir/arping -c2 -I eth0 192.168.255.1 -w10 &
    ARPING 192.168.255.1 from 192.168.255.133 eth0
    ARPING 192.168.255.1 from 192.168.255.133 eth0
    Unicast reply from 192.168.255.1 [50:EB:F6:87:9D:D0]  1.722ms
    Unicast reply from 192.168.255.1 [50:EB:F6:87:9D:D0]  1.726ms
    Unicast reply from 192.168.255.1 [50:EB:F6:87:9D:D0]  1.910ms
    Unicast reply from 192.168.255.1 [50:EB:F6:87:9D:D0]  1.915ms
    Sent 1 probes (1 broadcast(s))
    Sent 1 probes (1 broadcast(s))
    Received 2 response(s)
    Received 2 response(s)
    [ ENTER ]
    [1]-  Exit 1                  ./builddir/arping -c2 -I eth0 192.168.255.1 -w10
    [2]+  Exit 1                  ./builddir/arping -c2 -I eth0 192.168.255.1 -w10

84ca65c (fix for 67e070d) introduced this regression.

Later e594ca5 introduced more precise timing - before arping sent 2 probes instead of 1 with -w1.

Then 854873b unified behavior with ping, i.e. using -w (deadline) *without* -c (count) exit 0 if at least one reply arrived (backwards incompatibility, also now incompatible with busybox). But that still kept problematic using -w with -c on multiple instances / replies.

Fixing the problem by adding a special condition.

Also, when at it, move all exit code evaluation into finish() (before it was in finish() but also event_loop()). This improves code introduced in 67e070d.

Fixes: 84ca65c ("arping: fix sent vs received packages return value")
Fixes: https://github.com/iputils/iputils/issues/538
Reported-by: Mingyang Liu <papillon@yeah.net>

### DIFF
--- a/arping.c
+++ b/arping.c
@@ -317,11 +317,18 @@ static int finish(struct run_state *ctl)
 		printf("\n");
 		fflush(stdout);
 	}
+
+	/* arping exit code evaluation */
 	if (ctl->dad)
-		return (!!ctl->received);
+		return !!ctl->received;
+
 	if (ctl->unsolicited)
 		return 0;
-	return (!ctl->received);
+
+	if (ctl->timeout && ctl->count > 0 && !ctl->quit_on_reply)
+		return !(ctl->count <= ctl->received);
+
+	return !ctl->received;
 }
 
 static void print_hex(unsigned char *p, int len)
@@ -693,7 +700,7 @@ static void find_broadcast_address(struct run_state *ctl)
 
 static int event_loop(struct run_state *ctl)
 {
-	int exit_loop = 0, rc = 0;
+	int exit_loop = 0;
 	ssize_t s;
 	enum {
 		POLLFD_SIGNAL = 0,
@@ -833,7 +840,7 @@ static int event_loop(struct run_state *ctl)
 					      (struct sockaddr *)&from, &addr_len)) < 0) {
 					error(0, errno, "recvfrom");
 					if (errno == ENETDOWN)
-						rc = 2;
+						return 2;
 					continue;
 				}
 				if (recv_pack
@@ -849,17 +856,8 @@ static int event_loop(struct run_state *ctl)
 	close(sfd);
 	close(tfd);
 	freeifaddrs(ctl->ifa0);
-	rc |= finish(ctl);
-	if (ctl->unsolicited)
-		/* nothing */;
-	else if (ctl->dad && ctl->quit_on_reply)
-		/* Duplicate address detection mode return value */
-		rc |= !(ctl->brd_sent != ctl->received);
-	else if (ctl->timeout && !(ctl->count > 0))
-		rc |= !(ctl->received > 0);
-	else
-		rc |= (ctl->sent != ctl->received);
-	return rc;
+
+	return finish(ctl);
 }
 
 int main(int argc, char **argv)


arping.c
```diff
@@ -317,11 +317,22 @@ static int finish(struct run_state *ctl)
...
+
 	if (ctl->unsolicited)
 		return 0;
-	return (!ctl->received);
+
+	if (ctl->timeout) {
+		if (ctl->count > 0)
+			return !(ctl->count <= ctl->received);
+		else
+			return !(ctl->received > 0);
```
cfconrad, line 332: nitty, but you use `!(ctl->received > 0)` and `!ctl->received` I guess I would use only one of it everywhere...

and yes, what you mentioned here https://github.com/iputils/iputils/issues/538#issuecomment-2146244343 would help a lot :-)

pevik: Thanks, correct! You're right, that `ctl->received > 0` can be shortened just to `ctl->received`, which makes the condition simpler (only `-c` together with `-w` counts how many replies we get):

```C
/* arping exit code evaluation */
if (ctl->dad)
	return !!ctl->received;

if (ctl->unsolicited)
	return 0;

if (ctl->timeout && ctl->count > 0)
	return !(ctl->count <= ctl->received);

return !ctl->received;
```

`ctl->count` and `ctl->interval` are by default `-1`, therefore they need to be checked against `0`.

> and yes, what you mentioned here https://github.com/iputils/iputils/issues/538#issuecomment-2146244343 would help a lot :-)

I'll guess I merge for now just this fix (to minimize changes) and send another PR which implements this + create exit code section for arping man page. There are many parts which could be improved, but it's a question how much of it should be done (as for all iputils tools: a complete rewrite would be the best to cleanup).

michalsieron:
Hi, unfortunately this breaks `-f` flag, because in such case `ctl->count` being less than `ctl->received` is expected.
There should probably be a check for `ctl->quit_on_reply` somewhere.

pevik:
@michalsieron @cfconrad @shuangye  Could you please retest?
@michalsieron Could you please send me your email address (either as a comment here or to my mail pvorel@suse.cz), so that I can add your Reported-by: and/or Reviewed-by: (ideally you would post these tag here yourself (e.g. https://github.com/iputils/iputils/pull/546#pullrequestreview-2098596101).

michalsieron:
Yup. Seems to working just fine. Fixing both original issue with parallel arpinging and not regressing exit code when using `-f`. Tested after cherry-picking onto v20211215.

Reviewed-by: Michał Sieroń [michalwsieron@gmail.com](mailto:michalwsieron@gmail.com)

cfconrad:
Nothing to add from my site, LGTM!

xanoxes:
I have some questions about this merge. In the old realization, the 'ctl->brd_sent' can identify sendto failure in DAD mode. Due to the new realization, if sendto failed in sometime such like bond4 negotiation, we won't know from the return value.

xanoxes:
https://github.com/iputils/iputils/issues/557

PR #548, opened 2024-07-07
ci: Remove Centos 7

pevik:
Centos 7 is EOL since 2024-06. Therefore CI build fails with:

```
Could not resolve host: mirrorlist.centos.org; Unknown error
```
@robert-scheck Can you please review?


robert-scheck (approved):
Yes, CentOS 7 reached EOL on 2024-06-30, let's finish this chapter.

Issue #542, opened 2024-05-24
Possible integer overflow with -s arg


mimicria:
Hi!
Tainted variable `rts.datalen` is limited by the value `INT_MAX`:

https://github.com/iputils/iputils/blob/4db1de672559804bebcb7073d231924339ca8cd8/ping/ping.c#L531
Integer value `rts.datalen + 28` is used by calling function `malloc` without checking its bounds:

https://github.com/iputils/iputils/blob/4db1de672559804bebcb7073d231924339ca8cd8/ping/ping.c#L578

So UBSAN finds integer overflow in: https://github.com/iputils/iputils/blob/4db1de672559804bebcb7073d231924339ca8cd8/ping/ping.c#L994
where addition to `rts->datalen` also occurs.
```
../ping/ping.c:994:7: runtime error: signed integer overflow: -2147483641 + -1090518520 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../ping/ping.c:994:7 in
```
Steps to reproduce:
1) Build with clang and sanitizers
```
export CC="clang"
export CFLAGS="-O0 -g -fsanitize=address,undefined"
export LDFLAGS="-O0 -g -fsanitize=address,undefined"
./configure
make
```
2) Configure SAN:
```
export UBSAN_OPTIONS="log_path=$PWD/ubsan.log:check_initialization_order=1:halt_on_error=0"
export ASAN_OPTIONS="log_path=$PWD/asan.log:detect_leaks=1:check_initialization_order=1:verbosity=0:detect_stack_use_after_return=1:halt_on_error=0"
export LSAN_OPTIONS="log_path=$PWD/lsan.log:halt_on_error=0"
```
3) Run `./ping -s 2147483647 192.168.0.1`
4) `cat ubsan.log.xxx`

Bug was found by the Svace static analyzer.

pevik:
While this is correct, better will be adjust code to have a real max data size:

* IPv4: max `-s 65507`: `65535` (max IPv4 packet) `- 20` (IPv4 header) `- 8` (ICMP header) `= 65507`
* IPv6: max `-s 65527`: `65575` (max IPv6 packet) `- 40` (IPv6 header) `- 8` (ICMPv6 header) `= 65527`

pevik:
It looks to me it's a regression from a647e2cdbd9772bdaed61bf053b6c92c323b6061, before the maximum was 128000 (max packet size) - 8 (ICMP{,v6} header).

[...]

Issue #210, opened 2019-08-27
unify ping -s packet size checks


ghormoon:
currently, if user does not think about IP packet size limit at all and guesses random numbers to use with ping -s, he might get two different errors:

```
$ ping 8.8.8.8 -s 500000
ping: packet size too large: 500000
$ ping 8.8.8.8 -s 90000
Error: packet size 90000 is too large. Maximum is 65507
```
seems it's handledseparately at two parts of ping.c:

```                case 's':
                        datalen = strtol_or_err(optarg, _("invalid argument"), 0, MAXPACKET - 8);
                        break;
```
where MAXPACKET is defined in pin.h as 128000 - is there any reason for this big number?

and 

```
	if (datalen > 0xFFFF - 8 - optlen - 20)
		error(2, 0, _("packet size %d is too large. Maximum is %d"),
datalen, 0xFFFF - 8 - 20 - optlen);
```

i suggest unifying the output to show the limit properly, so if user inputs some number over 128000, it will tell him the proper limit too. or maybe check for the exact size at that point already.

kerolasa:
These size limitations feel like premature optimisation. I think it would be better to let user to define any size and let sendmsg() or struct sock_extended_err to inform if payload does not fit to package.

kerolasa:
Ping is changed not to be too clever what is, can, or could be allowed package size, see commit a647e2cdbd9772bdaed61bf053b6c92c323b6061.

While working with this I did notice errors vary by telling mtu exceeded or sendmsg() cannot handle packet of that size. When pinging IPv4 / IPv6 addresses size limits appeared to vary somewhat, and localhost appears to allow rather big packages. Now when artificial limits are gone it's interesting to see how os, network, and remote systems are behaving with various size variations.

PR #550, opened 2024-08-16
ping: Fix integer overflow for high -s values

pevik:
There are various fixes, which started as fix integer overflow for high -s value. But there are also other fixes (e.g. `ping: Fix EMSGSIZE on -s > 65527 on ICMP datagram socket` in the second commit).

# ping: Fix integer overflow for high -s values
a647e2c commit increased value max allowed -s value from 127992 to INT_MAX (2147483647). To have space for this range datalen was changed from signed int to size_t, but it forgot to change other int variables which also work with this data (hold, packlen).

This fixes signed integer overflow:

```
$ export CC="clang"
$ export CFLAGS="-O0 -g -fsanitize=address,undefined" 
$ export LDFLAGS="-O0 -g -fsanitize=address,undefined" 
$ ./configure && make
$ ./builddir/ping/ping -s 2147483647 ::1
../ping/ping6_common.c:317:7: runtime error: signed integer overflow: -2147483641 + -1174404560 cannot be
   represented in type 'int' SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior 
../ping/ping6_common.c:317:7 ./builddir/ping/ping: WARNING: probably, rcvbuf is not enough to hold preload
   PING ::1 (::1) 2147483647 data bytes

$ ./builddir/ping/ping -s 2147483647 127.0.0.1
../ping/ping.c:997:7: runtime error: signed integer overflow: -2147483641 + -1090518520 cannot be
   represented in type 'int' SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior 
../ping/ping.c:997:7 ./builddir/ping/ping: WARNING: probably, rcvbuf is not enough to hold preload 
  PING 127.0.0.1 (127.0.0.1) 2147483647(2147483675) bytes of data.
```

NOTE: I'm not sure if it makes sense to allow -s higher than 65507 (IPv4) or 65527 (IPv6) - more than these is over limit for IPv4/IPv6 packet payload. It would have to be sent as multiple packets (IPv4 fragmentation, IPv6 next_header packet chain or jumbo extension header). IMHO it never worked. That is probably the reason, why other implementations (busybox, fping, inetutils) limit data size to <= 65535.

Fixes: a647e2c ("ping: allow any package size to be defined by user")
Fixes: https://github.com/iputils/iputils/issues/542
Reported-by: mimicria
Suggested-by: Marius Tomaschewski <mt@suse.com>

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -645,8 +645,9 @@ xml:id="man.ping">
         <listitem>
           <para>Specifies the number of data bytes to be sent. The
           default is 56, which translates into 64 ICMP data bytes
-          when combined with the 8 bytes of ICMP header
-          data.</para>
+          when combined with the 8 bytes of ICMP header data.
+          Maximum allowed value is 127992, but though most systems
+          limit this to a smaller, system-dependent number.</para>
         </listitem>
       </varlistentry>
       <varlistentry>
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -646,7 +646,10 @@ xml:id="man.ping">
           <para>Specifies the number of data bytes to be sent. The
           default is 56, which translates into 64 ICMP data bytes
           when combined with the 8 bytes of ICMP header data.
-          Maximum allowed value is 127992, but though most systems
+          The maximum allowed value is 65507 for IPv4
+          (65467 when <option>-R</option> or <option>-T</option>
+          or Intermediate <emphasis remap="I">hop</emphasis>s)
+          or 65527 for IPv6, but most systems
           limit this to a smaller, system-dependent number.</para>
         </listitem>
       </varlistentry>
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -59,6 +59,7 @@
 #include <ifaddrs.h>
 #include <math.h>
 #include <locale.h>
+#include <sys/param.h>
 
 /* FIXME: global_rts will be removed in future */
 struct ping_rts *global_rts;
@@ -83,6 +84,12 @@ ping_func_set_st ping4_func_set = {
 #define	MAXICMPLEN	76
 #define	NROUTES		9		/* number of record route slots */
 #define TOS_MAX		255		/* 8-bit TOS field */
+
+/* max. IPv4 packet size - IPv4 header size - ICMP header size */
+#define ICMP_MAX_DATALEN (MAXPACKET - 20 - 8)
+
+/* max. IPv6 payload size - ICMPv6 Echo Reply Header */
+#define ICMPV6_MAX_DATALEN (MAXPACKET - sizeof (struct icmp6_hdr))
 
 #define CASE_TYPE(x) case x: return #x;
 
@@ -345,6 +352,8 @@ main(int argc, char **argv)
 		.ni.query = -1,
 		.ni.subject_type = -1,
 	};
+	unsigned char buf[sizeof(struct in6_addr)];
+
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
 
@@ -535,7 +544,8 @@ main(int argc, char **argv)
 			rts.opt_so_dontroute = 1;
 			break;
 		case 's':
-			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, MAXPACKET - 8);
+			/* real validation is done later */
+			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, INT_MAX);
 			break;
 		case 'S':
 			rts.sndbuf = strtol_or_err(optarg, _("invalid argument"), 1, INT_MAX);
@@ -622,6 +632,24 @@ main(int argc, char **argv)
 			hints.ai_family = AF_INET;
 	}
 
+	int max_s = MAX(ICMP_MAX_DATALEN, ICMPV6_MAX_DATALEN);
+
+	/* Detect based on -4 / -6 */
+	if (hints.ai_family == AF_INET)
+		max_s = ICMP_MAX_DATALEN - get_ipv4_optlen(&rts);
+	else if (hints.ai_family == AF_INET6)
+		max_s = ICMPV6_MAX_DATALEN;
+
+	/* Force limit on IPv4/IPv6 adresses */
+	if (inet_pton(AF_INET, target, buf))
+		max_s = ICMP_MAX_DATALEN - get_ipv4_optlen(&rts);
+	else if (inet_pton(AF_INET6, target, buf))
+		max_s = ICMPV6_MAX_DATALEN;
+
+	if (rts.datalen > max_s)
+		error(EXIT_FAILURE, 0, "invalid -s value: '%d': out of range: 0 <= value <= %d",
+		      rts.datalen, max_s);
+
 	if (rts.opt_verbose)
 		error(0, 0, "sock4.fd: %d (socktype: %s), sock6.fd: %d (socktype: %s),"
 			   " hints.ai_family: %s\n",
@@ -643,7 +671,6 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
-	unsigned char buf[sizeof(struct in6_addr)];
 	if (!strchr(target, '%') && sock6.socktype == SOCK_DGRAM &&
 		inet_pton(AF_INET6, target, buf) > 0 &&
 		(IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf))) {
@@ -1010,11 +1037,6 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 		if (setsockopt(sock->fd, IPPROTO_IP, IP_TTL, &ittl, sizeof ittl) == -1)
 			error(2, errno, _("cannot set unicast time-to-live"));
 	}
-
-
-	if (rts->datalen > 0xFFFF - 8 - rts->optlen - 20)
-		error(2, 0, _("packet size %d is too large. Maximum is %d"),
-		      rts->datalen, 0xFFFF - 8 - 20 - rts->optlen);
 
 	if (rts->datalen >= (int)sizeof(struct timeval))	/* can we time transfer */
 		rts->timing = 1;
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -756,6 +756,7 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 	unsigned char rspace[3 + 4 * NROUTES + 1];	/* record route space */
 	uint32_t *tmp_rspace;
 	struct sockaddr_in dst;
+	int ret;
 
 	if (argc > 1) {
 		if (rts->opt_rroute)
@@ -1034,10 +1035,10 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 
 	drop_capabilities();
 
-	hold = main_loop(rts, &ping4_func_set, sock, packet, packlen);
+	ret = main_loop(rts, &ping4_func_set, sock, packet, packlen);
 	free(packet);
 
-	return hold;
+	return ret;
 }
 
 static void pr_options(struct ping_rts *rts, unsigned char *cp, int hlen)
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1463,7 +1463,7 @@ int ping4_receive_error_msg(struct ping_rts *rts, socket_st *sock)
 		if (rts->opt_flood)
 			write_stdout("E", 1);
 		else if (e->ee_errno != EMSGSIZE)
-			error(0, 0, _("local error: %s"), strerror(e->ee_errno));
+			error(0, e->ee_errno, _("local error"));
 		else
 			error(0, 0, _("local error: message too long, mtu=%u"), e->ee_info);
 		rts->nerrors++;
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -114,6 +114,14 @@ static char *str_socktype(int socktype)
 	return "";
 }
 
+static int get_ipv4_optlen(struct ping_rts *rts)
+{
+	if (rts->opt_rroute || rts->opt_timestamp || rts->opt_sourceroute)
+		return 40;
+
+	return 0;
+}
+
 static void create_socket(struct ping_rts *rts, socket_st *sock, int family,
 			  int socktype, int protocol, int requisite)
 {
@@ -937,7 +945,6 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 		rspace[1 + IPOPT_OPTVAL] = IPOPT_RR;
 		rspace[1 + IPOPT_OLEN] = sizeof(rspace) - 1;
 		rspace[1 + IPOPT_OFFSET] = IPOPT_MINOFF;
-		rts->optlen = 40;
 		if (setsockopt(sock->fd, IPPROTO_IP, IP_OPTIONS, rspace, sizeof rspace) < 0)
 			error(2, errno, "record route");
 	}
@@ -960,7 +967,6 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 			if (setsockopt(sock->fd, IPPROTO_IP, IP_OPTIONS, rspace, rspace[1]) < 0)
 				error(2, errno, "ts option");
 		}
-		rts->optlen = 40;
 	}
 	if (rts->opt_sourceroute) {
 		int i;
@@ -976,8 +982,9 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 
 		if (setsockopt(sock->fd, IPPROTO_IP, IP_OPTIONS, rspace, 4 + rts->nroute * 4) < 0)
 			error(2, errno, "record route");
-		rts->optlen = 40;
 	}
+
+	rts->optlen = get_ipv4_optlen(rts);
 
 	/* Estimate memory eaten by single packet. It is rough estimate.
 	 * Actually, for small datalen's it depends on kernel side a lot. */
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -317,7 +317,6 @@ main(int argc, char **argv)
 	socket_st sock4 = { .fd = -1 };
 	socket_st sock6 = { .fd = -1 };
 	char *target;
-	char *outpack_fill = NULL;
 	static struct ping_rts rts = {
 		.interval = 1000,
 		.preload = 1,
@@ -515,10 +514,7 @@ main(int argc, char **argv)
 			break;
 		case 'p':
 			rts.opt_pingfilled = 1;
-			free(outpack_fill);
-			outpack_fill = strdup(optarg);
-			if (!outpack_fill)
-				error(2, errno, _("memory allocation failed"));
+			fill(&rts, optarg, rts.outpack, sizeof(rts.outpack));
 			break;
 		case 'q':
 			rts.opt_quiet = 1;
@@ -531,7 +527,7 @@ main(int argc, char **argv)
 			rts.opt_so_dontroute = 1;
 			break;
 		case 's':
-			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, INT_MAX);
+			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, MAXPACKET - 8);
 			break;
 		case 'S':
 			rts.sndbuf = strtol_or_err(optarg, _("invalid argument"), 1, INT_MAX);
@@ -577,14 +573,6 @@ main(int argc, char **argv)
 		error(2, EDESTADDRREQ, "usage error");
 
 	target = argv[argc - 1];
-
-	rts.outpack = malloc(rts.datalen + 28);
-	if (!rts.outpack)
-		error(2, errno, _("memory allocation failed"));
-	if (outpack_fill) {
-		fill(&rts, outpack_fill, rts.outpack, rts.datalen);
-		free(outpack_fill);
-	}
 
 	/* Create sockets */
 	enable_capability_raw();
@@ -695,7 +683,6 @@ main(int argc, char **argv)
 	}
 
 	freeaddrinfo(result);
-	free(rts.outpack);
 
 	return ret_val;
 }
@@ -1017,6 +1004,11 @@ int ping4_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 			error(2, errno, _("cannot set unicast time-to-live"));
 	}
 
+
+	if (rts->datalen > 0xFFFF - 8 - rts->optlen - 20)
+		error(2, 0, _("packet size %d is too large. Maximum is %d"),
+		      rts->datalen, 0xFFFF - 8 - 20 - rts->optlen);
+
 	if (rts->datalen >= (int)sizeof(struct timeval))	/* can we time transfer */
 		rts->timing = 1;
 	packlen = rts->datalen + MAXIPLEN + MAXICMPLEN;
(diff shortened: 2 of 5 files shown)


ping/ping6_common.c
```diff
@@ -104,8 +104,7 @@ unsigned int if_name2index(const char *ifname)
 int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 	      struct socket_st *sock)
 {
-	int hold, packlen;
-	size_t i;
+	size_t hold, i, packlen;
```
metan-ucw, line 107: The change for hold from `int`' to '`size_t` seem to be suspicions. We do pass the hold to a `setsockopt()` later in the code in:

```
        hold = 1;
        if (setsockopt(sock->fd, IPPROTO_IPV6, IPV6_RECVERR, (const void *)&hold, sizeof hold))
                error(2, errno, "IPV6_RECVERR");
```

Does that still work as expected?

Also the ping6_common() misuses the hold for a return value, I guess that it would be cleaner to write it as:

```
diff --git a/ping/ping6_common.c b/ping/ping6_common.c
index eb3b0d2..8431d56 100644
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -410,9 +410,9 @@ int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 
        drop_capabilities();
 
-       hold = main_loop(rts, &ping6_func_set, sock, packet, packlen);
+       int ret = main_loop(rts, &ping6_func_set, sock, packet, packlen);
        free(packet);
-       return hold;
+       return ret;
 }
```

pevik: > The change for `hold` from `int` to `size_t` seem to be suspicions. We do pass the hold to a `setsockopt()` later in the code in:

Very good catch. Strangely, ping did not warn even on non-zero exit code. I suppose `hold` must be changed as well, because it also stores value of `size_t`. But I created `int ret` and use it in separate commit.

ping/ping6_common.c
```diff
@@ -304,7 +303,7 @@ int ping6_run(struct ping_rts *rts, int argc, char **argv, struct addrinfo *ai,
 		rts->timing = 1;
 	}
 	packlen = rts->datalen + 8 + 4096 + 40 + 8; /* 4096 for rthdr */
-	if (!(packet = (unsigned char *)malloc((unsigned int)packlen)))
+	if (!(packet = malloc((unsigned int)packlen)))
```
metan-ucw, line 306: The cast to unsigned int here is useless as well. We are passing size_t here now which is the proper type to be passed to malloc().

ping/ping6_common.c
metan-ucw: What about the part where we pass sizeof(hold) to the setsockopt() is that code okay with hold changed to size_t?

And looking at kernel sources it does:

```
                if (optlen >= sizeof(int)) {
                        if (copy_from_sockptr(&val, optval, sizeof(val)))
                                return -EFAULT;
                 }
```

So if we pass size bigger sizeof(int) kernel will take a first sizeof(int) bytes from the value and use that. Now on 64bit the sizeof(int) == 32 and sizeof(size_t) == 64. So we take only the half of the value, on little-endian things are fine, since we will take the lower half of the integer and the value will be fine. But on big-endian we will take the upper half of the integer and wrongly interpret the value as 0.

So I guess that we should add another patch that decouples the use of the hold from the setsockopt() calls as well.

metan-ucw (changes requested):

gobenji:
> NOTE: I'm not sure if it makes sense to allow -s higher than 65507 (IPv4) or 65527 (IPv6) - more than these is over limit for IPv4/IPv6 packet payload. It would have to be sent as multiple packets (IPv4 fragmentation, IPv6 next_header packet chain or jumbo extension header).

I checked in the kernel code if it's possible to send a ping larger than ~64k. I don't think so.

For ping sockets, the length is clearly limited in:
```
int ping_common_sendmsg(int family, struct msghdr *msg, size_t len,
			void *user_icmph, size_t icmph_len)
{
	u8 type, code;

	if (len > 0xFFFF)
		return -EMSGSIZE;
```

For SOCK_RAW sockets, in case of ipv6 the error comes from __ip6_append_data(), I think because of:
```
/* We do not (yet ?) support IPv6 jumbograms (RFC 2675)
 * Unlike IPv4, hdr->seg_len doesn't include the IPv6 header
 */
#define IP6_MAX_MTU (0xFFFF + sizeof(struct ipv6hdr))
```

In case of ipv4, the error comes from net/ipv4/raw.c raw_sendmsg():
```
	err = -EMSGSIZE;
	if (len > 0xFFFF)
		goto out;
```

It seems reasonable to limit `-s <size>` to 64k or just under based on your calculations with the headers.

pevik:
> What about the part where we pass sizeof(hold) to the setsockopt() is that code okay with hold changed to size_t?
> 
> And looking at kernel sources it does:
> 
> ```
>                 if (optlen >= sizeof(int)) {
>                         if (copy_from_sockptr(&val, optval, sizeof(val)))
>                                 return -EFAULT;
>                  }
> ```

Ah, you mean [`do_ip_setsockopt()`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/ip_sockglue.c#n892) and [`do_ipv6_setsockopt()`](
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv6/ipv6_sockglue.c#n393). Looking, how exactly it's connected to [`do_sock_setsockopt()`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/socket.c#n2291), which is called later by `SYSCALL_DEFINE5(setsockopt, ...` they are called by `struct proto` `.setsockopt` members for both raw and ICMP datagram sockets (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/ping.c#n1008, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/raw.c#n938, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv6/raw.c#n1210, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv6/ping.c#n207).

> 
> So if we pass size bigger sizeof(int) kernel will take a first sizeof(int) bytes from the value and use that. Now on 64bit the sizeof(int) == 32 and sizeof(size_t) == 64. So we take only the half of the value, on little-endian things are fine, since we will take the lower half of the integer and the value will be fine. But on big-endian we will take the upper half of the integer and wrongly interpret the value as 0.
> 
> So I guess that we should add another patch that decouples the use of the hold from the setsockopt() calls as well.

I'll have look into this.

@gobenji WDYT?

MUSL [calls raw `setsockopt()` syscall with `long`](https://git.musl-libc.org/cgit/musl/tree/src/network/setsockopt.c#n32) (but only for 32bit, due return earlier), I'm not sure if this is valid for 64 bit.

pevik:
> So I guess that we should add another patch that decouples the use of the hold from the setsockopt() calls as well.

IMHO this is impossible to do: 3rd option `alloc` in `sock_setbufs()` is passed as `setsockopt()` parameter:

https://github.com/iputils/iputils/blob/0f12e6d5bb325df4eb9273b1e294a2cf94a53120/ping/ping_common.c#L440-L457

But this parameter is calculated from `rts->datalen`:
https://github.com/iputils/iputils/blob/0f12e6d5bb325df4eb9273b1e294a2cf94a53120/ping/ping.c#L996-L998

https://github.com/iputils/iputils/blob/0f12e6d5bb325df4eb9273b1e294a2cf94a53120/ping/ping6_common.c#L316-L318

`rts->optlen` is mostly `0`, but can be `40` for `-R` or `-Ttsprespec`.
For IPv4 datalen = 2147483647 optlen = 40: hold = datalen + 8; hold += ((hold + 511) / 512) * (optlen + 20 + 16 + 64 + 160) = 3405775158.515625
For IPv6 datalen = 2147483647; hold = datalen + 8; hold += ((hold + 511) / 512) * (40 + 16 + 64 + 160) = 3321889058.28125

It looks like the highest possible max `-s` would be 1354000000.

I think a647e2cdbd9772bdaed61bf053b6c92c323b6061 (which tried to unify error messages - https://github.com/iputils/iputils/issues/210) was simply wrong. I tend to revert it (I wonder if there is something useful from it nowadays), maybe remove extra `datalen > 0xFFFF - 8 - optlen - 20` check (we still have final `EMSGSIZE` error message) and replace `datalen = strtol_or_err(optarg, _("invalid argument"), 0, MAXPACKET - 8);` with 65507 or 65527 (https://github.com/pevik/iputils/wiki/Maximum-value-for-%E2%80%90s-(size)).

gobenji:
> So if we pass size bigger sizeof(int) kernel will take a first sizeof(int) bytes from the value and use that.
> [...]
> @gobenji WDYT?

I don't have a big endian machine handy to confirm the effect of passing a 64bit value but Cyril's analysis seems correct. Using an "int" variable would be my first choice; it's what setsockopt(2) "suggests":

> Most socket-level options utilize an int argument for optval.

The only point I can add is that the kernel code (in do_ipv6_setsockopt()) would reject something smaller than int:
```
	case IPV6_RECVERR:
		if (optlen < sizeof(int))
			return -EINVAL;
```

pevik:
@gobenji thanks a lot. Would you please have sometimes this week time for review of this patchset? I'd like to solve the issue and move on to make a new release.

doc/ping.xml
```diff
@@ -646,7 +646,10 @@ xml:id="man.ping">
           <para>Specifies the number of data bytes to be sent. The
           default is 56, which translates into 64 ICMP data bytes
           when combined with the 8 bytes of ICMP header data.
-          Maximum allowed value is 127992, but though most systems
+          Maximum allowed value is for IPv4 65507
+          (65467 when <option>-R</option> or <option>-T</option>
+          or Intermediate <emphasis remap="I">hop</emphasis>s),
+          for IPv6 65527, but though most systems
```
gobenji, line 652: I think the following reads better:

```
The maximum allowed value is 65507 for IPv4
(65467 when <option>-R</option> or <option>-T</option>
or Intermediate <emphasis remap="I">hop</emphasis>s)
or 65527 for IPv6, but most systems
```

ping/ping.c
```diff
@@ -337,6 +344,9 @@ main(int argc, char **argv)
 		.ni.query = -1,
 		.ni.subject_type = -1,
 	};
+
```
gobenji, line 347: Why empty line?

ping/ping.c
```diff
@@ -84,6 +85,12 @@ ping_func_set_st ping4_func_set = {
 #define	NROUTES		9		/* number of record route slots */
 #define TOS_MAX		255		/* 8-bit TOS field */
 
+/* max. IPv4 packet size - IPv4 header size - ICMP header size */
+#define ICMP_MAX_DATALEN (65535 - 20 - 8)
+
+/* max. IPv6 payload size) - ICMPv6 Echo Reply Header */
```
gobenji, line 91: stray ')'

gobenji (commented):
From the patch 2 log:
> This effectively reverts 0ab270b9232952ef51ddbf3e2e37a11cb5f2dfa0.

Is this commit id correct? It looks like it should be a647e2cdbd9772bdaed61bf053b6c92c323b6061 like in the following paragraph.

> which also work with this data (hold, packlen) to be able to store
> values > INT_MAX*2/3.

Where does 2/3 come from? I thought it's from
```
	hold = rts->datalen + 8;
	hold += ((hold + 511) / 512) * (rts->optlen + 20 + 16 + 64 + 160);
```
But with optlen 0 or 40, it works out to 0.51 or 0.58. Anyways, it's not a big deal.

From the patch 3 log:
> (Busybox: 65535, fping: 65507, inetutils: IPv4: 65399, IPv6: 65527), I
> suppose due 2, FreeBSD limits on IPv4: 65507).

The parens are unbalanced; "I suppose due 2" looks like it doesn't belong.

I think MAXPACKET can be lowered in patch 3. Maybe to 65535?

ping/ping.c
pevik: Readability. iputils 30 years code style si horrible. But sure, I can remove it.

doc/ping.xml
pevik: Thanks, fixed.

ping/ping.c
pevik: Thanks, removed!

pevik:
> I think MAXPACKET can be lowered in patch 3. Maybe to `65535`?

You're probably right, it would be a simple change, I'll probably modify it tomorrow.

BTW although higher values aren't currently possible, it could be done:
https://lore.kernel.org/netdev/CANn89iK6Zr4bTaMOevGkNZ-KYHGFaE-8x5y95UgJ5+AwJgdwJg@mail.gmail.com/
I tried that, but obviously more work would be needed (besides the limit on ICMP datagram socket), I'm not sure if jumboheader should be on ping - userspace side or in kernel. I was for a moment thinking whether keep just the revert (not lower the value to `65535`), but I don't think anybody is missing this functionality, thus we can lower the value (if anybody implements it, it can be raised again).

gobenji:
> but I don't think anybody is missing this functionality

Yeah, exactly. If someone is patching their kernel to enable larger icmp packets, they can patch their iputils too.

gobenji (commented):
> * raw IPv6 socket is limited similarly in __ip6_append_data() in
>   include/net/ip6_route.h:
> 
>     /* We do not (yet ?) support IPv6 jumbograms (RFC 2675)
>     * Unlike IPv4, hdr->seg_len doesn't include the IPv6 header
>     */
>     #define IP6_MAX_MTU (0xFFFF + sizeof(struct ipv6hdr))

According to Eric's [email](https://lore.kernel.org/netdev/CANn89iK6Zr4bTaMOevGkNZ-KYHGFaE-8x5y95UgJ5+AwJgdwJg@mail.gmail.com/), the limit is due to IPV6_MAXPLEN. I was mistaken in my earlier comment.

pevik:
Also, back to `65535` packet size limitation (if we ignore future possibility to IPv6 jumbograms, i.e. limit to current kernel limitation): it should probably be `65527` (`-s` is payload, this will be on IPv6 `65535` in kernel check, `-s` >  `65527` currently fails). And with that for IPv4 `-s 65507` generates  `65535` packet (`-s` > `65507` fails). I suppose busybox chose `65535` due simplicity (they have helpers for datalen https://git.busybox.net/busybox/tree/networking/ping.c#n955).
OTOH  `65535` would allow to test possible future kernel IPv6 jumbograms support.

pevik:
> > * raw IPv6 socket is limited similarly in __ip6_append_data() in
> >   include/net/ip6_route.h:
> >   /* We do not (yet ?) support IPv6 jumbograms (RFC 2675)
> >   
> >   * Unlike IPv4, hdr->seg_len doesn't include the IPv6 header
> >     */
> >     #define IP6_MAX_MTU (0xFFFF + sizeof(struct ipv6hdr))
> 
> According to Eric's [email](https://lore.kernel.org/netdev/CANn89iK6Zr4bTaMOevGkNZ-KYHGFaE-8x5y95UgJ5+AwJgdwJg@mail.gmail.com/), the limit is due to IPV6_MAXPLEN. I was mistaken in my earlier comment.

I updated the commit message.

ping/ping.c
```diff
@@ -84,6 +85,12 @@ ping_func_set_st ping4_func_set = {
 #define	NROUTES		9		/* number of record route slots */
 #define TOS_MAX		255		/* 8-bit TOS field */
 
+/* max. IPv4 packet size - IPv4 header size - ICMP header size */
+#define ICMP_MAX_DATALEN (65535 - 20 - 8)
+
+/* max. IPv6 payload size - ICMPv6 Echo Reply Header */
+#define ICMPV6_MAX_DATALEN (65535 - sizeof (struct icmp6_hdr))
+
```
metan-ucw, line 93: Shouldn't we also lower the MAXPACKET now to 65535? We declare it on the stack almost twice as big, fill it with data, buf never use the upper half after this change.

ping/ping.c
```diff
@@ -527,7 +536,7 @@ main(int argc, char **argv)
 			rts.opt_so_dontroute = 1;
 			break;
 		case 's':
-			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, MAXPACKET - 8);
+			rts.datalen = strtol_or_err(optarg, _("invalid argument"), 0, max_s);
```
metan-ucw, line 539: I guess that I would avoid limiting the upper bound here, because the switch we do later on has a better error message, so I would pass LONG_MAX as the upper bound here instead.

metan-ucw (changes requested):
Looks mostly good, minor stuff pointed out in the comments.

ping/ping.c
gobenji: datalen is an int. Limitting to LONG_MAX leads to a strange behavior:
```
$ ./builddir/ping/ping -s -1 localhost
./builddir/ping/ping: invalid argument: '-1': out of range: 0 <= value <= 9223372036854775807

$ ./builddir/ping/ping -s 4294967316 localhost
PING localhost (::1) 20 data bytes
28 bytes from localhost (::1): icmp_seq=1 ttl=64 time=0.042 ms
```

metan-ucw: Ah, missed that, so we should limit it to INT_MAX to get the nicer message for reasonably larger values.

pevik:
I implemented 2 changes Cyril requested (+ use `MAXPACKET` in `ICMP_MAX_DATALEN` and `ICMPV6_MAX_DATALEN`) + add few minor related cleanup.
I also set proper size when pinging numeric target (i.e. IPv4/IPv6 address).

Also I found that forcing a correct payload (`-s` value) fixed a slowdown introduced in 87dbb3a5db657d5eae6934707beaf0507980a1c3: on `-s <65535-65528>` on both IPv4 and IPv6 on both SOCK_DGRAM which took too long (`recvmsg()` returned `EAGAIN` but for loop not exited with break, thus it continued till the timeout).

ping/ping.c
pevik: FYI I put there `INT_MAX`.

pevik: @gobenji Thanks for your time!

ping/ping.c
```diff
@@ -647,7 +647,7 @@ main(int argc, char **argv)
 	int target_ai_family = hints.ai_family;
 	hints.ai_family = AF_UNSPEC;
 
-	unsigned char buf[sizeof(struct in6_addr)];
+	char buf[INET6_ADDRSTRLEN];
```
gobenji: I don't think this is correct. buf used for inet_pton() will contain a binary address structure. It's not related to  INET6_ADDRSTRLEN. inet_pton(3) documents the needed buffer size and I think it's already correct before this patch.

pevik: Thanks a lot, I'll remove the patch (I only move the variable declaration so that it can be reused).

pevik:
@gobenji @metan-ucw I added 2 new commits (`EMSGSIZE` on -s > 65527 + other trivial fix). Hopefully it should be ready to merge.

ping/ping.c
pevik: @gobenji  FYI `sizeof(struct in6_addr)` is `16`, which is the same as `INET_ADDRSTRLEN`. `INET6_ADDRSTRLEN` is `46`, which should be for really strange addresses, e.g. IPv4-mapped formIPv4-mapped form `ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255`, but it was working with original form, thus I remove this commit to speedup the review.

ping/ping.c
```diff
@@ -1420,8 +1420,13 @@ int ping4_receive_error_msg(struct ping_rts *rts, socket_st *sock)
...
-		if (errno == EAGAIN || errno == EINTR)
+		if (errno == EAGAIN || errno == EINTR) {
 			local_errors++;
+
+			/* SOCK_DGRAM fails on -s > 65527  */
+			if (saved_errno == EMSGSIZE)
+				rts->nerrors++;
+		}
 		goto out;
 	}
```
metan-ucw, line 1453: Uff this is an ugly hack, so we depend on the fact that the errno is set by the sendto() in ping4_send_probe() but we do not clear the errno before the sendto() so it may have been set by something in the ping initialization code. We really should check if the sendto() returned -1 before we check the errno. Is there a reason why can't we increment the rts->nerrors directly in the send_probe functions?

pevik: Thanks for catching this. I agree, this is not elegant and errorprone. I guess the handling is then done in `pinger()` function, I would put it there (after `else if (errno == ENOBUFS || errno == ENOMEM)`), it would be on a single place:

https://github.com/iputils/iputils/blob/0f12e6d5bb325df4eb9273b1e294a2cf94a53120/ping/ping_common.c#L361-L380

Here we could even consider early quit (why to keep sending packets which are dropped)?

I suppose the place for validating recvmsg() handled slightly differently for IPv4 and IPv6 is forced by protocol differences.

pevik: @metan-ucw FYI the error handling moved, let's continue in the new code location.

ping/ping_common.c
```diff
@@ -401,6 +399,10 @@ int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock)
 		/* Socket buffer is full. */
 		tokens += rts->interval;
 		return MIN_INTERVAL_MS;
+	} else if (errno == EMSGSIZE) {
+		/* SOCK_DGRAM fails on -s > 65527 */
+		rts->nerrors++;
+		i = 0;
```
pevik, line 405: Maybe I should just quit here because it does not makes sense to keep sending (it will fail anyway, kernel will refuse them all, it's not MTU based). @metan-ucw @gobenji WDYT?

I would implemented it probably like:
```c 
error(2, errno, "sendmsg");
```
which would not print statistics.

metan-ucw: The error message in ping4_receive_error_msg() seems to suggest that it used to be MTU based or may be MTU based in some cases, see the "error(0, 0, _("local error: message too long, mtu=%u"), e->ee_info);". That does look like a leftover though, since in my testing this error gets triggered if we cross the 64k packet size by just a small bit and it prints the network MTU which is a few order of magnitudes smaller than the maximal allowed packet size.

pevik: I wonder myself why `CMSG_FIRSTHDR()` did not catch this. I need to search, but I guess it was for the raw socket.

metan-ucw: I mean it does print the correct MTU but I have no idea what is the connection to maximal ping packet size here. My guess would be that decades ago ping packet size was limited by MTU, now it's not but we still get MTU printed in the case of EMSGSIZE which does seem to be confusing.

ping/ping_common.c
```diff
@@ -401,6 +401,10 @@ int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock)
 		/* Socket buffer is full. */
 		tokens += rts->interval;
 		return MIN_INTERVAL_MS;
+	} else if (errno == EMSGSIZE) {
+		/* SOCK_DGRAM fails on -s > 65527 */
```
gobenji, line 403: I think the comment is confusing since SOCK_RAW also fails. It will also be confusing after '-s' gets clamped down.
How about changing it to: "for example, sendto with len > 65527 on SOCK_DGRAM fails with this errno"

gobenji (commented):
>     Reproducible with:
> 
>         # sysctl -w net.ipv4.ping_group_range="0 2147483647" # ICMP datagram socket
>         $ ./builddir/ping/ping -v -i0.2 -c2 -s 65528 127.0.0.1
>         $ ./builddir/ping/ping -v -i0.2 -c2 -s 65528 ::1
> 
>         # sysctl -w net.ipv4.ping_group_range="0 2147483647" # raw socket socket

This command is a copy/paste of the previous one and seems incorrect. Is it meant to be ="1 0"?

gobenji (commented):
This seems like a reasonable change but unrelated to the rest of the MR to "Fix integer overflow". I would leave it out for another MR of cleanups.

ping/ping_common.c
pevik: Thanks fixed!

pevik:
> This seems like a reasonable change but unrelated to the rest of the MR to "Fix integer overflow". I would leave it out for another MR of cleanups. Edit: this comment is about commit "ping: Use error() consistently".

Yes, it's not related. OTOH it's so hard to find anybody for a review, that I decided to squeeze it in.

pevik:
> ...
> This command is a copy/paste of the previous one and seems incorrect. Is it meant to be ="1 0"?

Thanks, also fixed!

metan-ucw (approved):
Latest version looks good to me.

pevik:
@gobenji I'd like to merge tonight. Would you add your RBT or ABT?

pevik:
Only added Cyril's RBT (no functional change).

gobenji (approved):
Reviewed-by: Benjamin Poirier <benjamin.poirier@gmail.com>

PR #556, opened 2024-09-18
meson: fix installation of prebuilt manpages in dist tarball

eli-schwartz:
page contains the name of the program we are building/installing a manual for. In the custom targets for building the manpages, we append .xml and .8 in turn -- we don't replace a ".xml" in the page name.

As there is nothing to replace, we end up checking if e.g. "./arping" exists, and trying to install it as a manual if so -- rather than arping.8 as intended. Sync the logic so that this works.

---

Reproduce by trying to install iputils from a release tarball without xsltproc.

@thesamesam oops. :(

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -28,8 +28,8 @@ if meson.version().version_compare('>=0.58.0')
 	out_html_pages = []
 	generated_docs_exist = true
 	foreach page : manpages
-		man_page_name = page.replace('.xml', '.8')
-		html_page_name = page.replace('.xml', '.html')
+		man_page_name = page + '.8'
+		html_page_name = page + '.html'
 		out_man_pages += man_page_name
 		out_html_pages += html_page_name
 		if not fs.exists(man_page_name) or not fs.exists(html_page_name)


eli-schwartz:
CI is all broken due to 2e8b3b27e8e779d1e18a4dcbf9aa28d4e31a5b4a and 6d1f6e0f822db602c9569b2a668c22421bd96c2b.

- The first commit implemented a check for meson dist that hardcoded a specific release version instead of the current release version.
- The second commit fixed the check to pass any time the current release version is the same as the present date.

Either check for any tarball in meson-dist or extract the version number from meson.build / the most recent git tag. I guess it depends on what exactly you're trying to test?

pevik:
@eli-schwartz Thank you for the fix. I'm sorry for late reply (very long vacation) and issues with the CI (fixed today).

eli-schwartz:
Thanks, and no problem. :) I have had some pretty late response times myself recently due to the holiday season so I get that feeling a lot myself.



> Also, during make I get:

Do you mean before my patch, or in a git checkout, or... ?

The "Warning: program compiled against libxml 212 using older 209" is a bit unusual but just means that your distro hasn't recompiled their libxslt package against every minor bugfix release of libxml2.

PR #558, opened 2024-10-17
ping: support IPv4-Mapped-in-IPv6 target addresses

iam-TJ:
Additional feature:

Addresses of the form `::ffff:127.0.0.1` would be treated as IPv6 but should be converted to IPv4 and handled as such.
```
iputils$ build/ping/ping -n -v ::ffff:127.0.0.1
IPv4-Mapped-in-IPv6 address; using IPv4 127.0.0.1
build/ping/ping: sock4.fd: 3 (socktype: SOCK_RAW), sock6.fd: -1 (socktype: 0), hints.ai_family: AF_INET

ai->ai_family: AF_INET, ai->ai_canonname: '127.0.0.1'
64 bytes from 127.0.0.1: icmp_seq=1 ident=45999 ttl=64 time=0.032 ms
64 bytes from 127.0.0.1: icmp_seq=2 ident=45999 ttl=64 time=0.053 ms
64 bytes from 127.0.0.1: icmp_seq=3 ident=45999 ttl=64 time=0.054 ms
^C
3 packets transmitted, 3 received, 0% packet loss, time 2039ms
rtt min/avg/max/mdev = 0.032/0.046/0.054/0.010 ms
```

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -353,6 +353,7 @@ main(int argc, char **argv)
 		.ni.subject_type = -1,
 	};
 	unsigned char buf[sizeof(struct in6_addr)];
+	struct in6_addr a6;
 
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
@@ -618,6 +619,15 @@ main(int argc, char **argv)
 			error(0, 0, _("WARNING: ident 0 => forcing raw socket"));
 
 		hints.ai_socktype = SOCK_RAW;
+	}
+
+	if (inet_pton(AF_INET6, target, &a6)) {
+		if (IN6_IS_ADDR_V4MAPPED(&a6)) {
+			target = strrchr(target, ':') + 1;
+			hints.ai_family = AF_INET;
+			if (rts.opt_verbose)
+				printf("IPv4-Mapped-in-IPv6 address; using IPv4 %s\n", target);
+		}
 	}
 
 	if (hints.ai_family != AF_INET6) {


pevik:
```
```

Thanks a lot for your PR, very nice improvement. It is also nice that whole thing works when forcing a protocol (both `-4` and `-6`).

Could you please reveal your real name? `Tj` is against kernel rules for signing the work we follow?

https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

iam-TJ:
> Could you please reveal your real name? `Tj` is against kernel rules for signing the work we follow?

It is my real name.

pevik:
> > Could you please reveal your real name? `Tj` is against kernel rules for signing the work we follow?
> 
> It is my real name.

If it's your _full_ name, than I'm sorry for asking.

pevik:
Thanks a lot for your fix, merged. I dared to unify verbose reporting (print to stderr, translate, use comma).

aral-matrix:
> Aside: I wrote this because a C++ programmer complained that Debian's ping "doesn't work" in our Matrix support room; instead of complaining the time and energy is best spent on fixing.

@all: you are welcome ;)

aral-matrix:
I just realized, there's a potential bug in this implementation.

```
			target = strrchr(target, ':') + 1;
```

Is the code that subsequently uses ```target``` equipped to deal with an address of the form ```7f00::1```?
In other words: I can see that this fix addresses ```ping ::ffff:127.0.0.1``` - but would it work on ```ping ::ffff:7f00:1``` as well?

pevik:
@iam-TJ @aral-matrix FYI given a discussion in busybox mailing list I'm seriously thinking about reverting this feature instead of using one of your fixes.

https://lists.busybox.net/pipermail/busybox/2024-October/090975.html

> I can understand why anyone would think this translation is normal, because
> the socket API normally unmaps IPv4-mapped IPv6 addresses for compatibility:
> 
> https://www.rfc-editor.org/rfc/rfc3493.html#section-3.7
> >  Applications may use AF_INET6 sockets to open TCP connections to IPv4
> >  nodes, or send UDP packets to IPv4 nodes, by simply encoding the
> >  destination's IPv4 address as an IPv4-mapped IPv6 address, and
> >  passing that address, within a sockaddr_in6 structure, in the
> >  connect() or sendto() call.
> 
> But this is not the case with ping because ping uses IPv6 raw sockets,
> and that seems to stop sendto() from performing the convenience unmapping.
> 
> Sadly, this patch breaks the case where an advanced user wants ping
> to send a packet with an IPv4-mapped IPv6 addresses. For example, you
> might be testing or diagosing firewall rules in that subnet.
> 
> Additionally, I can't see any combination of -4 or -6 option flags that
> could be used as a guard to make the unmapping behaviour expected.

You got Cc in the discussion. Ideally, the discussion would be hold in the busybox mailing list (keep on single place). Could you please subscribe to the busybox mailing list (https://lists.busybox.net/mailman/listinfo/busybox) and post your opinion there?

aral-matrix:
> Could you please subscribe to the busybox mailing list (https://lists.busybox.net/mailman/listinfo/busybox) and post your opinion there?

[x] done :)

auerswal:
> Aside: I wrote this because a C++ programmer complained that Debian's ping "doesn't work" in our Matrix support room; instead of complaining the time and energy is best spent on fixing.

@iam-TJ: Would it be possible for you to expand on this "doesn't work" complaint? What was attempted to be achieved, but did not work as intended? What happened instead?

aral-matrix:
@auerswal : don't get sidetracked :)
https://lists.busybox.net/pipermail/busybox/2024-October/090979.html

yvs2014:
> target = strrchr(target, ':') + 1;

there could be many invalid patterns, like
`0:0:0:0:0:ffff::` or `::ffff:feed:cafe`

aral-matrix:
@yvs2014 : https://github.com/iputils/iputils/pull/558#issuecomment-2424066026 :) already spotted

PR #561, opened 2024-10-19
Handle hex-encoded IPv4-Mapped-in-IPv6 addresses

iam-TJ:
Hexadecimal encoding such as `::ffff:7f:1`  would break IPv4 handling; the hex value needs converting to dotted decimal.

I've also added tests (I think correctly!) for both.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -354,6 +354,7 @@ main(int argc, char **argv)
 	};
 	unsigned char buf[sizeof(struct in6_addr)];
 	struct in6_addr a6;
+	char dotted_decimal[16];
 
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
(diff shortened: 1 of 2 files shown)


aral-matrix:
stop racing me to fixes :p

pevik:
Closing due reverting the original feature, see https://github.com/iputils/iputils/pull/567.

aral-matrix:
Pity - despite not understanding many of the technical details, what I have seen in the discussion was a lot of flailing about and mostly personal reluctance to have this feature - there's not really been a conclusive argument(*) how IPv4 mapped in IPv6 (or hostnames that resolve to such an address form) should be interpreted *other* than as IPv4.

* The one notable exception being network troubleshooting - and I proposed a solution how to leave existing behavior unchanged (by combining -4 and -6 options).

I still believe the feature is good and should eventually be reconsidered, possibly in a different form of implementation.

pevik:
There was a conflict with another issue, which we agreed is more important: https://github.com/iputils/iputils/issues/252 which resulted into 8ed7ffc999e2a541e06ee48faf26a323dfe487c2.

Once glibc is fixed (see https://public-inbox.org/libc-alpha/a26674d3f48af0f67b9abb3e54c1228854cc0bd5@rjp.ie/) 8ed7ffc999e2a541e06ee48faf26a323dfe487c2 can be reverted and that fix `ping -c1 -4 ::ffff:127.0.0.1`. Unfortunately `-6` and default (without switch) will not work. But based on `IN6_IS_ADDR_V4MAPPED()` just setting `hints.ai_family = AF_INET;` could work. The problematic part was modifying the target, I would like to avoid it.

Anyway, feel free to create an issue so that we don't forget on it.

aral-matrix:
Mhh - as you know, my proposed fix https://github.com/iputils/iputils/pull/562 works a bit different than this one here - I propose to do a proper interpretation of the result(s) returned by ```getaddrinfo``` - I think that does not introduce the same conflict with #252 ?

> Anyway, feel free to create an issue so that we don't forget on it.

Since my according pull request is not yet closed, I would propose to keep #562 open instead of creating a new issue, if that's okay with you?

PR #562, opened 2024-10-19
ping ipv4 addresses masked as ipv6 after resolving hostname etc

aral-matrix:
Regarding https://github.com/iputils/iputils/pull/558 and https://github.com/iputils/iputils/pull/561, as the original "complainer" ;), I would like to propose a different solution to support for pinging ipv4 addresses masked in an ipv6 - in this pull request I do so by detecting an ipv4 *after* the call to getaddrinfo, thus allowing hostname resolution and even if that returns an ipv4 masked in ipv6, create a proper ipv4 addrinfo entry to call ping4_run on.

Lars Uffmann <coding23@uffmann.name>

**PS:** Someone smarter than me (see comments ;) should review the values assigned to the addrinfo struct that I believe do not matter for the purpose of the ping tool. Accordingly - remove my comments.
Also, feel free to move my ```SA6_IS_ADDR_V4MAPPED``` macro into the code proper.

PPS: For the verbose output, I did not modify the target string, because it *could* be a hostname in my implementation, so that the verbose output looks like so:
```
build/ping/ping: IPv4-Mapped-in-IPv6 address, using IPv4 ::ffff:7f00:1
```
or like so:
```
build/ping/ping: IPv4-Mapped-in-IPv6 address, using IPv4 hostname
```
927a2ec111791e4a9b220a529893e3cfcd523f3b addresses this with an snprintf of the resolved ipv4.

PPPS: This pull request does not add the tests that @iam-TJ added in https://github.com/iputils/iputils/pull/561

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -93,6 +93,9 @@ ping_func_set_st ping4_func_set = {
 
 #define CASE_TYPE(x) case x: return #x;
 
+const unsigned char IPV4_MASK_IN_IPV6[ 12 ] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff };
+#define SA6_IS_ADDR_V4MAPPED( addr ) ( memcmp( &( (struct sockaddr_in6 * ) addr)->sin6_addr, IPV4_MASK_IN_IPV6, 12 ) == 0 )
+
 static char *str_family(int family)
 {
 	switch (family) {
@@ -353,7 +356,6 @@ main(int argc, char **argv)
 		.ni.subject_type = -1,
 	};
 	unsigned char buf[sizeof(struct in6_addr)];
-	struct in6_addr a6;
 
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
@@ -621,14 +623,6 @@ main(int argc, char **argv)
 		hints.ai_socktype = SOCK_RAW;
 	}
 
-	if (inet_pton(AF_INET6, target, &a6) && IN6_IS_ADDR_V4MAPPED(&a6)) {
-			target = strrchr(target, ':') + 1;
-			hints.ai_family = AF_INET;
-
-			if (rts.opt_verbose)
-				error(0, 0, _("IPv4-Mapped-in-IPv6 address, using IPv4 %s"), target);
-	}
-
 	if (hints.ai_family != AF_INET6) {
 		create_socket(&rts, &sock4, AF_INET, hints.ai_socktype, IPPROTO_ICMP,
 			      hints.ai_family == AF_INET);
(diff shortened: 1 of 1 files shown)


aral-matrix:
and help! how do I sign a commit? :D Sorry for all newbie mistakes, this is my first ever pull request to any official tool.

iam-TJ:
I think this is mis-guided. The usage of IPv4-mapped IPv6 addresses is for **IPv6 only**  applications and has no relation to DNS or hostnames.

In all cases an application doing name resolving will resolve a name via either or both of an A (IPv4) or AAAA (IPv6) resource record. There is no expectation nor suggestion for IPv4-mapped addresses in AAAA resource records.

In the intended use-case of an IPv6-only application that resolves a hostname to an IPv4 address and would thus not be able to make a connection, that application can choose to create an IPv4-mapped address with the prefix `::ffff:/96`. 

The application can then ask the dual-stack capable Operating System to create a connection using the IP4-mapped IPv6 address. The OS will then exchange IPv4 packets between nodes.

iam-TJ:
Some observations to consider:

 1. If an hostname resolves to an IPv4 address and the OS is dual-stack then applications should use` AF_INET`  to attempt the connection.
 2. If the `AF_INET` fails ("network unreachable") then a fallback to IPv4-mapped IPv6 should be attempted.
 3. If the OS is single stack (IPv6 disabled) then nothing is done.
 4. Currently Linux (kernel) cannot disable IPv4 (it hard codes 127.0.0.1) so there is no IPv6 single-stack possibility.
 
So the only case needing to be handled is (2) alongside an explicit IPv4-mapped IPv6 address target on the command-line.

Note: I run IPv6 only networks so "network unreachable" is observed except when a (valid) FQDN is provided which triggers the gateway DNS64/NAT64.

iam-TJ:
**Update**: (2) is wrong. If `AF_INET` fails then an IPv-mapped address is not going to work since it is the same thing (to the OS).

So is there actually anything to do?

I think this is mis-guided. The usage of IPv4-mapped IPv6 addresses is for **IPv6 only**  applications and has no relation to DNS or hostnames.

It is intended to support the use-case where the application is provided with an IPv4 address (via configuration or operator input) and can only create IPv6 sockets. If it is dual-stack aware it has no problem handling the IPv4 connection.

In all cases of an application doing name resolving it will resolve a name via `getaddrinfo()`  most likely via DNS  with either or both of an A (IPv4) or AAAA (IPv6) resource record (possibly also by a `host` file or some other mechanism via `nsswitch`). There is no expectation nor suggestion for IPv4-mapped addresses in AAAA resource records.

In the intended use-case of an IPv6-only application that resolves a hostname to an IPv4 address and would thus not be able to make create an IPv4 socket, that application can choose to create an IPv4-mapped address with the prefix `::ffff:/96` .

The application can then ask the dual-stack capable Operating System to create am IPv6 socket using the IP4-mapped IPv6 address. The OS will exchange IPv4 packets between nodes.

See RFC4291, RFC4038,  RFC5952.

In this case `ping` calls `getaddrinfo()` and then iterates through the address list returned so it will be dealing with the resolved addresses. Since the resolver should not be returning an IPv4-mapped address I cannot see a case to deal with.

aral-matrix:
I am not a network-icist (read: not an expert), I just program stuff using available sockets functionality, so I can't adequately respond to much of your thoughts.
What I see is what "works", and derive from behavior of applications how I believe they should compare.

And what I know is I can put this line in my /etc/hosts:
```
::ffff:7f00:1       hostname
```
and it will be used to resolve ```hostname```. And ssh will correctly use ipv4 to connect to 127.0.0.1. So that's why I think if ping was to support ipv4 unmapping from ipv6, then it would be more useful / more in line with other observed application behavior to do so after getaddrinfo.

iam-TJ:
> ```
> ::ffff:7f00:1       hostname
> ```
The point is this representation is not suggested, recommended, or expected. There is absolutely no case for doing that instead of:
```
127.0.0.1 hostname
```
As I noted, the only expected use-case is an IPv6-only application that has IPv4 addresses given to it by configuration or operator and being able to embed them in an IPv6 address and have the operating system transparently use IPv4 even though the application is using `AF_INET6` sockets.

aral-matrix:
> > ```::ffff:7f00:1       hostname```

> The point is this representation is not suggested, recommended, or expected. There is absolutely no case for doing that instead > of:
> ```127.0.0.1 hostname```

To be fair, the same could be said about David's concern on the busybox mailing list - there's no case for testing firewall rules with ::ffff:7f00:1 instead of 127.0.0.1 :) I respect that he thought it might be a use case though, and the way I develop my software is maximum feature support when it comes to "the intention of the user is absolutely and undeniably clear from the value of the parameter".
Also, there is a very good use case for ipv4 mapped ipv6 addresses:
When transitioning a system to ipv6, this allows to implement ipv6 socket functionality across the board while the network is still running on ipv4 - and eventually switch over to ipv6, and being able to roll back to ipv4.

Hence I think if it behaves well with other tools (as I said, I checked with ssh), and *if* this "unmasking" feature is implemented in ping, it'd be better to do it properly.

PR #563, opened 2024-10-21
ping: Accept IPv4-Mapped-in-IPv6 hostnames

iam-TJ:
Follow-up and less invasive alternative to issue #562 

Do not merge as-is; currently contains pre requisite commits proposed in issue #561
 
Although extremely unusual, hostnames may be resolved to ::ffff:/96 addresses. Do early detection of such hostnames and force IPv4 mode.
    
Also fixes handling when "-4" forces `hints.ai_familty == AF_INET` due to an over enthusiastic patch in issue #253 - `AF_UNSPEC` that only needs to be set when  "-6" is forced. Without this the IPv4-mapped address incorrectly takes the `ping6_run()` path.
    
Tested using /etc/hosts:
```    
::ffff:7f00:1 hostname
```    
without and with "-4", "-6", and with target ipv6-linklocal addresses to ensure no IPv6 link-local regression.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -354,6 +354,7 @@ main(int argc, char **argv)
 	};
 	unsigned char buf[sizeof(struct in6_addr)];
 	struct in6_addr a6;
+	char dotted_decimal[16];
 
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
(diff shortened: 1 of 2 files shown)


yvs2014:
```
-			target = strrchr(target, ':') + 1;
+			target = strchr(target + 2, ':') + 1;
```
it's not too much better, with `0:0:0:0:0:ffff:127.0.0.1` it fails with `Name or service not known`?

p.s.
frankly speaking ip4 mapped addresses has to work with ipv4 ping by default (like with `ping` from `inetutils` for example), I think it will also work with `iputils-ping -4` in the same way (if you revert another 'get-around' introduced in https://github.com/iputils/iputils/commit/d217170)

It's only not totally clear what's correct to do with ip4 mapped addresses if address family is not specified

iam-TJ:
> ```
> -			target = strrchr(target, ':') + 1;
> +			target = strchr(target + 2, ':') + 1;
> ```
> 
> it's not too much better, with `0:0:0:0:0:ffff:127.0.0.1` it fails with `Name or service not known`?

Garrr! I have another alternative where I used the ``ffff` as the key but in all the permutations I was testing to avoid regressions with link-local I seem to have forgotten to use it! I'll do some more testing and then force push the alternative. Thanks for spotting that.

PR #564, opened 2024-10-22
test: ping: avoid deprecation warnings

iam-TJ:
Since meson replaces ':' with '_' in test names any way, as well as writing the deprecation warnings, just do it ourselves to make reviewing build tests easier to scan.

### DIFF
--- a/test/ping/meson.build
+++ b/test/ping/meson.build
@@ -31,6 +31,7 @@ cmd_name = 'ping '
 args = ['-V']
 
 name = 'ping ' + ' '.join(args)
+name = name.replace(':', '_')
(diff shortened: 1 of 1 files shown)


pevik:
This will not work on old Ubuntu Bionic:
test/ping/meson.build:34:0: ERROR: Unknown method "replace" for a string.

Thank you for looking into it, but I'm going to replace testing with proper testsuite anyway.

tacerus:
Submitted https://github.com/iputils/iputils/pull/570 for a solution which works with older Meson.

pevik:
Merged the alternative solution: https://github.com/iputils/iputils/pull/570.

Issue #565, opened 2024-11-13
ping -w -c counts error responses towards exiting


nmeyerhans:
Forwarded from Debian bug [456191](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=456191), which has been open for ages.

From the original request,
```
> thule:~# ping -w 120 -c 1 thule-3
> PING thule-3 (***.37.54) 56(84) bytes of data.
> From thule.uk.xensource.com (***.33.107) icmp_seq=1 Destination Host Unreachable
> From thule.uk.xensource.com (***.33.107) icmp_seq=2 Destination Host Unreachable
> From thule.uk.xensource.com (***.33.107) icmp_seq=3 Destination Host Unreachable
> 
> --- thule-3 ping statistics ---
> 3 packets transmitted, 0 received, +3 errors, 100% packet loss, time 2009ms
> , pipe 3
> thule:~# echo $? 
> 1
> thule:~#
> 
> This is a less than helpful behaviour, particularly with modern
> kernels which send host unreachable when the host is not replying to
> arp.  It makes -w useless in most of the situations where you might
> want it.
```

This behavior is still present, but I'm at a loss to explain why, and in particular why it's implied when the `-w` flag is used, as [seen in ping_common.c](https://github.com/iputils/iputils/blob/aeb124025c9d802ea2a00842dca4c7db66b780bb/ping/ping_common.c#L602-L603):

```
		if (rts->deadline && rts->nerrors)
			break;
```

The behavior has been present since at least version 20020124 (from 2002, the oldest version present in the Debian historical archives).  However, the Debian bug report suggests that the kernel behavior may have changed since the ping behavior was introduced ("particularly with modern kernels which send host unreachable when the host is not replying to arp"), so it's not clear that the current ping behavior was intended when the code was written.

In any case, since the current behavior has been in place for so long, I'm not entirely comfortable changing the default. My inclination is to add a new command-line switch to tell ping to ignore errors when running with a deadline.

PR #566, opened 2024-11-13
ping: add -g flag to ignore errors when deadline is active

nmeyerhans:
Address https://github.com/iputils/iputils/issues/565 by adding a new commandline switch (-g) to avoid exiting if errors are received when deadline (-w) is in effect.

The default behavior is unchanged:
```
$ time ./build/ping/ping -w20 -c1 192.168.1.100
PING 192.168.1.100 (192.168.1.100) 56(84) bytes of data.
From 192.168.1.211 icmp_seq=1 Destination Host Unreachable
From 192.168.1.211 icmp_seq=2 Destination Host Unreachable
From 192.168.1.211 icmp_seq=3 Destination Host Unreachable

--- 192.168.1.100 ping statistics ---
4 packets transmitted, 0 received, +3 errors, 100% packet loss, time 3072ms
pipe 4
./build/ping/ping -w20 -c1 192.168.1.100  0.00s user 0.00s system 0% cpu 3.077 total
```

However, when the `-g` flag is used in combination with `-w`, errors aren't considered when determining whether to exit (either because the deadline has passed or because errors received + responses received >= packets sent):

```
$ time ./build/ping/ping -w20 -c1 192.168.1.100 -g
PING 192.168.1.100 (192.168.1.100) 56(84) bytes of data.
From 192.168.1.211 icmp_seq=1 Destination Host Unreachable
From 192.168.1.211 icmp_seq=2 Destination Host Unreachable
From 192.168.1.211 icmp_seq=3 Destination Host Unreachable
From 192.168.1.211 icmp_seq=4 Destination Host Unreachable
From 192.168.1.211 icmp_seq=5 Destination Host Unreachable
From 192.168.1.211 icmp_seq=6 Destination Host Unreachable
From 192.168.1.211 icmp_seq=7 Destination Host Unreachable
From 192.168.1.211 icmp_seq=8 Destination Host Unreachable
From 192.168.1.211 icmp_seq=9 Destination Host Unreachable
From 192.168.1.211 icmp_seq=10 Destination Host Unreachable
./build/ping/ping: sendmsg: No route to host
From 192.168.1.211 icmp_seq=11 Destination Host Unreachable
From 192.168.1.211 icmp_seq=12 Destination Host Unreachable
From 192.168.1.211 icmp_seq=14 Destination Host Unreachable
From 192.168.1.211 icmp_seq=15 Destination Host Unreachable
From 192.168.1.211 icmp_seq=16 Destination Host Unreachable
From 192.168.1.211 icmp_seq=17 Destination Host Unreachable
From 192.168.1.211 icmp_seq=18 Destination Host Unreachable
From 192.168.1.211 icmp_seq=19 Destination Host Unreachable

--- 192.168.1.100 ping statistics ---
20 packets transmitted, 0 received, +18 errors, 100% packet loss, time 19445ms
pipe 4
./build/ping/ping -w20 -c1 192.168.1.100 -g  0.00s user 0.01s system 0% cpu 20.006 total
``` 

I do think it would be worth considering changing the default behavior, but that's a little more disruptive. Short-circuiting the deadline because a host isn't responding is surprising behavior and it's not clear why we do that. Thoughts?


pevik:
Generally LGTM, I would just consider slightly different usage: instead of `-g` modify `-w` behavior might be better to for your use case _replace_ `-w` with `-g` (less typing).

And thanks for filling up https://github.com/iputils/iputils/issues/565, I haven't noticed that (due issues and PRs having very similar numbers).

It would be great if you could add a description into man page (new switch _have to_ be described in man (ideally mentioning when it's useful instead of `-w` as it's in issue). Also adding little more info from the issue into the commit message would be great (if github dies we lost all the info, that's why I try to put more message into the commit message itself).

pevik (changes requested):
Please add `-g` to the man page. Also, ideally change the meaning to `-g` would be used without `-w`.

pevik:
@nmeyerhans gently ping

yvs2014:
In regard of adding new options (not referring to the gist of changes), can be considered extending '-w' functionality to avoid introducing new options? It could be like
```
  int_option = [get it within range -MAX..MAX];
  if (int_option < 0) { int_option = -int_option; ignores_errors = true; }
```

pevik:
No, new option is IMHO better than incompatible changes in the existing version.

nmeyerhans:
Thinking further about this, I don't want to add yet another flag to ping to control this. IMO we should either fix the current behavior, which I believe to be broken, or ignore it and let people work around it however they see fit.

pevik:
> Thinking further about this, I don't want to add yet another flag to ping to control this. IMO we should either fix the current behavior, which I believe to be broken, or ignore it and let people work around it however they see fit.

Because ping is used in many monitoring tools, I was for adding a new option instead of modifying (again) the functionality. But maybe we should reconsider it. @metan-ucw WDYT?

PR #567, opened 2024-11-15
Revert "ping: support IPv4-Mapped-in-IPv6 target addresses"

pevik:
This reverts commit 8ed7ffc999e2a541e06ee48faf26a323dfe487c2.

This is an incomplete implementation. While there are two attempts (PR#562 [1] PR#562 [2] and PR#563 [3]) to fix it, it's probably not a good idea to support IPv4-Mapped-in-IPv6 target addresses.

There were concerns [4] about supporting this when discussion on BusyBox mailing list about patch bringing this functionality to BusyBox ping implementation:

* ping is a low-level diagnostic tool, maybe it's not good to assume things like this, because IPv4-mapped IPv6 address format is not supported by raw sockets
* it's not clear how -4 -6 switches should behave

Although inetutils ping at least partly supports the feature [5]:

- explicit ipv4 ping:
```
% ping -c1 ::ffff:127.0.0.1
PING ::ffff:127.0.0.1 (127.0.0.1): 56 data bytes
64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0,100 ms
```

- explicit ipv6 ping:
```
% ping6 -c1 ::ffff:127.0.0.1
PING ::ffff:127.0.0.1 (::ffff:127.0.0.1): 56 data bytes
./ping/ping6: sending packet: Network is unreachable
```

Because also fping did not want to support it [6], let's revert the feature.

[1] https://github.com/iputils/iputils/pull/561
[2] https://github.com/iputils/iputils/pull/563
[3] https://github.com/iputils/iputils/pull/562
[4] https://lists.busybox.net/pipermail/busybox/2024-October/090975.html
[5] https://lists.busybox.net/pipermail/busybox/2024-October/090985.html
[6] https://github.com/schweikert/fping/pull/356

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -353,7 +353,6 @@ main(int argc, char **argv)
 		.ni.subject_type = -1,
 	};
 	unsigned char buf[sizeof(struct in6_addr)];
-	struct in6_addr a6;
 
 	/* FIXME: global_rts will be removed in future */
 	global_rts = &rts;
@@ -619,14 +618,6 @@ main(int argc, char **argv)
 			error(0, 0, _("WARNING: ident 0 => forcing raw socket"));
 
 		hints.ai_socktype = SOCK_RAW;
-	}
-
-	if (inet_pton(AF_INET6, target, &a6) && IN6_IS_ADDR_V4MAPPED(&a6)) {
-			target = strrchr(target, ':') + 1;
-			hints.ai_family = AF_INET;
-
-			if (rts.opt_verbose)
-				error(0, 0, _("IPv4-Mapped-in-IPv6 address, using IPv4 %s"), target);
 	}
 
 	if (hints.ai_family != AF_INET6) {
--- a/po/cs.po
+++ b/po/cs.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: iputils\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-10-18 19:07+0200\n"
+"POT-Creation-Date: 2024-11-15 17:30+0100\n"
 "PO-Revision-Date: 2024-10-28 15:00+0000\n"
 "Last-Translator: Pavel Borecki <pavel.borecki@gmail.com>\n"
 "Language-Team: Czech <https://translate.fedoraproject.org/projects/iputils/"
@@ -146,9 +146,9 @@ msgstr "Rozhraní „%s“ nepodporuje protokol ARP\n"
 msgid "WARNING: using default broadcast address.\n"
 msgstr "POZOR: používá se výchozí všesměrová adresa.\n"
 
-#: arping.c:905 arping.c:908 arping.c:911 ping/ping.c:409 ping/ping.c:456
-#: ping/ping.c:508 ping/ping.c:516 ping/ping.c:560 ping/ping.c:563
-#: ping/ping.c:566 ping/ping.c:580 tracepath.c:474 tracepath.c:477
+#: arping.c:905 arping.c:908 arping.c:911 ping/ping.c:408 ping/ping.c:455
+#: ping/ping.c:507 ping/ping.c:515 ping/ping.c:559 ping/ping.c:562
+#: ping/ping.c:565 ping/ping.c:579 tracepath.c:474 tracepath.c:477
 #: tracepath.c:480 tracepath.c:502
 msgid "invalid argument"
 msgstr "neplatný argument"
@@ -290,7 +290,7 @@ msgid "too long scope name"
 msgstr "název síťové oblasti je příliš dlouhý"
 
 #: ping/node_info.c:347 ping/node_info.c:389 ping/ping6_common.c:308
-#: ping/ping.c:1068
+#: ping/ping.c:1059
 msgid "memory allocation failed"
 msgstr "nepodařilo se přidělit paměť"
 
@@ -344,7 +344,7 @@ msgstr ""
 "  subject-name=jméno\n"
 "  subject-fqdn=jméno\n"
 
-#: ping/ping6_common.c:99 ping/ping.c:756
+#: ping/ping6_common.c:99 ping/ping.c:747
 #, c-format
 msgid "unknown iface: %s"
 msgstr "neznámé rozhraní: %s"
@@ -353,7 +353,7 @@ msgstr "neznámé rozhraní: %s"
 msgid "scope discrepancy among the nodes"
 msgstr "nejednotnost síťové oblasti mezi uzly"
 
-#: ping/ping6_common.c:225 ping/ping.c:926
+#: ping/ping6_common.c:225 ping/ping.c:917
 #, c-format
 msgid "Warning: source address might be selected on device other than: %s"
 msgstr "Pozor: zdrojová adresa může být vybrána na zařízením jiném než: %s"
@@ -399,7 +399,7 @@ msgstr "jmenovku toku nelze nastavit"
 msgid "can't send flowinfo"
 msgstr "údaje o toku nelze poslat"
 
-#: ping/ping6_common.c:397 ping/ping.c:1070
+#: ping/ping6_common.c:397 ping/ping.c:1061
 #, c-format
 msgid "PING %s (%s) "
 msgstr "PING %s (%s) "
@@ -409,7 +409,7 @@ msgstr "PING %s (%s) "
 msgid ", flow 0x%05x, "
 msgstr ", tok 0x%05x, "
 
-#: ping/ping6_common.c:404 ping/ping.c:1072
+#: ping/ping6_common.c:404 ping/ping.c:1063
 #, c-format
 msgid "from %s %s: "
 msgstr "od %s %s: "
@@ -554,7 +554,7 @@ msgstr "Zkrácení MLD"
 msgid "unknown icmp type: %u"
 msgstr "neznámý typ ICMP: %u"
 
-#: ping/ping6_common.c:547 ping/ping.c:1489
+#: ping/ping6_common.c:547 ping/ping.c:1480
 msgid "local error"
 msgstr "místní chyba"
 
@@ -563,12 +563,12 @@ msgstr "místní chyba"
 msgid "local error: message too long, mtu: %u"
 msgstr "místní chyba: zpráva je příliš dlouhá, MTU: %u"
 
-#: ping/ping6_common.c:571 ping/ping.c:1525
+#: ping/ping6_common.c:571 ping/ping.c:1516
 #, c-format
 msgid "From %s icmp_seq=%u "
 msgstr "Od %s pořadí=%u "
 
-#: ping/ping6_common.c:677 ping/ping.c:1639
+#: ping/ping6_common.c:677 ping/ping.c:1630
 #, c-format
 msgid " icmp_seq=%u"
 msgstr " pořadí=%u"
@@ -623,16 +623,16 @@ msgstr "; seq=%u;"
 msgid "packet too short: %d bytes"
 msgstr "paket je příliš krátký: %d bajtů"
 
-#: ping/ping6_common.c:923 ping/ping.c:1768
+#: ping/ping6_common.c:923 ping/ping.c:1759
 #, c-format
 msgid "From %s: "
 msgstr "Od %s: "
 
-#: ping/ping6_common.c:964 ping/ping.c:1873
+#: ping/ping6_common.c:964 ping/ping.c:1864
 msgid "WARNING: failed to install socket filter"
 msgstr "POZOR: soketový filtr se nepodařilo nainstalovat"
 
-#: ping/ping.c:103 ping/ping.c:733
+#: ping/ping.c:103 ping/ping.c:724
 #, c-format
 msgid "unknown protocol family: %d"
 msgstr "neznámá generace protokolu: %d"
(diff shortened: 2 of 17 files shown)


nmeyerhans (dismissed):
+1 on reverting this change. As noted on the mailing lists, it's not clear what use cases are enabled with this patch, and if it introduces confusing or unexpected behavior then it'd better to drop it.

If there is a use case that I'm missing, I'd love to learn more about it. If it truly makes sense to support this functionality, we can reconsider it, but we should make sure we're confident in the implementation first.

yvs2014:
> Although inetutils ping at least partly supports the feature [5]:
>     * explicit ipv4 ping:
> ```
> % ping -c1 ::ffff:127.0.0.1
> PING ::ffff:127.0.0.1 (127.0.0.1): 56 data bytes
> 64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0,100 ms
> ```

just to note:
it's libc `getaddrinfo()` feature, not inetutils-ping itself.
It'd work with `iputils-ping -4` in the same way if not to force AF_UNSPEC (user specified -4 was unconditionally rewritten in https://github.com/iputils/iputils/pull/253/commits/e53ec147eb735e07244c2f07fbf97bd9f97b2dd5#)

pevik:
> just to note:
> it's libc `getaddrinfo()` feature, not inetutils-ping itself.

Thanks, I'll later amend the commit message!
@yvs2014 BTW would you also ack reverting this?

> It'd work with `iputils-ping -4` in the same way if not to force AF_UNSPEC (user specified -4 was unconditionally rewritten in [e53ec14](https://github.com/iputils/iputils/commit/e53ec147eb735e07244c2f07fbf97bd9f97b2dd5))

In the end d217170ec9478a1983127428ad94db33c8bc5649 was merged (no code change, but just to clarify which commit is in master). Indeed, reverting d217170ec9478a1983127428ad94db33c8bc5649 (with resolved conflict) on the top of this PR (another revert) really allows to support it.
```
$ ./builddir/ping/ping -4 -c1 ::ffff:127.0.0.1
PING ::ffff:127.0.0.1 (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.042 ms

--- ::ffff:127.0.0.1 ping statistics ---
rtt min/avg/max/mdev = 0.042/0.042/0.042/0.000 ms
```

```diff
diff --git ping/ping.c ping/ping.c
index 22c0dba..d4ce843 100644
--- ping/ping.c
+++ ping/ping.c
@@ -677,13 +677,7 @@ main(int argc, char **argv)
        if (rts.tclass)
                set_socket_option(&sock6, IPPROTO_IPV6, IPV6_TCLASS, &rts.tclass, sizeof(rts.tclass));
 
-       /* getaddrinfo fails to indicate a scopeid when not used in dual-stack mode.
-        * Work around by always using dual-stack name resolution.
-        *
-        * https://github.com/iputils/iputils/issues/252
-        */
        int target_ai_family = hints.ai_family;
-       hints.ai_family = AF_UNSPEC;
 
        if (!strchr(target, '%') && sock6.socktype == SOCK_DGRAM &&
                inet_pton(AF_INET6, target, buf) > 0 &&

```

d217170ec9478a1983127428ad94db33c8bc5649 was tring to fix https://github.com/iputils/iputils/issues/252, which is about pinging IPv6 local scope address without scope ID. Fix d217170ec9478a1983127428ad94db33c8bc5649 enabled it on raw socket, but it does not work on ICMP datagram socket anyway (see c4b8b44a0403f52efc1c76adca9ee2181688ea9c). Maybe we could revert d217170ec9478a1983127428ad94db33c8bc5649 to get IPv4-Mapped-in-IPv6 with `-4` working. 
@rpigott @nmeyerhans @yvs2014 WDYT?

pevik:
@nmeyerhans @yvs2014 @rpigott Could you please review?

Once you review, please add your `Reviewed-by:` (fully reviewed) or at least `Acked-by:` (somehow agree with the change) tag.

rpigott:
I don't understand this change?

I don't really see why we care about IPv4-mapped ipv6 support. It all gets emitted as ipv4 anyway. Honestly, ping6 failing when given an ipv4 mapped ipv6 addr seems like the correct behavior to me. Support for scope id seems much more valuable imo. 

> https://github.com/iputils/iputils/commit/d217170ec9478a1983127428ad94db33c8bc5649 was tring to fix https://github.com/iputils/iputils/issues/252, which is about pinging IPv6 local scope address without scope ID. Fix https://github.com/iputils/iputils/commit/d217170ec9478a1983127428ad94db33c8bc5649 enabled it on raw socket, but it does not work on ICMP datagram socket anyway (see https://github.com/iputils/iputils/commit/c4b8b44a0403f52efc1c76adca9ee2181688ea9c).

My patch wasn't intended to address pinging explicit link-local ipv6 addrs without a scope id (these are malformed), but _hostnames_ that resolve to link-local ipv6 addresses — these are not normally found in DNS. Though I only called out sd-resolved synthetic responses when I submitted the patch, it's also possible to get ipv6 with scope id from getaddrinfo if the local resolver stub utilizes mdns, either with sd-resolved or avahi. In this case link-local addresses may be returned. If the patch is reverted, I expect `ping -6 host.local` will sometimes fail when it could otherwise succeed when it resolves to a link local address.

We should probably change it to use AF_INET when -4 is specified, though. It's only AF_INET6 that is broken in libc.

yvs2014:
+1 to revert both commits ipv4-mapped-in-ipv6 and scopeid-link-local and to match other ping implementations

On ipv4-mapped-in-ipv6:
There's only doubt about implicit behavior (-4/-6 is not specified) .  Should-or-shouldn't it behave like 'ping -4' (i.e. convert mapped address to ipv4)

On scopeid-link-local:
If it's needed in definite cases, how to condition it to those ones only (to some libc or their versions) to not break other code

I think both these cases fall into a broader category:
if there's lack of some extra features (in libc or whatsoever), should it be implemented in ping itself?

nmeyerhans:
> I don't really see why we care about IPv4-mapped ipv6 support. It all gets emitted as ipv4 anyway. Honestly, ping6 failing when given an ipv4 mapped ipv6 addr seems like the correct behavior to me. Support for scope id seems much more valuable imo.

+1; if `getaddrinfo()` comes back with a link-scoped address, we should use that.  Doing so is more important than gracefully handling IPv4-mapped IPv6 addresses.

> My patch wasn't intended to address pinging explicit link-local ipv6 addrs without a scope id (these are malformed), but hostnames that resolve to link-local ipv6 addresses — these are not normally found in DNS. Though I only called out sd-resolved synthetic responses when I submitted the patch, it's also possible to get ipv6 with scope id from getaddrinfo if the local resolver stub utilizes mdns, either with sd-resolved or avahi. In this case link-local addresses may be returned. If the patch is reverted, I expect ping -6 host.local will sometimes fail when it could otherwise succeed when it resolves to a link local address.

With the change proposed in this PR, we get the following behavior:

```
~/iputils$ ./build/ping/ping -6 _gateway
PING _gateway (fe80::cc43:72ff:fe53:7222) 56 data bytes
^C
--- _gateway ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 2035ms

~/iputils$ ./build/ping/ping -4 _gateway
PING _gateway (172.18.0.1) 56(84) bytes of data.
64 bytes from 172.18.0.1: icmp_seq=1 ttl=64 time=0.122 ms
64 bytes from 172.18.0.1: icmp_seq=2 ttl=64 time=0.153 ms
64 bytes from 172.18.0.1: icmp_seq=3 ttl=64 time=0.144 ms
^C
--- _gateway ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2002ms
rtt min/avg/max/mdev = 0.122/0.139/0.153/0.013 ms
~/iputils$ ./build/ping/ping _gateway
PING _gateway (fe80::cc43:72ff:fe53:7222%enp0s2) 56 data bytes
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=1 ttl=64 time=0.122 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=2 ttl=64 time=0.156 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=3 ttl=64 time=0.166 ms
^C
--- _gateway ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 0.122/0.148/0.166/0.018 ms
```

That's a change from the behavior that's been in place since d217170ec9478a1983127428ad94db33c8bc5649 was committed in early 2020:

```
~/iputils$ .//build/ping/ping -6 _gateway
PING _gateway (fe80::cc43:72ff:fe53:7222%enp0s2) 56 data bytes
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=1 ttl=64 time=0.135 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=2 ttl=64 time=0.164 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=3 ttl=64 time=0.156 ms
^C
--- _gateway ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 0.135/0.151/0.164/0.012 ms
~/iputils$ ./build/ping/ping -4 _gateway
PING _gateway (172.18.0.1) 56(84) bytes of data.
64 bytes from 172.18.0.1: icmp_seq=1 ttl=64 time=0.119 ms
64 bytes from 172.18.0.1: icmp_seq=2 ttl=64 time=0.139 ms
64 bytes from 172.18.0.1: icmp_seq=3 ttl=64 time=0.163 ms
^C
--- _gateway ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 0.119/0.140/0.163/0.018 ms
~/iputils$ ./build/ping/ping _gateway
PING _gateway (fe80::cc43:72ff:fe53:7222%enp0s2) 56 data bytes
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=1 ttl=64 time=0.108 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=2 ttl=64 time=0.169 ms
64 bytes from fe80::cc43:72ff:fe53:7222%enp0s2: icmp_seq=3 ttl=64 time=0.176 ms
^C
--- _gateway ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 0.108/0.151/0.176/0.030 ms
```

(Note that that's using glibc; musl seems to have other issues)

I would consider that a regression.

I would like to hold off on merging any changes related to this functionality until we're sure about what changes we're making and what we actually want the behavior to be.

pevik:
Thanks for your comments. Sure, I'd like to find some general agreement before merging anything.

> ~/iputils$ ./build/ping/ping -6 _gateway
> PING _gateway (fe80::cc43:72ff:fe53:7222) 56 data bytes
> ^C
> --- _gateway ping statistics ---
> 3 packets transmitted, 0 received, 100% packet loss, time 2035ms

@nmeyerhans Could you please update the examples with `-v` to see which socket is used? I your examples are using raw socket.

@rpigott Can you please clarify more the original problem? https://github.com/iputils/iputils/commit/d217170ec9478a1983127428ad94db33c8bc5649 mentions `systemd-resolved` and `ping -6 _gateway`, but who sets `_gateway` hostname on link scope ID? Shouldn't be gateway on a global scope (which does not need scope ID)? I suppose it's set in `/etc/hosts`, but posting relevant config (`/etc/hosts` and `ip addr`) would help. Also testing on other implementations helps (either to fix iputils ping if we are wrong or report a bug to them if others are wrong).

As I wrote above, trying to ping on IPv6 link scope without scope ID is not a good idea and it will IMHO work only on raw socket (it never worked on ICMP datagram socket, this might be fixable in kernel, but nobody cared because scope ID is just expected). Shouldn't be the use case or other software be fixed? Aren't we just workarounding some bug / wrong use case?

rpigott:
`_gateway` is a hostname for which sd-resolved gives a synthetic response: the address of the default gateway, which is commonly a link-local address in ipv6:

```
$ ip route get fibmatch 2001:db8::1
default via fe80::daec:5eff:fede:cd3c dev enp7s0 proto static metric 256 onlink pref medium
```

When nss is configured with the nss-resolve module, `getaddrinfo` can resolve this name including the correct scope id, but only when requesting AF_UNSPEC, not AF_INET6. This is a limitation of the nss interface and not a flaw in nss-resolve — AF_INET6 is broken by design in getaddrinfo+glibc. This affects all users of getaddrinfo with ipv6. E.g. with `getent`:

```
$ getent -s resolve ahostsv6 _gateway | grep -m1 fe80
fe80::daec:5eff:fede:cd3c STREAM _gateway
$ getent -s resolve ahosts _gateway | grep -m1 fe80
fe80::daec:5eff:fede:cd3c%2 STREAM 
```
Notice that the scope id is returned only with AF_UNSPEC.

But as I stated above, the synthetic name `_gateway` is not the only name which may resolve to a link-local ipv6 address, as link-local addresses are widely used in mDNS which in turn is used with nss-resolve or nss-mdns in applications that call `getaddrinfo`. E.g.:

```
$ getent -s resolve ahostsv6 $HOST.local | grep -m1 fe80
fe80::7686:e2ff:fe24:4f96 STREAM 
$ getent -s resolve ahosts $HOST.local | grep -m1 fe80
fe80::7686:e2ff:fe24:4f96%2 STREAM 
```

If the only answer to a query for $HOST.local is a link-local address (which may be the common case in ipv6 networks without a GUA prefix), a `ping -6 $HOST.local` that queries AF_INET6 will fail but one that queries AF_UNSPEC will succeed.

yvs2014:
I suppose the case with setting AF_UNSPEC for ping6 could be fixed with checking out two conditions if 
- explicitly requested -6 family
- in6 address is link-local

And if it's considered broken in libc
> It's only AF_INET6 that is broken in libc
> AF_INET6 is broken by design in getaddrinfo+glibc

what were upstream's (glibc?) objections against to fix it?

rpigott:
I never submitted any bug report to glibc. I first discovered this issue [here](https://bbs.archlinux.org/viewtopic.php?id=252459).

I reckon the only possible fix is new ABI and an update to existing nss modules which admittedly was less appealing to me than a simple fix in ping. If you know where to report or who to contact I suppose we can see if its feasible to fix in glibc.

pevik:
> If you know where to report or who to contact I suppose we can see if its feasible to fix in glibc.

I guess sending a report/question to [glibc mailing list](https://sourceware.org/mailman/listinfo/libc-alpha) would be quickest than a file bug to glibc bugzilla. General instructions are in: https://sourceware.org/glibc/wiki/FilingBugs.

nmeyerhans:
> @nmeyerhans Could you please update the examples with `-v` to see which socket is used? I your examples are using raw socket.

Done. I replaced the existing runs with output generated with `-v`, showing non-RAW ICMP sockets, and I also added output showing the behavior with RAW sockets.

It seems that the specific case that's definitively broken by this PR is when `-6` (`AF_INET6`) is specified and a non-RAW socket is used.  However, RAW socket behavior depends on the routing configuration to handle link-local addresses without a scope ID, so hosts with multiple `fe80::/64` routes will see failures there too.

rpigott:
Pavel's comment in https://sourceware.org/bugzilla/show_bug.cgi?id=14413 is indeed this exact problem, and correctly points out that neither nss_gethostbyname3_r/nss_gethostbyname4_r are adequate to resolve link-local addresses.

I can send a mail, but I don't expect any movement when it has been known for 10+ years.

rpigott:
https://sourceware.org/pipermail/libc-alpha/2024-November/161566.html

yvs2014:
Spotted at testing ping6 via icmp socket there's two distinct cases:
- one of them is related to resolv (like those cases with[out] scopeid _gateway)
- the second one is ping6 pure-link-local-ipaddress without adding %scope

fe80 addresses are successfully pinged on some Linuses, and failed on other ones:
Outgoing packets go via loopback when its' successful, and when it's failed - there's EINVAL (maybe directly from kernel's net/ipv6/ping.c) error returned by sendto() with nothing emitted.

In the first case (resolv) if conditionally apply d217170 commit, there's no collisions with ipv4-mapped addresses even they both can be used with setting AF_UNSPEC.
(tested with https://github.com/yvs2014/iputils/commit/cd3d7fcd4db6cc704c012c1c1c019781a2e74759, suppose it can be written better but for tests it was enough)

pevik:
I guess first two commits are OK. Maybe we should merge them and handling the remaining problem separately (I would create a new ticket with summary of the problem. Also, getting folks from other project would help to find the best solution.) If you agree, please review them and put your `Reviewed-by:` tag here.

I'll try next week to replace the third one with cleanup of your suggestion. Code definitely needs cleanup, but first we should agree that exception based on `IN6_IS_ADDR_LINKLOCAL()` is a good idea (IMHO yes) @nmeyerhans @rpigott WDYT?

BTW from the very quick look I would move `ping6_unspec()` into `ping6_run()` (in ping/ping6_common.c) or at least have this function in ping/ping6_common.c.

Also, we have `IN6_IS_ADDR_LINKLOCAL(buf) || IN6_IS_ADDR_MC_LINKLOCAL(buf)` related warning above in ping/ping.c. 1) if `IN6_IS_ADDR_MC_LINKLOCAL()` is correct (not sure how to test), it should be in the workaround as well. 2) These two parts of code (warning and workaround) are related and should be somehow connected.

pevik:
Link to glibc discussion: https://public-inbox.org/libc-alpha/64440483-ec96-4fd3-8bfd-e63f88f5bfbc@redhat.com/T/#t.

pevik:
OK, I reverted the last commit which is controversial. Hopefully there will be glibc fix, then we could finally revert d217170ec9478a1983127428ad94db33c8bc5649.
And I try to investigate why it's not working on musl.

rpigott:
Reviewed-by: Ronan Pigott <ronan@rjp.ie>

aral-matrix:
> Although inetutils ping at least partly supports the feature [5]:
> 
>     * explicit ipv4 ping:
> 
> 
> ```
> % ping -c1 ::ffff:127.0.0.1
> PING ::ffff:127.0.0.1 (127.0.0.1): 56 data bytes
> 64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0,100 ms
> ```

Which version introduced this behavior, if I may ask? On my debian stable, this functionality is not attempting to ping the ipv4 destination.

yvs2014:
> Which version introduced this behavior, if I may ask?
>
  If I'm not mistaken with looking up, it was introduced in glibc-2.8 (year ~2007 ~2008) https://github.com/bminor/glibc/blame/master/nss/getaddrinfo.c#L2544

> On my debian stable, this functionality is not attempting to ping the ipv4 destination.
>
  Probably it's not busybox ping nor inetutils one, more likely you've tried to iputils-ping with https://github.com/iputils/iputils/commit/d217170ec9478a1983127428ad94db33c8bc5649 which prevents to use this libc feature.

aral-matrix:
> Probably it's not busybox ping nor inetutils one, more likely you've tried iputils-ping with [d217170](https://github.com/iputils/iputils/commit/d217170ec9478a1983127428ad94db33c8bc5649) which prevents to use this libc feature.

Indeed, I am using iputils - to be honest, I never understood why @pevik asked TJ and myself to explain the issue on a busybox mailing list, nor what busybox actually is ^^ I just rolled with it and assumed this github here would eventually flow back into iputils ping.

Thanks for the clarification - I didn't know there were such divergent versions of ping in existence.

PR #568, opened 2024-11-18
Fix ping man page syntax error

NDagestad:
Hi, 

The ping manpage got an addition in september to document the IPUTILS_PING_PTR_LOOKUP environment variable. The new section that was added does not put it's text in a <para> tag, this caused the next section not to be treated as such.

(It was also indented with tabs instead of spaces like the rest of the document so I changed that as well)

### DIFF
--- a/doc/ping.xml
+++ b/doc/ping.xml
@@ -790,9 +790,9 @@ xml:id="man.ping">
     <info>
       <title>ENVIRONMENT</title>
     </info>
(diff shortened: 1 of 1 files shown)


pevik:
Thanks a lot, merged (with added details into the commit message).

pevik:
@nmeyerhans This fix might be worth to backport to Debian (e8c9d6d7aebd16b0617f00d128cbcd24c9a1094d).

PR #569, opened 2024-11-27
Fix Coverity: Validate MTU to Prevent Tainted Data and Arguments in

lucckb:
tracepath.c

Addressed two critical vulnerabilities identified by Coverity:

1. Tainted Data in sendto:
   - Issue: The expression `ctl->mtu - ctl->overhead` was passed directly to `sendto` without proper validation, potentially allowing buffer overflows.
   - Fix: Added a validation check to ensure `ctl->mtu` exceeds `ctl->overhead` and does not exceed a predefined `MAX_MTU` before invoking `sendto`.

2. Tainted Argument in `recv`:
   - Issue: The variable `ctl->mtu`, derived from user input via the `-l` option, was used directly as the buffer size in `recv`, risking excessive memory allocation and buffer overflows.
   - Fix: Introduced a conditional check based on the address family (`AF_INET` or `AF_INET6`) to ensure `ctl->mtu` does not exceed the maximum allowed MTU (`DEFAULT_MTU_IPV4` or `DEFAULT_MTU_IPV6`) before calling `recv`.

These modifications enhance the security posture of `tracepath.c` by mitigating potential buffer overflow vulnerabilities and improving overall error handling mechanisms.

[ type description here; PLEASE REMOVE THIS LINE AND THE LINES BELOW BEFORE SUBMITTING THIS PULL REQUEST ]

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -370,6 +370,9 @@ static int probe_ttl(struct run_state *const ctl)
 		clock_gettime(CLOCK_MONOTONIC, &hdr->ts);
 		ctl->his[ctl->hisptr].hops = ctl->ttl;
 		ctl->his[ctl->hisptr].sendtime = hdr->ts;
+		if (ctl->mtu <= ctl->overhead) {
+			error(2, 0, "MTU <= OVERHEAD");
+		}
 		if (sendto(ctl->socket_fd, ctl->pktbuf, ctl->mtu - ctl->overhead, 0,
 			   (struct sockaddr *)&ctl->target, ctl->targetlen) > 0)
 			break;
@@ -384,6 +387,12 @@ static int probe_ttl(struct run_state *const ctl)
 
 	if (i < MAX_PROBES) {
 		data_wait(ctl);
+		const int allow_max =
+			(ctl->ai->ai_family==AF_INET6)?(DEFAULT_MTU_IPV6):(DEFAULT_MTU_IPV4);
+		if (ctl->mtu > allow_max) {
+			error(2, 0, "MTU <= MAX_MTU");
+			return 0;
+		}
 		if (recv(ctl->socket_fd, ctl->pktbuf, ctl->mtu, MSG_DONTWAIT) > 0) {
 			printf(_("%2d?: reply received 8)\n"), ctl->ttl);
 			return 0;


pevik:
I guess you scan with https://scan.coverity.com. IMHO coverity is wrong.

> Issue: The expression ctl->mtu - ctl->overhead was passed directly to sendto without proper validation, potentially allowing buffer overflows.

This is not true: `if (ctl.mtu <= ctl.overhead)` check is already there (see `goto pktlen_error`).

> Issue: The variable ctl->mtu, derived from user input via the -l option, was used directly as the buffer size in recv, risking excessive memory allocation and buffer overflows.
> Fix: Introduced a conditional check based on the address family (AF_INET or AF_INET6) to ensure ctl->mtu does not exceed the maximum allowed MTU (DEFAULT_MTU_IPV4 or DEFAULT_MTU_IPV6) before calling recv.

MTU is limited by the above thus nothing bad can happen. And running `-l 65558` (i.e. `DEFAULT_MTU_IPV4 + DEFAULT_OVERHEAD_IPV4 - 1`) allows sending correctly. Why? Because first it goes through localhost which usually has MTU which is much lower (e.g. physical device has usually 1500 as the default, max can be e.g. 2300) and `-l` value is replaced to this much lower value.

https://github.com/iputils/iputils/blob/5f45734a17cc6f7e82999db75c60d1bec330456e/tracepath.c#L305-L307

tracepath.c
```diff
@@ -384,6 +387,12 @@ static int probe_ttl(struct run_state *const ctl)
 
 	if (i < MAX_PROBES) {
 		data_wait(ctl);
+		const int allow_max =
+			(ctl->ai->ai_family==AF_INET6)?(DEFAULT_MTU_IPV6):(DEFAULT_MTU_IPV4);
```
pevik, line 391: nit: Brackets aren't needed. I would also use the constants directly
```
if (ctl->mtu > (ctl->ai->ai_family==AF_INET6 ? DEFAULT_MTU_IPV6 : DEFAULT_MTU_IPV4)
    error(2, 0, "MTU <= MAX_MTU");
```
Return isn't needed.

nmeyerhans:
> I guess you scan with https://scan.coverity.com. IMHO coverity is wrong.
> 
> > Issue: The expression ctl->mtu - ctl->overhead was passed directly to sendto without proper validation, potentially allowing buffer overflows.
> 
> This is not true: `if (ctl.mtu <= ctl.overhead)` check is already there (see `goto pktlen_error`).

While reviewing this, I noticed one minor issue in the current code.  We [attempt to validate](https://github.com/iputils/iputils/blob/5f45734a17cc6f7e82999db75c60d1bec330456e/tracepath.c#L473-L474) the `-l` value in the `getopt()` loop:

```
		case 'l':
			ctl.mtu = strtol_or_err(optarg, _("invalid argument"), ctl.overhead, INT_MAX);
```

However, we haven't set `ctl.overhead` yet, so it's still set to its initial value of zero.  Because of the subsequent check (`goto pktlen_error`) this is not a practical problem, but it is somewhat confusing. IMO, because we can't know the final value of `ctl.overhead` until after `getaddrinfo()`, which we can't run until after we've processed the command line, we should change the `-l` validation to validate that `0 < value < INT_MAX`.  That's effectively what it's already doing and it reduces the potential for confusion.

Otherwise, I think I'm in agreement that the coverity tests are alerting on a non-issue. It's trivial to demonstrate this by passing extreme values (very small or very large) via `-l` and observing the behavior.

PR #570, opened 2024-11-30
tests: replace colons with underscores

tacerus:
In test names, replace colons in IPv6 addresses with underscores to resolve

```
DEPRECATION: ":" is not allowed in test name "ping -c1 ::1", it has been replaced with "_"
```

which was introduced by meson.git commit 55cf399ff8b9.

---

I alternatively have a cleaner patch which uses `.replace()`, however it requires bumping Meson to `>=0.58` - let me know if you'd prefer this.


yvs2014:
> unfortunately, the new version would also introduce a new deprecation warning
>
  I think it doesn't need to replace some deprecation warnings with other ones unless it's additional functionality (true for test/ping/meson.build as well as for main meson.build. As an example there's compat>=0.40 test version without warnings and future notices: https://github.com/yvs2014/iputils/blob/master/test/ping/meson.build and https://github.com/yvs2014/iputils/blob/master/meson.build)

nmeyerhans:
> > I think it doesn't need to replace some deprecation warnings with other ones unless it's additional functionality
> 
> Agree. Older meson is here to support still used Ubuntu bionic, which is still EOL, but the repos are still live. I guess next year (or after next release) will be the time to drop the support which allows to increase the version.

How important is it really that we can build iputils master on Ubuntu 18.04?  It contains an earlier release of iputils and always will, so there's no impact to the distribution itself if we break support. I'm not sure there's a lot of value in supporting >3 LTS versions of any distro.

Ubuntu 18.04 will be supported in some form by Canonical until [at least 2028](https://ubuntu.com/blog/18-04-end-of-standard-support)

yvs2014:
> I'm not sure there's a lot of value in supporting >3 LTS versions of any distro.
>
  In this case it needs more agreements which distributions are taken for references as well as how many,
  speaking about setting minimal 0.58 meson:
  Debian10 has meson-0.49, Ubuntu20 has meson-0.53 (if I'm not mistaken)
  
  On other hand I don't see anything bad in adding `version_compare` conditions to get rid of deprecation warnings and support older versions for a bit longer.

pevik:
> How important is it really that we can build iputils master on Ubuntu 18.04?

It's not important, but OTOH why to deliberately break it when we don't need to raise new version? Sure, once we really need a newer meson support for Bionic will go away.

Issue #496, opened 2023-10-25
Add JSON output


pevik:
Add JSON output for all tools or at least for `ping`. Using probably [libjson-c](https://github.com/json-c/json-c) or write it from scratch (like [iproute2](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/) project does).

PR #571, opened 2024-12-01
JSON output for ping

tacerus:
**Commit:**

Introduce a "-j" argument to have `ping` optionally return machine readable
JSON objects instead of the default text. The output with "-j" can be
piped into JSON parsing tools such as `jq`.
To enable the functionality, a dependency on libjson-c is introduced, however
care was taken for the application to continue operating normally if built
without this dependency - only "-j" will not be available in this case.

TODO: all errors and warnings to JSON
TODO: optional output options (such as verbose) to JSON
TODO: tests (good would be to pipe the -j output into a separate tool
^ ideally, _all_ output variations should be valid JSON if combined
with -j
which then validates if output is valid and if the correct keys are
present, however I could so far not find how to achieve this with
meson.build - will it require a wrapper script which find_program() can
pick up?)

Signed-off-by: Georg Pfuetzenreuter <mail@georg-pfuetzenreuter.net>

---

**Sample output** of a simple run:

```
$ builddir/ping/ping -n -c2 -j4 localhost
{ "bytes": 64, "host": "127.0.0.1", "seq": 1, "ttl": 64, "time": "0.023" }
{ "bytes": 64, "host": "127.0.0.1", "seq": 2, "ttl": 64, "time": "0.019" }
{ "host": "localhost", "transmitted": 2, "received": 2, "duplicates": 0, "corrupted": 0, "errors": 0, "loss": 0, "time": 1022, "rtt": { "min": "0.019", "avg": "0.021", "max": "0.023", "mdev": "0.002" } }

$ builddir/ping/ping -n -c3 -s 128 -j ipv6-localhost
{ "bytes": 136, "host": "::1", "seq": 1, "ttl": 64, "time": "0.020" }
{ "bytes": 136, "host": "::1", "seq": 2, "ttl": 64, "time": "0.022" }
{ "bytes": 136, "host": "::1", "seq": 3, "ttl": 64, "time": "0.029" }
{ "host": "ipv6-localhost", "transmitted": 3, "received": 3, "duplicates": 0, "corrupted": 0, "errors": 0, "loss": 0, "time": 2029, "rtt": { "min": "0.020", "avg": "0.023", "max": "0.029", "mdev": "0.003" } }
```

---

**Relates** to https://github.com/iputils/iputils/issues/496.


pevik:
Thanks for addressing https://github.com/iputils/iputils/issues/496.

By default it fails:
```
../meson.build:112:15: ERROR: C shared or static library 'json-c' not found
```
It would be nice if json support was optional, e.g. `-DUSE_JSON=false` was the default.

But best would be to not require `json-c` for printing json (simpler than parse json).

iproute2 uses it's own code which should be enough to reuse:
https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/json_print.c
https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/json_print_math.c
https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/json_writer.c

Other option would be to reuse writing parts of [ujson](https://github.com/metan-ucw/ujson), see vendoring whole lib in [LTP](https://github.com/linux-test-project/ltp/commit/c50a0e0751c8c274a93fb5dbb9f7a46be8051e76) (if iproute2 parts turn out to be non usable).

pevik:
Valid JSON (see https://jsonlint.com/) should be wrapped output into `[ ]` and have comma after items:

```
[ { "bytes": 64, "host": "127.0.0.1", "seq": 1, "ttl": 64, "time": "0.023" },
{ "bytes": 64, "host": "127.0.0.1", "seq": 1, "ttl": 64, "time": "0.023" },
{ "bytes": 64, "host": "127.0.0.1", "seq": 2, "ttl": 64, "time": "0.019" },
{ "host": "localhost", "transmitted": 2, "received": 2, "duplicates": 0, "corrupted": 0, "errors": 0, "loss": 0, "time": 1022, "rtt": { "min": "0.019", "avg": "0.021", "max": "0.023", "mdev": "0.002" } } ]

```

Also, I was thinking with putting the output into single line (there is [`jq`](https://github.com/jqlang/jq) for formatting the final JSON), but missing new line could cause problems for some terminals/terminal emulators with longer ping run).

```
[ { "bytes": 64, "host": "127.0.0.1", "seq": 1, "ttl": 64, "time": "0.023" }, { "bytes": 64, "host": "127.0.0.1", "seq": 1, "ttl": 64, "time": "0.023" }, { "bytes": 64, "host": "127.0.0.1", "seq": 2, "ttl": 64, "time": "0.019" }, { "host": "localhost", "transmitted": 2, "received": 2, "duplicates": 0, "corrupted": 0, "errors": 0, "loss": 0, "time": 1022, "rtt": { "min": "0.019", "avg": "0.021", "max": "0.023", "mdev": "0.002" } } ]
```

I wonder if put into JSON also errors:
```
$ ./builddir/ping/ping -j nonexisting
./builddir/ping/ping: nonexisting: Name or service not known
```
Maybe not needed.

Also, I would split CI changes into its own commit (easier to follow the changes).

tacerus:
Thanks for the input and links!

> It would be nice if json support was optional, e.g. -DUSE_JSON=false was the default.

Sure, I can make it opt-in.

> iproute2 uses it's own code which should be enough to reuse

It looked a bit involved to port, and I did not want to introduce any "fragile" JSON. But I can try it.
It should be noted that json_print from iproute2 is licensed under the GPL only, but maybe json_write will suffice.

> Other option would be to reuse writing parts of ujson

That's a new one, will check.

> Valid JSON (see https://jsonlint.com/) should be wrapped output into [ ] and have comma after items:

`[]` makes for an array. I wanted a stream of individual JSON map objects which can be live read when piped into a tool like `jq`. If we add all maps to an array, I can only print at the very end, and there will not be a live stream (otherwise the stream would be broken JSON until it completed).

> I wonder if put into JSON also errors.

Yes, that was one of the TODOs. It is not "urgent", as the errors are in stderr, and hence do not break a stream into tools like `jq`, but it would be nice to have both stderr and stdout valid JSON.

tacerus:
> Also, I would split CI changes into its own commit (easier to follow the changes).

Sure!

pevik:
> It looked a bit involved to port, and I did not want to introduce any "fragile" JSON. But I can try it.

Thank you. Any vendored code helps to use the feature (no new dependency).

> It should be noted that json_print from iproute2 is licensed under the GPL only, but maybe json_write will suffice.

Hopefully we can combine GPLv2 and BSD-3-Clause (ping original), at least that's what I understand from:
https://www.gnu.org/licenses/license-list.en.html#ModifiedBSD
(3-clause BSD license is compatible with GPL)

https://opensource.stackexchange.com/questions/7180/can-i-combine-gplv2-and-bsd-3-clause-code-and-keep-the-bsd-permissions

BTW [`json-c`](https://github.com/json-c/json-c/wiki) is MIT License, but because it's a library, there would be a linking exception.

We already link with [`libcap`](https://git.kernel.org/pub/scm/libs/libcap/libcap.git/), but it's dual license [GPLv2 and BSD-3-Clause](https://git.kernel.org/pub/scm/libs/libcap/libcap.git/tree/License), thus no problem to use it for BSD-3-Clause.

tacerus:
Thanks for looking into it, I wasn't sure if it's compatible.

Regarding iproute2, I briefly tested implementing their JSON logic, and have it partially working, but it is not quite ideal - unless I am missing something in my port, it prints the JSON "piece by piece", which is fine for a tool which provides a single load of output, but troubles me with ping, where our output is more dynamic. It requires care to always get the start and end of an object right (as `ping` can run for an arbitrary amount of time, and with different output in between, it is easy to leave behind an invalid JSON if printed character by character without knowing what might come prior or next).

For example during my testing, I accidentally produced this invalid output:

```
$ builddir/ping/ping -n localhost -4j -c3
{"bytes":64,{"host":"127.0.0.1",{"seq":1,{"ttl":64,{"time":"0.017"}
,{"bytes":64,{"host":"127.0.0.1",{"seq":2,{"ttl":64,{"time":"0.016"}
,{"bytes":64,{"host":"127.0.0.1",{"seq":3,{"ttl":64,{"time":"0.028"}
```

With the current implementation in the patch using libjson-c, each object is constructed in memory, and printed only at the end, as a whole valid line.

From a short glance, it seems ujson is unfortunately using a similar construction method as iproute2.

But I understand the preference not to introduce another dependency.

tacerus:
I now refactored it to use custom logic. It is a bit more fragile than the libjson-c approach, but hopefully less fragile than the character-by-character printing solution from iproute2. Still need to do more testing.

ping/ping_common.c
```diff
@@ -806,35 +813,51 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
 		uint8_t *cp, *dp;
 
 		print_timestamp(rts);
-		printf(_("%d bytes from %s:"), cc, from);
+		if (!rts->opt_json)
+			printf(_("%d bytes from %s:"), cc, from);
```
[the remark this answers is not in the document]
tacerus: Makes sense! Will check.

ping/ping_common.c
```diff
@@ -430,8 +430,13 @@ int pinger(struct ping_rts *rts, ping_func_set_st *fset, socket_st *sock)
 	if (i == 0 && !rts->opt_quiet) {
 		if (rts->opt_flood)
 			write_stdout("E", 1);
-		else
-			error(0, errno, "sendmsg");
+		else {
+			if(!rts->opt_json)
+				error(0, errno, "sendmsg");
+
+			error_json(rts, 0, "sendmsg", strerror(errno), PING_JSON_NUL, "");
```
pevik, line 437: As I wrote I'm not sure if error msg is really important to have in JSON. But if you prepared it, it'd be great to be  `if (!rts->opt_json)` was handled internally. I.e. have custom macro / inline function, which either call `error()` or `error_json()` based on `if (!rts->opt_json)`.

pevik:
Frankly speaking when looking into ping code again, ping code would deserve complete rewrite. Then implementing features like this would be easy. Beware the first version of the code is from 1980+ and modifications did not help much to cleanup.

tacerus:
Yeah, there is also some ping/ping6 duplication that could possibly be avoided.

ping/ping_common.c
```diff
@@ -806,35 +813,51 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
 		if (rts->opt_verbose && rts->ident != -1)
-			printf(_(" ident=%d"), ntohs(rts->ident));
+			if (!rts->opt_json)
+				printf(_(" ident=%d"), ntohs(rts->ident));
 
-		if (hops >= 0)
-			printf(_(" ttl=%d"), hops);
+		if (hops >= 0) {
+			if (!rts->opt_json)
+				printf(_(" ttl=%d"), hops);
```
pevik, line 831: I also wonder if whole "json or plain output" shouldn't be somehow encapsulated into internal functions. This would be hard to use on `("%d bytes from %s:"), cc, from`, where two values are printed, but for simple printing e.g. `_(" ident=%d")` it could be doable:

```c
// ping/ping_json.h
#define PRINT(msg, json_msg, type, val) \
	do { \
		if (rts->opt_json) \
			printf(msg, val); \
		else \
			construct_json(rts, type, json_msg, val); \
	} while (0)
#define PRINT_INT(msg, json_msg, val) PRINT(msg, json_msg, val, PING_JSON_INT)
#define PRINT_STR(msg, json_msg, val) PRINT(msg, json_msg, val, PING_JSON_STR)

// here
PRINT_INT(_(" ident=%d"), "ident", hops);
```
Obviously it would help if we don't translate whole C parsing (e.g. `" ident=%d"` or `" ttl=%d"`), but just the word (e.g. `"ident"` or `"ttl"`). Also it's questionable why these technical words were translated.

tacerus: Good idea, thanks for the example, will check as well. But you are saying it would break translation of these terms if implementing it using this way, would that be acceptable or will I need to implement the additional gettext logic?

pevik:
> Yeah, there is also some ping/ping6 duplication that could possibly be avoided.

I was thinking in the past how to get rid of it. Or whether rewrite from scratch would be better :) (That would take long time, thus some `ping2` implementation with subset of options would be probably way to go). But before big refactoring/rewrite it'd be great to write tests to easily verify the refactoring/rewrite.

ping/ping_common.c
pevik: The simple case (the above example which has duplicity of the strings) is ok. Additional gettext i18n improvements can be done later (unless you want to do it yourself). I don't want to overcomplicate the task.

ping/ping_common.c
```diff
@@ -806,35 +813,51 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
+				sprintf(outtime, "%ld.%02ld", (triptime + 5) / 1000,
 				       ((triptime + 5) % 1000) / 10);
 			else
-				printf(_(" time=%ld.%03ld ms"), triptime / 1000,
+				sprintf(outtime, "%ld.%03ld", triptime / 1000,
 				       triptime % 1000);
+
+			if (!rts->opt_json)
+				printf(_(" time=%s ms"), outtime);
+			construct_json(rts, PING_JSON_STR, "time", outtime);
```
pevik, line 859: I somehow thought that storing format in string is better than the time:
https://github.com/iputils/iputils/pull/574

But that way here would need another `sprintf()` so maybe it's not a good idea, maybe I should drop that PR.

tacerus: On a short glance, your patch looks cleaner than my multitude of `sprintf()` calls.

ping/ping_common.c
tacerus: Added a commit implementing this, with some slight adjustments to allow for >= 1 format specifiers using a single macro. What do you think?

tacerus:
I will clean up the commits later, leaving for now for easier diff-ing.

ping/ping_common.c
tacerus: I also implemented a macro for this now, but it is not so useful yet.
It's cumbersome to facilitate all cases in a macro or function, because the arguments passed to `error()` in the current code vary a lot. Either the message is only a type of error, like "sendmsg" (ok for JSON error type value). Or it is a long error message, like "Message too long" (ok for JSON error message value). Or it is a warning message, like "WARNING: probably, rcvbuf is not enough to hold preload" (does not fit in JSON error map at all).
In some places it is also a format string, which is particularly problematic to implement in the macro because I also use variadic arguments `error_json()`.

However I might consider dropping the latter - the additional nesting in the JSON to show different "types" of errors which I currently have might be over engineering, and I should simply make a key "error" and a value containing whatever message would otherwise be printed using `error()`.

pevik: Sounds good, I like `ERROR()`.

BTW I know it's a draft and you probably plan it, but final version should not have commits which was later removed, e.g. libjson added and then removed etc (you can fix it with `git rebase -i master` and `git push -f`).

ping/ping_common.c
pevik: LGTM, thank you!

ping/ping_common.c
tacerus: Sure, as already mentioned in https://github.com/iputils/iputils/pull/571#issuecomment-2518355436 I will clean up commits before marking as ready.

ping/ping_common.c
tacerus: Great, thanks for checking.

tacerus:
Hi @pevik,

there are some code paths which I struggle to enter for testing of error output.

So far I tested:
- "sendmsg: Message too long" with `-s`
- "local error: message too long" with `s` + `-A`
- "failed to set mark" and "missing cap_net_admin+p or cap_net_raw+p" with `-m`

Do you have any tricks to trigger the other `error()` (or now with my patch, `ERROR()`) calls without modifying the code?

test/ping/test_json.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python3
```
pevik, line 1: FYI I plan to add tests similar to perl `Test` based tests like https://github.com/schweikert/fping/tree/develop/ci (e.g. https://github.com/schweikert/fping/blob/develop/ci/test-01-basics.pl).
Although I don't like perl, using well defined testsuite helps to concentrate on tests.
It's ok to have this, but just warn you that it will be replaced.

tacerus: Sounds good, happy to replace it when we have a common framework to handle these tests!

tacerus:
Hi @pevik, any idea regarding https://github.com/iputils/iputils/pull/571#issuecomment-2525243503? Shall I finalize it as is or leave out branches I cannot validate?

ping/ping_common.c
tacerus: Rebased and adjusted.

pevik:
FYI resulted JSON is not valid: 1) missing commas between objects, 2) pings replies should be in array, 3) using keys for objects:

 i.e. instead of

```sh
$ ./builddir/ping/ping -j -c2 ::1 | jq
```
```json
{
  "bytes": 64,
  "host": "::1",
  "seq": 1,
  "ttl": 64,
  "time": "0.050"
}
{
  "bytes": 64,
  "host": "::1",
  "seq": 2,
  "ttl": 64,
  "time": "0.065"
}
{
  "host": "::1",
  "transmitted": 2,
  "received": 2,
  "duplicates": 0,
  "corrupted": 0,
  "errors": 0,
  "loss": 0,
  "time": 1028,
  "rtt": {
    "min": "0.050",
    "avg": "0.057",
    "max": "0.065",
    "mdev": "0.007"
  }
}
```
it should be:
```json
{
  "replies": [
    {
      "bytes": 64,
      "host": "::1",
      "seq": 1,
      "ttl": 64,
      "time": "0.042"
    },
    {
      "bytes": 64,
      "host": "::1",
      "seq": 2,
      "ttl": 64,
      "time": "0.063"
    }
  ],
  "result": {
    "host": "::1",
    "transmitted": 2,
    "received": 2,
    "duplicates": 0,
    "corrupted": 0,
    "errors": 0,
    "loss": 0,
    "time": 1017,
    "rtt": {
      "min": "0.042",
      "avg": "0.052",
      "max": "0.063",
      "mdev": "0.010"
    }
  }
}
```

pevik:
I'm sorry I forget that. I'm not happy about it, but understand there is no other way than separate objects.

BTW ping should write json object with summary when pressing <ctrl>+c when pinging without `-c`.

tacerus:
Thanks; for running without `-c`, printing of the summary upon Ctrl+C seems to already be working:

```
$ builddir/ping/ping -n -4 192.168.5.1 -j
{"bytes": 64, "host": "192.168.5.1", "seq": 1, "ttl": 255, "time": "0.234"}
{"bytes": 64, "host": "192.168.5.1", "seq": 2, "ttl": 255, "time": "0.293"}
{"bytes": 64, "host": "192.168.5.1", "seq": 3, "ttl": 255, "time": "0.301"}
^C{"host": "192.168.5.1", "transmitted": 3, "received": 3, "duplicates": 0, "corrupted": 0, "errors": 0, "loss": 0, "time": 2025, "rtt": {"min": "0.234", "avg": "0.276", "max": "0.301", "mdev": "0.029"}}
```

However this will unfortunately not show when piping:

```
georg@bittersuse ~/W/g/iputils (json)> builddir/ping/ping -n -4 192.168.5.1 -j |jq
{
  "bytes": 64,
  "host": "192.168.5.1",
  "seq": 1,
  "ttl": 255,
  "time": "0.271"
}
{
  "bytes": 64,
  "host": "192.168.5.1",
  "seq": 2,
  "ttl": 255,
  "time": "0.398"
}
{
  "bytes": 64,
  "host": "192.168.5.1",
  "seq": 3,
  "ttl": 255,
  "time": "0.342"
}
^C⏎  
```

Because Ctrl+C sends SIGINT to both processes.

pevik:
Ah right, thanks. OK, I guess code is working as expected, now is the time to cleanup the commits - some of them are unnecessarily split. We talk about it, but let me know if it's not obvious for you. Also some TODO might not be relevant.

tacerus:
I can finalize and clean up, I was just wondering about https://github.com/iputils/iputils/pull/571#issuecomment-2525243503, do you have any idea how to test the other cases or do you think it is fine to leave them untested?

test/ping/test_json.py
```diff
@@ -0,0 +1,57 @@
...
+if len(argv) > 1:
+  ping = argv[1]
+else:
+  ping = 'builddir/ping/ping'
+
+for arguments in [
+  ['-4', 'localhost'],
+  ['::1'],
+  ['-V'],
+  ['-s5000', '-Mdo', 'opensuse.org'],
```
pevik, line 26: FYI I tried to avoid forcing tests to connect over the internet. But OTOH real `getaddrinfo()` functionality (DNS) cannot be tested without internet.

test/ping/test_json.py
```diff
@@ -0,0 +1,57 @@
...
+  for i, line in enumerate(Popen([ping, '-jc', str(count)] + arguments, stdout=PIPE).stdout, 1):
+    print(line)
+    data = loads(line)
+    print(data)
+
+    assert isinstance(data, dict), 'data type does not match'
+
+    if arguments[0] == '-V':
+      assert 'version' in data, f'Missing "version"'
+      continue
```
pevik, line 37: Why to test version in JSON test? But anyway, this will be rewritten anyway (unfortunately to perl, because I haven't find python implementation which would work as easily as perl `Test::Command`).

ping/ping_json.h
```diff
@@ -0,0 +1,47 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef IPUTILS_PING_JSON_H
+#define IPUTILS_PING_JSON_H
+
+#define PING_JSON_NUL 0
```
pevik, line 32: `PING_JSON_NUL` is not used at all.

ping/ping_json.h
```diff
@@ -0,0 +1,47 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
pevik, line 27: IMHO the above is BSD-3-Clause. Enough would be just:

```
 /*
  * SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2024 Georg Pfuetzenreuter <mail+ip@georg-pfuetzenreuter.net>
  */
```
Would you mind just to release your code as GPLv2+? Hopefully it's possible as these two should be compatible (https://opensource.stackexchange.com/questions/7180/can-i-combine-gplv2-and-bsd-3-clause-code-and-keep-the-bsd-permissions).

```
 /*
  * SPDX-License-Identifier: GPL-2.0-or-later
  * Copyright (c) 2024 Georg Pfuetzenreuter <mail+ip@georg-pfuetzenreuter.net>
  */
```
LICENSE would have `ping: BSD-3-Clause and GPL-2.0-or-later (JSON)`.
Or do you think it's a bad idea?

test/ping/test_json.py
tacerus: I understand, the problem is that this test needs an address which is _not_ localhost, and I cannot predict which network segment the tests are executed in - hence some host on the internet was my best bet.

test/ping/test_json.py
tacerus: To assess whether the version information was printed in JSON notation as expected.

ping/ping_json.h
tacerus: I wouldn't mind, but I wanted to stay aligned with the rest of ping. I will have to read into it.

ping/ping_json.h
tacerus: Thanks, removing.

ping/ping_output.h
```diff
@@ -0,0 +1,74 @@
...
+	do { \
+		if (rts->opt_json) \
+			construct_json(rts, json_type, json_key, JSON_VALUE(__VA_ARGS__)); \
+		else \
+			printf(msg, __VA_ARGS__); \
+	} while (0)
+#define PRINT_INT(msg, json_key, ...) PRINT(msg, json_key, PING_JSON_INT, __VA_ARGS__)
+#define PRINT_STR(msg, json_key, ...) PRINT(msg, json_key, PING_JSON_STR, __VA_ARGS__)
+
+#define ERRORF(status, errnum, ...) \
```
pevik, line 43: I'm not happy with macro names (not understand what they mean), but let's leave it now. I would also not create new headers but put code into `ping/ping.h`.

test/ping/test_json.py
pevik: Yeah, unlike errors, `-V` gets into JSON. Sorry for the noise.

ping/ping_output.h
tacerus: I understand, some are ambiguous too. I split it into a separate header file because ping.h is already quite large, but if you prefer I can merge it back in?

pevik:
> > Do you want to test error handling because you touched that? In the end string is not formatted in JSON, thus not changed much...
> 
> Some of the error messages were adapted to be printed in JSON format if executed with `-j`.

It would be good to have them all or none (not something in between). At least these aren't:
```
$ ./builddir/ping/ping -j
./builddir/ping/ping: usage error: Destination address required

$ ./builddir/ping/ping -j -c2 ::1 -s5588888888
./builddir/ping/ping: invalid argument: '5588888888': out of range: 0 <= value <= 2147483647
```
If it's too complicated, we can ignore errors in JSON. But as you've done the infrastructure, it shouldn't be hard to convert rest of them.

ping/ping_output.h
pevik: It does not look important enough to have it separated. In the end JSON support would get into other tools (at least to arping), then it would go to `iputils_common.h`.

But it's not important, C code itself matters more (+ good names for macros and reconsider if `static inline` would be more readable instead of macros). Next week after LTP release I'll have a deeper look and ask @metan-ucw to do review as well.

pevik:
Does it make any sense to `Refactor JSON errors` to be a separate commit? IMHO not.
And certainly the last commit `Mark internal JSON functions as static` is not useful on it's own.
But maybe you plan it just haven't done it yet.

I would separate into it's own commit only adding tests. Usually some code preparation are separated change smaller, but I don't see anything to be here.

tacerus:
Sure, I will look into the other error messages later, but adjusted the commits already now.

tacerus:
For "usage error" and "invalid argument", it is not quite possible to use the existing macros, because at the point they occur `rts` needs to be accessed differently. I can add the "expanded" variant to all these occurrences or introduce yet another macro for this purpose. Do you think that is worth it or would you rather just skip the JSON output for usage errors?

ping/ping_output.h
metan-ucw: First of all, why are there such horrid macros to begin with? We can have just a regular functions such as 'void print_int(const char *msg, const char *key, int val)' instead. These can be static inline as well as @pevik suggested, or we can put them into ping_output.c. The biggest problem with macros as these is that we loose all type checking.

ping/ping.c
```diff
@@ -595,6 +597,23 @@ main(int argc, char **argv)
...
 
+	if (rts.opt_version) {
+		char version_info[50];
+
+		if (rts.opt_json) {
+			sprintf(version_info, IPUTILS_VERSION("ping"));
+			// TODO: cleaner JSON version output, only version number in "version" and add map with build config
+			version_info[strcspn(version_info, "\n")] = 0;
+			construct_json(&rts, PING_JSON_STR, "version", version_info);
+			print_json_packet(&rts);
```
metan-ucw, line 608: This is way to ugly and we have just the version in `PACKAGE_VERSION` macro, so we should really just pass the `PACKAGE_VERSION` to the function that constructs the JSON.

ping/ping.c
```diff
@@ -1773,9 +1795,13 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
...
 		pr_options(rts, opts, olen + sizeof(struct iphdr));
 
-		putchar('\n');
-		fflush(stdout);
+		// handled by print_json()
+		if(!rts->opt_json) {
+			putchar('\n');
+			fflush(stdout);
+		}
 	}
```
metan-ucw, line 1803: Generally I would rather vote for adding an API for the generic output and do all the stuff there.

In this place we should call a function `output_finish_line()` or something like that and keep the code that deals with `stdout` in the `ping_output.h` or `ping_output.c`.

And the same with the version, we should just have a function called `output_ping_version()` call it from here and let the code in `ping_output.c` do the work. The code in `ping.c` is way to complex already adding more code there will not help.

metan-ucw (changes requested):

ping/ping_output.h
tacerus: I understand, and agree functions would be better suited; in an earlier comment both variants were suggested, I opted for the macro one, but it turned out to require multiple hoops.
Will try to refactor.

ping/ping.c
tacerus: Any reason `PACKAGE_VERSION` was not used for the regular output so far? I would suggest to adopt both equally, but `PACKAGE_VERSION` returns only

```
20240905-42-g9599aaae034e+libcap: yes, IDN: yes, NLS: yes, error.h: yes, getrandom(): yes, __fpending(): yes
```


whereas `IPUTILS_VERSION("ping")` yields


```
ping from iputils 20240905-42-g2366f85cb9f3+
libcap: yes, IDN: yes, NLS: yes, error.h: yes, getrandom(): yes, __fpending(): yes
```

I think the `ping from iputils` part is not needed, but it is there so far.

ping/ping.c
tacerus: Thanks, good idea, will adapt.

tacerus:
Hi @metan-ucw, thanks for your comments, I added some commits to incorporate the suggestions, please check before I combine them.

Now with ping_output.c, do we want to keep ping_json.{c,h} or should I merge both into ping_output.{c,h}?

ping/ping_output.c
```diff
@@ -0,0 +1,100 @@
...
+}
+
+inline void ping_print_finish(struct ping_rts *rts) {
+	if(!rts->opt_json) {
+		putchar('\n');
+		fflush(stdout);
+	}
+}
+
+void ping_error(struct ping_rts *rts, int status, int errnum, char *msg) {
```
pevik, line 75: This function is not needed, the following `ping_errorf()` can handle the case when only unformatted message is passed. Please delete this function and rename `ping_errorf()` to `ping_error()`.

ping/ping_output.h
```diff
@@ -0,0 +1,36 @@
...
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
```
pevik, line 26: Could you please instead of this old garbabe just use SPDX?
```
/* SPDX-License-Identifier: BSD-3-Clause */
```
I would prefer to have new code in GPLv2+, but not sure if possible to mix (https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses states these two are compatible)
```
/* SPDX-License-Identifier: GPL-2.0-or-later */
```

ping/ping_output.c
tacerus: Thanks for the hint, adjusted.

ping/ping_output.h
tacerus: Sure, adjusted. I am fine with GPL, but I have not yet found time to read into it and IANAL. I am not sure if dual licensing causes any issues for the project because there is no way to compile out the other functionality, so it would not be possible to update the software should there be some environment where GPL is not possible (not that I know of any myself though).

tacerus:
Squashed away the refactoring commits now, as there didn't seem to be any objections to the new functions. :)

test/ping/test_json.py
```diff
@@ -0,0 +1,68 @@
...
+      assert data.get('ewma')
+    elif 'error' in data:
+      errdata = data['error']
+      assert isinstance(errdata, list)
+      if arguments[1] == '-Mdo':
+        assert errdata[0] == 'sendmsg'
+        # in Alpine, it is "Message too large"
+        assert errdata[1] in ['Message too long', 'Message too large']
+    else:
+      print('Excess data! Missing test coverage or stray output?')
```
pevik, line 67: nit: printing errors into stderr? Because `assert()` also does output into stderr.

pevik:
> Squashed away the refactoring commits now, as there didn't seem to be any objections to the new functions. :)

FYI I normally do a local branch backup to keep the old code. Because pushing new commits is really unusual (make things easy for you but not for the reviewers).

tacerus:
I keep local branch backups as well, I had feedback from reviewers on other projects in the past, they prefer to review individual commits for easier diff-ing. Good to know you are fine with pushing squashed commits directly.

test/ping/test_json.py
tacerus: Good idea, adjusted.

tacerus:
Albeit not being complete - there are still regular print's in branches I did not touch yet due to not being able to enter them for testing (https://github.com/iputils/iputils/pull/571#issuecomment-2585412298) - I think it is time to mark as ready to get a foundation in place.

ping/ping_common.c
```diff
@@ -941,25 +949,48 @@ int finish(struct ping_rts *rts)
...
+		sprintf(rttmin, "%ld.%03ld", (long)rts->tmin / 1000, (long)rts->tmin % 1000);
+		sprintf(rttavg, "%lu.%03ld", (unsigned long)(tmavg / 1000), (long)(tmavg % 1000));
+		sprintf(rttmax, "%ld.%03ld", (long)rts->tmax / 1000, (long)rts->tmax % 1000);
+		sprintf(rttmdev, "%ld.%03ld", (long)tmdev / 1000, (long)tmdev % 1000);
+
+		if (!rts->opt_json)
+			printf(_("rtt min/avg/max/mdev = %s/%s/%s/%s ms"),
+				rttmin, rttavg, rttmax, rttmdev);
+
+		construct_json_statistics(rts, tv, rttmin, rttavg, rttmax, rttmdev);
```
metan-ucw, line 966: Th code looks much better, but why don't we add a function `ping_print_statistics()` into the `ping_output.h` and put all this code in there? And the same tome for the flood option.

If I was writing this patch, most if not all the `if (!rts->opt_json)` would have been in the `ping_output.c`.

tacerus: Sure that sounds good, and I refactored it further now to keep all the JSON conditions in `ping_output.c` as well as to deduplicate some of the printing logic between ping and ping6, but my main issue still is that I don't know how to test some of the code branches I am touching, so with such big printing refactoring there is a good chance I might break formatting of some odd error conditions.

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+#include "ping.h"
+#include "stdarg.h"
+
+static inline void json_emergency(char *msg) {
+	error(1, 0, "%s", msg);
+}
+
+static inline void test_buflen(int have) {
+	if (have < 0 || have >= PING_JSON_MAX)
+		json_emergency("Overflow during JSON construction.");
```
pevik, line 15: nit: Although I'm not a fan of iputils being translated, when we translate everything, this should be also translated (with `_(...)`). Also please no dot (`.`) at the end of the message.

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+	char * val_str;
+	int val_int;
+	unsigned int val_uint;
+	
+	va_list ap;
+	va_start (ap, *key);
+
+	if (*rts->json_packet) {
+		json_continue(rts->json_packet);
+	} else {
```
pevik, line 108: nit: brackets aren't needed when both `if` and `else` clauses contain single statement (FYI https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces).

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+			json_kv_int(rts->json_packet, key, val_int);
+			break;
+
+		case PING_JSON_UINT:
+			val_uint = va_arg (ap, unsigned int);
+			json_kv_uint(rts->json_packet, key, val_uint);
+			break;
+	}
+
+
```
pevik, line 133: nit: please no two blank lines

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+static void json_kv_object(char *part, char *key, char *value) {
+	json_end_object(value);
+	size_t curlen = strlen(part);
+	test_buflen(snprintf(part + curlen, PING_JSON_MAX - curlen, "\"%s\": %s", key, value));
+}
+
+void construct_json(struct ping_rts *rts, int ptype, char *key, ...) {
+	if (!rts->opt_json) {
+		return;
+	}
```
pevik, line 97: nit also here please without `{ }` brackets.

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+	if (rts->pipesize > 1) {
+		json_continue(rts->json_stats);
+		json_kv_int_continue(rts->json_stats, "pipe", rts->pipesize);
+	}
+}
+
+void construct_json_statistics_flood(struct ping_rts *rts, char *ipg, char *ewma) {
+	if (!rts->opt_json || !*rts->json_stats) {
+		return;
+	}
```
pevik, line 175: and here no `{ .. }` brackets.

ping/ping_output.c
```diff
@@ -0,0 +1,230 @@
...
+		printf(msg, json_val);
+	}
+}
+
+inline void ping_print_version(struct ping_rts *rts) {
+	if (rts->opt_json) {
+		construct_json(rts, PING_JSON_STR, "version", PACKAGE_VERSION);
+		print_json_packet(rts);
+	} else {
+		printf(PACKAGE_VERSION);
```
pevik, line 38: You replaced `printf(IPUTILS_VERSION("ping"));` with `printf(PACKAGE_VERSION)` why? That caused to print just git hash instead of `ping from iputils 20240905-42-ge6cc75f\n`. Please put it back.

Before:
```C
$ ./builddir/ping/ping -V
ping from iputils 20240905-42-ge6cc75f
libcap: yes, IDN: yes, NLS: yes, error.h: yes, getrandom(): yes, __fpending(): yes
```
With your patch:
```C
$ ./builddir/ping/ping -V
20240905-44-gbe6d3dblibcap: yes, IDN: yes, NLS: yes, error.h: yes, getrandom(): yes, __fpending(): yes
```

ping/ping_output.c
```diff
@@ -0,0 +1,230 @@
...
+	if (!rts->opt_json) {
+		printf(_(" (truncated)"));
+		if (rts->ni.subject_type == IPUTILS_NI_ICMP6_SUBJ_IPV4)
+			putchar('\n');
+	}
+	construct_json(rts, PING_JSON_STR, "info", "truncated");
+}
+
+inline void ping_print_packet(struct ping_rts *rts) {
+	if (!rts->opt_json) {
```
pevik, line 70: Please use:
```C
if (rts->opt_json)
    return;
```
to avoid whole function code being indented.

ping/ping_output.c
```diff
@@ -0,0 +1,230 @@
+/*
+ * SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2024-2025 Georg Pfuetzenreuter <mail+ip@georg-pfuetzenreuter.net>
```
pevik, line 3: nit: We never bothered with updating copyright (now code changed a lot) for C code, I should have updated it. You base your code on _this_ file on existing iputils code (I would say at least from 2015 78aa5c1516d6f299331b4cd0efac5f0f53a28c0f, which was based on old code, some algorithms are older than 2006 when project was imported to git), IMHO there should be also iputils project copyright. Maybe them add below your copyright?
```
* Iputils project (c) 2014-2024
```

ping/ping_common.c
```diff
@@ -877,89 +884,14 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
+	ping_print_finish(rts);
 
-		tmdev = llsqrt(tmvar);
-
-		printf(_("rtt min/avg/max/mdev = %ld.%03ld/%lu.%03ld/%ld.%03ld/%ld.%03ld ms"),
-		       (long)rts->tmin / 1000, (long)rts->tmin % 1000,
-		       (unsigned long)(tmavg / 1000), (long)(tmavg % 1000),
-		       (long)rts->tmax / 1000, (long)rts->tmax % 1000,
-		       (long)tmdev / 1000, (long)tmdev % 1000);
-		comma = ", ";
```
pevik, line 949: This is ugly, but actually very old change from 2010 964d0776f59e8ca5917c6368f90974446c2b2b32 (clearer would be to store `bool` and print `", "` below if set).

ping/ping_json.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifndef IPUTILS_PING_JSON_H
+#define IPUTILS_PING_JSON_H
+
+#define PING_JSON_STR 1
+#define PING_JSON_INT 2
+#define PING_JSON_ARR 3
+#define PING_JSON_UINT 4
+#define PING_JSON_MAX 1000
```
pevik, line 13: I suppose this is sufficient, because FQDN is 255.

OTOH if I add crazy long record into `/etc/hosts`:

```
192.168.122.43 test
192.168.122.43 A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100
```

it works with normal ping:
```
$ ./builddir/ping/ping test
64 bytes from A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100A111111111abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij12345678909999999100 (192.168.122.43): icmp_seq=1 ttl=64 time=0.180 ms
```
but not with JSON:
```
$ ./builddir/ping/ping test -j
./builddir/ping/ping: Overflow during JSON construction.
```

ping/ping_json.h
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2024 Georg Pfuetzenreuter <mail+ip@georg-pfuetzenreuter.net>
+ */
+
+#ifndef IPUTILS_PING_JSON_H
+#define IPUTILS_PING_JSON_H
+
+#define PING_JSON_STR 1
+#define PING_JSON_INT 2
+#define PING_JSON_ARR 3
+#define PING_JSON_UINT 4
```
pevik, line 12: nit:I would define this as `enum`.

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+	json_kv_int(part, key, value);
+	json_continue(part);
+}
+
+/* so far only supports arrays of strings */
+static void json_kv_array(char *part, char *key, va_list ap) {
+	size_t curlen = strlen(part);
+	test_buflen(snprintf(part + curlen, PING_JSON_MAX - curlen, "\"%s\": [", key));
+
+	char * value;
```
pevik, line 67: A bit more consistent formatting would be nice. `*` should be without following space `char *value`. I would also name it just `val` (clear enough).

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+	if (!rts->opt_json) {
+		return;
+	}
+
+	char * val_str;
+	int val_int;
+	unsigned int val_uint;
+	
+	va_list ap;
+	va_start (ap, *key);
```
pevik, line 104: Also here please no space before `(` (=> `va_start (ap, *key);`).

ping/ping.c
pevik: FYI `ping from iputils` is definitely needed because there are other ping implementations, e.g. busybox or inetutils. And because some distros (e.g. Debian) support all of them, it's needed to see not only the version but from which project it is.

Also it would be unrelated change (if we wanted it, it would have to be in a separate commit).

pevik: > Any reason `PACKAGE_VERSION` was not used for the regular output so far?

`PACKAGE_VERSION` is used as part of `IPUTILS_VERSION()` it's not intended to be used separately. Please use `IPUTILS_VERSION("ping")` to fix the output.

ping/ping_output.c
pevik: Thanks for info, I finally understand what you mean and replied there.

ping/ping.c
```diff
@@ -595,6 +597,11 @@ main(int argc, char **argv)
 		}
 	}
 
+	if (rts.opt_version) {
+		ping_print_version(&rts);
+		exit(0);
```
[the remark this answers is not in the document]
tacerus: I introduced it because argument parsing needs to complete to populate `opt_json` before calling `ping_print_version()`. If I call the function during the `-V` argument switch statement, JSON output will only invoke if `-j` is passed after `-V` in the command line, as opposed to allowing `-j` at any place in the command line.

I can replace `rts.opt_version` with a `opt_version` boolean local to `main()`.

ping/ping_output.c
tacerus: I added

```
 * Copyright (c) 2014-2024 Iputils project
 * Copyright (c) 1989-2006 The Regents of the University of California
```

now, but I'm not sure who/what's between 2006 and 2014, would it classify as `Iputils project` as well? Or do you think the `Iputils project` line is enough? I'm not sure quite how old the algorithms the commit you mentioned bases on are.

ping/ping_json.h
tacerus: Right, RFC 1123 says

>       Host software MUST handle host names of up to 63 characters and
>       SHOULD handle host names of up to 255 characters.

Since we allow more for non-JSON output, it would make sense to allow it for JSON as well (even if it won't be possible to match it exactly as the surrounding output is different). I cannot quite find what the maximum for non-JSON output is though, do you know if there is one?

ping/ping_json.c
tacerus: My bad, aligned now, and s/value/val/g.

ping/ping_json.c
```diff
@@ -0,0 +1,217 @@
...
+ */
+
+#include "ping.h"
+#include "stdarg.h"
+
+static inline void json_emergency(char *msg) {
+	error(1, 0, "%s", msg);
+}
+
+static inline void test_buflen(int have) {
```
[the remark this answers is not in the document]
tacerus: Good point, renamed.

ping/ping_common.c
tacerus: I thought as much, but using a bool I will have to add if-conditionals and duplicate the `printf()` calls every time one can either occur with or without a comma.

tacerus: Never mind that, I got it now.

ping/ping_output.c
tacerus: Adjusted, but I kept the `construct_json()` call inside the condition, as otherwise `opt_json` gets uselessly checked twice if it is false (once in `ping_print_packet()` and once in `construct_json()`).

ping/ping.c
tacerus: Thanks for the clarification! Reverted to use `IPUTILS_VERSION("ping")` in non-JSON output, but kept it for JSON output for now (if wanted in the future, the build information should probably be under a separate key).

ping/ping_json.c
tacerus: Sure!

ping/ping_output.c
tacerus: Thanks!

pevik:
@tacerus Thanks for fixing remaining problems.On a first look LGTM, I'll have deeper look next week + ping @metan-ucw. Hopefully we could finally merge. Thanks for your patiency.

ping/ping_output.c
```diff
@@ -0,0 +1,235 @@
...
+ * Copyright (c) 2014-2024 Iputils project
+ * Copyright (c) 1989-2006 The Regents of the University of California
+ */
+
+#include "ping.h"
+#include "stdarg.h"
```
pevik, line 9: Please system headers should be in `<>` => `#include <stdbool.h>`
nit: Also add `#include <stdarg.h>` for `bool` (which I would prefer `bool` over `_Bool`.

pevik:
@tacerus FYI I did the above minor changes (`bool`) + made basic formatting cleanup to fix tiny issues reported by [`checkpatch.pl` from kernel](https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl):

```
$ kernel/scripts/checkpatch.pl -f --no-tree --terse --no-summary --ignore CONST_STRUCT ping/ping_json.h
$ kernel/scripts/checkpatch.pl -f --no-tree --terse --no-summary --ignore CONST_STRUCT ping/ping_json.c
$ kernel/scripts/checkpatch.pl -f --no-tree --terse --no-summary --ignore CONST_STRUCT ping/ping_output.h
$ kernel/scripts/checkpatch.pl -f --no-tree --terse --no-summary --ignore CONST_STRUCT ping/ping_json.c
```
(NOTE: if you want to try it yourself, you also need https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/spelling.txt)

iputils code is not perfect, but let's new code to be modern and use kernel formatting.

Unless @metan-ucw finds anything else, I'm going to merge with changes made in branch `json.fixes` in my fork: https://github.com/pevik/iputils/tree/json.fixes. Ideally you add my fork, apply the changes and force push (`git remote add pevik https://github.com/pevik/iputils.git && git fetch pevik && git reset --hard pevik/json.fixes && git push -f`) so that @metan-ucw can do review on the latest code.

I do not post the diff here (500 lines), but you can see it with `git diff tacerus/json..pevik/json.fixes`.

pevik:
```
```

ping/ping_output.c
```diff
@@ -0,0 +1,235 @@
...
+
+	if (!rts->opt_json)
+		printf(_("rtt min/avg/max/mdev = %s/%s/%s/%s ms"),
+			rttmin, rttavg, rttmax, rttmdev);
+
+	construct_json_statistics(rts, tv, rttmin, rttavg, rttmax, rttmdev);
+}
+
+inline void ping_print_finish(struct ping_rts *rts) {
+	_Bool comma = 0;
```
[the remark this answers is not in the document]
tacerus: Also good with me.

tacerus:
Thanks for reviewing and for the styling improvements! The diff looks ok, I pushed the your commits.

ping/ping_json.c
```diff
@@ -0,0 +1,241 @@
...
+}
+
+static void json_start_object(char *part)
+{
+	test_buflen(snprintf(part, PING_JSON_MAX, "{"));
+}
+
+static void json_end_object(char *part)
+{
+	size_t curlen = strlen(part);
```
metan-ucw, line 34: Wouldn't it make sense to turn the json buffer into a structure that remembers the current offset to the buffer?

Also why do we buffer the data like this, is there a reason why aren't we just printing them into stdout? The only reason I can think of is that an error may happen during a construction of a JSON result data and we do not want to print the partially constructed reply and just emit the error instead.

tacerus: I explained this partially in https://github.com/iputils/iputils/pull/571#issuecomment-2513195980 when different libraries were discussed - your summary is pretty much the same - we do not want to leave behind defective JSON which can break tools processing the piped JSON output.

pevik:
Also missing `-j` in help, e.g.

```diff
diff --git ping/ping_common.c ping/ping_common.c
index df2bfcb..ecb04d4 100644
--- ping/ping_common.c
+++ ping/ping_common.c
@@ -67,6 +67,7 @@ void usage(void)
                "                     destinations or for -f), override -n\n"
                "  -I <interface>     either interface name or address\n"
                "  -i <interval>      seconds between sending each packet\n"
+               "  -j                 JSON output (experimental)\n"
                "  -L                 suppress loopback of multicast packets\n"
                "  -l <preload>       send <preload> number of packages while waiting replies\n"
                "  -m <mark>          tag the packets going out\n"
```
Or feel free to again use ;json.fixes; in my fork: https://github.com/pevik/iputils/tree/json.fixes.

tacerus:
Thanks!

ping/ping_json.c
metan-ucw: Right, we do not want to print broken JSON. However doing strlen() each time we want to append to the buffer looks ugly to me, I would at least store the offset to the buffer.

Also sorry that I missed comment #571 the ujson library can be trivially made to buffer data when writing, all that is needed is a callback that appends to the buffer instead of the default callback that writes into a file.

tacerus: Sure, that I understand, I will look into storing it in a struct.

And no worries, eventually we concluded not to introduce additional libraries, and I suppose we don't need all the extra functionality ujson provides anytime soon.

okias:
```
Acked-by: David Heidelberg <david@ixit.cz>
```
and tested too! Great work!

okias:
```
  "host": "ip-89-177-23-149.bb.vodafone.cz (89.177.23.149)",
```
any chance to split these two? It seems to be not optimally parseable.

ping/ping_common.c
```diff
@@ -806,25 +808,27 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
 		uint8_t *cp, *dp;
 
 		print_timestamp(rts);
-		printf(_("%d bytes from %s:"), cc, from);
+		ping_print_int(rts, _("%d bytes "), "bytes", cc);
+		// TODO: split host IP address and name into separate attributes in JSON?
```
okias, line 812: yes, please

tacerus:
Hi @okias, thanks for checking. That was a concern of mine as well, but I only left it as TODO in the code

```
+               // TODO: split host IP address and name into separate attributes in JSON?
+               ping_print_str(rts, _("from %s:"), "host", from);
```

as it will require some more logic (splitting the existing `from` felt a bit ugly, better might be to keep both values separate and then only join them in non-JSON mode).

okias:
> as it will require some more logic (splitting the existing from felt a bit ugly, better might be to keep both values separate and then only join them in non-JSON mode).

Yes, keeping the values separate was what I was thinking too.

I think the options are:
1. Do the split in a separate PR, and then base this PR on top of it.

2. Do the split as a follow-up commit.

3. Do it after merging this—though in that case, it would be good to avoid a release until the split PR lands.

Personally, I’d lean towards option 1, but it’s up to @pevik to decide how he’d like to handle it. `// I just got too excited about it! :)`

tacerus:
I did not intend to include such change in or before this patch. It probably requires refactoring of `*_pr_addr()` and adjusting places where that gets called.

pevik:
@okias Thanks for pointing this out. I'd leave it as a separate effort, it'd be good to get this finally merged.

ping/ping_json.c
tacerus: Adjusted.

ping/ping_json.c
```diff
@@ -0,0 +1,247 @@
...
+
+	json_kv_str_continue(&rts->json_stats, "ipg", ipg);
+	json_kv_str(&rts->json_stats, "ewma", ewma);
+}
+
+void print_json_and_reset(struct ping_json_buffer *json_packet)
+{
+	printf("%s}\n", json_packet->object);
+	fflush(stdout);
+	*json_packet->object = '\0';
```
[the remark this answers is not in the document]
tacerus: Thanks, done.

tacerus:
Seems the added changes + rebase to master caused some tests to break, and Valgrind reports a memory leak, I'll look into that.

tacerus:
Solved that now (by initializing `json_rtt.size` first - not fully clear why it isn't needed for the other buffers).

pevik:
@tacerus Would you mind to rebase to the current master?

tacerus:
Sure, done.

pevik:
Thank you, the ball is on our side. This time I'd really like to get it merged.
@metan-ucw FYI

test/ping/meson.build
```diff
@@ -93,3 +93,6 @@ foreach dst : [ '127.0.0.1' ] + ipv6_dst
 	test(name, cmd, args : args, should_fail : not run_as_root)
   endforeach
 endforeach
+
+jsontest = find_program('test_json.py', 'test/ping/test_json.py', 'test/ping/test_json.py', required : true)
```
pevik, line 97: nit: this could be just 
```python
jsontest = find_program('test_json.py', 'test/ping/test_json.py', required : true)
```
or even
```python
jsontest = find_program('test_json.py', required : true)
```

pevik:
I found a bug.

Previously:
```
$ ping -c1 -i0.001 127.0.0.1
ping: cannot flood, minimal interval for user must be >= 2 ms, use -i 0.002 (or higher)
```

But now the output is shortened (missing `= 2 ms, use -i 0.002 (or higher)`):
```
$ ./builddir/ping/ping -c1 -i0.001 127.0.0.1
./builddir/ping/ping: cannot flood, minimal interval for user must be >
```

tacerus:
Now it should be restored in both modes:

```
$ builddir/ping/ping -c1 -i0.001 127.0.0.1
builddir/ping/ping: cannot flood, minimal interval for user must be >= 2 ms, use -i 0.002 (or higher)

$ builddir/ping/ping -c1 -i0.001 127.0.0.1 -j
{"error": "cannot flood, minimal interval for user must be >= 2 ms, use -i 0.002 (or higher)"}
```

ping/ping_output.c
```diff
@@ -0,0 +1,255 @@
...
+	sprintf(rttmin, "%ld.%03ld", (long)rts->tmin / 1000, (long)rts->tmin % 1000);
+	sprintf(rttavg, "%lu.%03ld", (unsigned long)(tmavg / 1000), (long)(tmavg % 1000));
+	sprintf(rttmax, "%ld.%03ld", (long)rts->tmax / 1000, (long)rts->tmax % 1000);
+	sprintf(rttmdev, "%ld.%03ld", (long)tmdev / 1000, (long)tmdev % 1000);
+
+	if (!rts->opt_json)
+		printf(_("rtt min/avg/max/mdev = %s/%s/%s/%s ms"),
+			rttmin, rttavg, rttmax, rttmdev);
+
+	construct_json_statistics(rts, tv, rttmin, rttavg, rttmax, rttmdev);
```
metan-ucw, line 160: This part is a bit strange, we shouldn't call to the `construct_json_statistics()` unconditionally just to print a newline from that functions `if (!rts->opt_json)`. Rather than that this should really be:

```
if (!rts->opt_json) {
		printf(_("rtt min/avg/max/mdev = %s/%s/%s/%s ms"),
			     rttmin, rttavg, rttmax, rttmdev);
        putchar('\n');
} else {
        construct_json_statistics(rts, tv, rttmin, rttavg, rttmax, rttmdev);
}
```

Generally we shouldn't call to the functions that build and print json if json output is not enabled.

tacerus: Good point, adjusted.

tacerus:
This new CI test failure does not sound related to my changes, or am I missing something?

https://github.com/iputils/iputils/actions/runs/17012246798/job/48229840070?pr=571

```
+ [ -x builddir/arping ]
+ file builddir/arping
+ grep -E arping.*32-bit .*(executable|shared object).*80386.*dynamically linked
Error: Process completed with exit code 1.
```

pevik (changes requested):
msg should be 300 (adding this comment to show my comments as public).

pevik (approved):
Thank you! For both commits:
```
```

I suggest to merge with following diff to the first commit. Either do it (you can reset to my fork https://github.com/pevik/iputils/commits/json.fixes/) and rebase or I'll do it before merge):
```diff
diff --git ping/ping.h ping/ping.h
index c4cbc11..f9efaee 100644
--- ping/ping.h
+++ ping/ping.h
@@ -136,6 +136,7 @@ typedef struct ping_func_set_st {
 
 /* late include as dependent on ping_rts */
 #include "ping_json.h"
+#include "ping_output.h"
 
 /* Node Information query */
 struct ping_ni {
diff --git ping/ping_json.h ping/ping_json.h
index c516ec1..e236639 100644
--- ping/ping_json.h
+++ ping/ping_json.h
@@ -6,7 +6,7 @@
 #ifndef IPUTILS_PING_JSON_H
 #define IPUTILS_PING_JSON_H
 
-#define PING_JSON_MAX 1000
+#define PING_JSON_MAX 4000
 
 enum PING_JSON_TYPE {
        PING_JSON_STR,
@@ -15,8 +15,6 @@ enum PING_JSON_TYPE {
        PING_JSON_UINT,
 };
 
-#include "ping/ping_output.h"
-
 struct ping_json_buffer {
        size_t size;
        char object[PING_JSON_MAX];
diff --git po/POTFILES po/POTFILES
index 83465b3..7fbbc9d 100644
--- po/POTFILES
+++ po/POTFILES
@@ -1,7 +1,9 @@
 arping.c
 clockdiff.c
 ping/node_info.c
-ping/ping6_common.c
 ping/ping.c
+ping/ping6_common.c
 ping/ping_common.c
+ping/ping_json.c
+ping/ping_output.c
 tracepath.c

```

metan-ucw (approved):
With the two minor things I've pointed out fixed you can add my `Reviewed-by: Cyril Hrubis <chrubis@suse.cz>`

pevik:
OK, I see first output regression: https://github.com/iputils/iputils/issues/605.

tacerus:
Thanks for reviewing and merging, I will investigate your follow-ups next week.

PR #574, opened 2024-12-04
Reduce duplicity in translation strings

pevik:
Tiny i18n related improvement. Phrase `"time=... ms"` should be translated only once. To achieve it avoid float printf formatting in `_()` gettext.


pevik:
@nmeyerhans This would be touched in https://github.com/iputils/iputils/pull/571 anyway, thus do cleanup before.

Issue #584, opened 2025-05-04
Signed 64-bit integer overflow in RTT calculation 


Zephkek:
Hello,

A specially crafted ICMP Echo Reply can trigger a possible signed 64-bit integer overflow in the `iputils` ping RTT calculation. This occurs due to uncontrolled multiplication of the timestamp seconds by 1,000,000, causing undefined behavior. The vulnerability may present itself as runtime errors detected by AddressSanitizer or as silent failures, resulting in repeated zero RTT readings and incorrect statistics.

### Affected Versions

* `iputils ping` (current `master` branch)
* Ubuntu package version `iputils-ping 3:20240117-1build1`

### Environment

* Ubuntu 24.04.1 LTS (Noble Numbat)
* Kernel: `5.15.167.4-microsoft-standard-WSL2`
* Architecture: `x86_64`

### Detailed Steps and PoC

Full details, reproduction steps, and PoC script can be found here:

[https://github.com/Zephkek/ping-rtt-overflow/](https://github.com/Zephkek/ping-rtt-overflow/)

Bug found by Mohamed Maatallah

carnil:
Looks this issue got CVE-2025-47268

nmeyerhans:
The real world security implications of this seem dubious, at best. I might get unreliable results from pinging a malicious host?

Zephkek:
I'd imagine it's more than just unreliable results atleast in my opinion. The unreasonable RTT values from the overflow could cause problems in monitoring systems that rely on iputils ping. Also, I noticed that the affected RTT values impact the ping adaptive mode interval making it enter a busy-wait state that consumes 100% CPU regardless of user privileges this could also be used for DoS. 

https://github.com/user-attachments/assets/df6d892a-0a6c-467b-902c-f736ae121759

pevik:
> Also, I noticed that the affected RTT values impact the ping adaptive mode interval making it enter a busy-wait state that consumes 100% CPU regardless of user privileges this could also be used for DoS.

No, high load caused by `-A` used in your video is _not_ affected by the bug. `-A` causes high load even without crafted RTT values (e.g. another problem). But IMHO `-A` is not used in Nagios/Icinga monitoring plugins.

Zephkek:
Based on the debugging, the chain of events appears clear. The `update_interval` calculation relies on the RTT value, which gets corrupted due to the integer overflow triggered by the poc (as confirmed by ASan).

[...]

PR #585, opened 2025-05-05
ping: Fix signed 64-bit integer overflow in RTT calculation

pevik:
Crafted ICMP Echo Reply packet can cause signed integer overflow in

1) triptime calculation:
triptime = tv->tv_sec * 1000000 + tv->tv_usec;

2) tsum2 increment which uses triptime
rts->tsum2 += (double)((long long)triptime * (long long)triptime);

3) final tmvar:
tmvar = (rts->tsum2 / total) - (tmavg * tmavg)
---

    $ export CFLAGS="-O1 -g -fsanitize=address,undefined -fno-omit-frame-pointer"
    $ export LDFLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer"
    $ meson setup .. -Db_sanitize=address,undefined
    $ ninja
    $ ./ping/ping -c2 127.0.0.1

    64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.061 ms
    ../ping/ping_common.c:757:25: runtime error: signed integer overflow: -2513732689199106 * 1000000 cannot be represented in type 'long int'
    ../ping/ping_common.c:757:12: runtime error: signed integer overflow: -4975495174606980224 + -6510615555425289427 cannot be represented in type 'long int'
    ../ping/ping_common.c:769:47: runtime error: signed integer overflow: 6960633343677281965 * 6960633343677281965 cannot be represented in type 'long int'
    24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
    ./ping/ping: Warning: time of day goes back (-7256972569576721377us), taking countermeasures
    ./ping/ping: Warning: time of day goes back (-7256972569576721232us), taking countermeasures
    24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
    ../ping/ping_common.c:265:16: runtime error: signed integer overflow: 6960633343677281965 * 2 cannot be represented in type 'long int'
    64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.565 ms

    2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 1002ms
    ../ping/ping_common.c:940:42: runtime error: signed integer overflow: 1740158335919320832 * 1740158335919320832 cannot be represented in type 'long int'
    rtt min/avg/max/mdev = 0.000/1740158335919320.832/6960633343677281.965/-1623514645242292.-224 ms

To fix the overflow check allowed ranges of struct timeval members:
* tv_sec <-LONG_MAX/1000000, LONG_MAX/1000000>
* tv_usec <0, 999999>

Fix includes 2 new error messages (needs translation).

After fix:

    $ ./ping/ping -c2 127.0.0.1
    64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.059 ms
    ./ping/ping: Warning: overflow tv_usec -6510615555425457380 us
    ./ping/ping: Warning: invalid tv_sec -1789369274859522 s
    24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
    ./ping/ping: Warning: overflow tv_usec -6510615555425413387 us
    ./ping/ping: Warning: invalid tv_sec -2006106209517570 s
    24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
    64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.118 ms

    2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 1002ms
    rtt min/avg/max/mdev = 0.000/0.044/0.118/0.048 ms

Fixes: https://github.com/iputils/iputils/issues/584
Fixes: CVE-2025-472
Link: https://github.com/Zephkek/ping-rtt-overflow/
Co-developed-by: Cyril Hrubis <chrubis@suse.cz>
Reported-by: Mohamed Maatallah <hotelsmaatallahrecemail@gmail.com>

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -754,16 +754,32 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
 
 restamp:
 		tvsub(tv, &tmp_tv);
-		triptime = tv->tv_sec * 1000000 + tv->tv_usec;
-		if (triptime < 0) {
-			error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), triptime);
+
+		if (tv->tv_usec >= 1000000) {
+			error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
+			tv->tv_usec = 999999;
+		}
+
+		if (tv->tv_usec < 0) {
+			error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
+			tv->tv_usec = 0;
+		}
+
+		if (tv->tv_sec > TV_SEC_MAX_VAL) {
+			error(0, 0, _("Warning: invalid tv_sec %ld s"), tv->tv_sec);
+			triptime = 0;
+		} else if (tv->tv_sec < 0) {
+			error(0, 0, _("Warning: time of day goes back (%ld s), taking countermeasures"), tv->tv_sec);
 			triptime = 0;
 			if (!rts->opt_latency) {
 				gettimeofday(tv, NULL);
 				rts->opt_latency = 1;
 				goto restamp;
 			}
+		} else {
+			triptime = tv->tv_sec * 1000000 + tv->tv_usec;
 		}
+
 		if (!csfailed) {
 			rts->tsum += triptime;
 			rts->tsum2 += (double)((long long)triptime * (long long)triptime);
--- a/iputils_common.h
+++ b/iputils_common.h
@@ -10,6 +10,9 @@
 	  !!__builtin_types_compatible_p(__typeof__(arr), \
 					 __typeof__(&arr[0]))])) * 0)
 
+/* 1000001 = 1000000 tv_sec + 1 tv_usec */
+#define TV_SEC_MAX_VAL (LONG_MAX/1000001)
+
 #ifdef __GNUC__
 # define iputils_attribute_format(t, n, m) __attribute__((__format__ (t, n, m)))
 #else


ping/ping_common.c
```diff
@@ -754,7 +754,25 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
+			tv->tv_usec = 999999;
+		}
+
+		if (tv->tv_usec < 0) {
+			error(0, 0, _("Warning: overflow tv_usec %ld us"), tv->tv_usec);
+			tv->tv_usec = 0;
+		}
+
+		/* 1000001 = 1000000 tv_sec + 1 tv_usec */
+		if (tv->tv_sec > LONG_MAX/1000001 || tv->tv_sec < -LONG_MAX/1000001) {
```
nmeyerhans, line 769: It might be nice to give a name to `LONG_MAX/1000001`, maybe with `#define`?

Zephkek: Something like would be nice:
```c
#define USEC_PER_SEC 1000000 
#define USEC_MAX (USEC_PER_SEC - 1) 
#define SEC_SAFE_MAX (LONG_MAX / (USEC_PER_SEC + 1))
```
These replace magic numbers and create a clear boundary for detecting integer overflow when converting time units.

pevik: Although the other 2 definitions make sense, I would prefer to postpone adding them later after this is fixed (it's an unrelated cleanup - `1000000` should be used on more places not just here).

pevik:
Please, when you're finish with your review, add your `Reviewed-by:` or `Acked-by:` tag.

ping/ping_common.c
```diff
@@ -754,7 +754,24 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
+			error(0, 0, _("Warning: underflow tv_usec %ld us"), tv->tv_usec);
+			tv->tv_usec = 999999;
+		}
+
+		if (tv->tv_usec < 0) {
+			error(0, 0, _("Warning: overflow tv_usec %ld us"), tv->tv_usec);
+			tv->tv_usec = 0;
+		}
+
+		if (tv->tv_sec > TV_SEC_MAX_VAL || tv->tv_sec < -TV_SEC_MAX_VAL) {
```
metan-ucw, line 768: Isn't tv->tv_sec < 0 invalid anyway? That would mean that the packet traveled back in time.

pevik: It makes sense, but IMHO this is handled later this part:

https://github.com/iputils/iputils/blob/3bb2d7302618d8c3708d3c93adcd90f342f5b6dc/ping/ping_common.c#L758-L766

I also wondered if I should move it to handle it via `tv->tv_sec < 0`. But how about `if (!rts->opt_latency) { ... }` part? Is it relevant for crafted RTT values as well?

metan-ucw: At the start of the `gather_statistics()` we do `tv_sub()` where we calculate the difference between the time we send the packet and the time we received a reply. We have no way knowing if we got negative value because of wall clock change or because of a crafted value.

And in the case of the crafted value the problem is even worse, I guess that if we send a timestamp that is ahead by hours ping will get stuck in the loop, trying to restamp it for hours consuming 100% of CPU time. It would make more sense to discard such sample from the statistics.

So I would do:

* Remove the restamp goto
* Check for negative value right in the tv_sec and set triptime to 0 if it was negative
* Skip the part where we add to the rts->tsum if triptime == 0

metan-ucw: Ah I was blind, we actually use rts->opt_latency to guard against infinite loop. However the restamping is still questionable, we are not getting a good sample by pretending it arrived a tiny bit later.

Also idea for a future, we should switch to `CLOCK_MONOTONIC` timer that is not going to go backwards unlike the wall clock.

Zephkek: Using `CLOCK_MONOTONIC` for send times seems practical initially, potentially obtaining a `timespec` but converting to `timeval` for the ICMP payload. This mainly improves the reliability of RTTs at the current microsecond precision by avoiding wall-clock issues.

For receive times, kernel monotonic `timespec` timestamps (via `SO_TIMESTAMPNS`/`SO_TIMESTAMPING`) would be ideal, with user-space `clock_gettime(CLOCK_MONOTONIC)` post-`recvmsg` as a fallback.

This isolates RTT from wall-clock changes, yielding more reliable results. Pursuing full nanosecond RTT precision (e.g., storing `timespec` in packets) could be a separate, subsequent step if needed.

pevik: > Also idea for a future, we should switch to `CLOCK_MONOTONIC` timer that is not going to go backwards unlike the wall clock.

+1. I suppose using `CLOCK_MONOTONIC` can wait after this fix (best would be if one of you could create a ticket for it). Also I would prefer to keep `resample` for now to keep the CVE fix minimal (postpone fix for later) unless the benefit pays for bigger change.

Therefore back to my original question what from the 3 points would you implement now? Only this?
* Check for negative value right in the tv_sec and set triptime to 0 if it was negative
(e.g. move whole content of  `if (triptime < 0) { .. }` into new `else if (tv->tv_sec < 0)`.

metan-ucw: No need for the else branch, we can do:

```
    if (tv->tv_sec > TV_SEC_MAX_VAL) {

    }

   if (tv->tv_sec < 0) {
   
   }
```

Apart from that it does sound like a plan to me. Let's limit the changes for this particular fix and then do a bigger cleanup once this is dealt with.

pevik:
Branch rebased.

ping/ping_common.c
pevik: The only thing is that keeping also negative: if (tv->tv_sec > TV_SEC_MAX_VAL || tv->tv_sec < -TV_SEC_MAX_VAL) {`

pevik: The only thing is that keeping also negative separate

```C
if (tv->tv_sec > TV_SEC_MAX_VAL || tv->tv_sec < -TV_SEC_MAX_VAL) {
    /* underflow or overflow => likely crafted */
} else if (triptime < 0) {
    /* 
     * Negative value but small enough to be valid - can be both crafted or time goes back
     * => I would keep the warning about time goes back.
     */
}
```

 e.g. to have final code
```C
restamp:
		tvsub(tv, &tmp_tv);

		if (tv->tv_usec >= 1000000) {
			error(0, 0, _("Warning: underflow tv_usec %ld us"), tv->tv_usec);
			tv->tv_usec = 999999;
		}

		if (tv->tv_usec < 0) {
			error(0, 0, _("Warning: overflow tv_usec %ld us"), tv->tv_usec);
			tv->tv_usec = 0;
		}

		if (tv->tv_sec > TV_SEC_MAX_VAL || tv->tv_sec < -TV_SEC_MAX_VAL) {
			error(0, 0, _("Warning: invalid tv_sec %ld s"), tv->tv_sec);
			triptime = 0;
		} else if (tv->tv_sec < 0) {
			error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), tv->tv_sec);
			triptime = 0;
			if (!rts->opt_latency) {
				gettimeofday(tv, NULL);
				rts->opt_latency = 1;
				goto restamp;
			}
		} else {
			triptime = tv->tv_sec * 1000000 + tv->tv_usec;
		}

		if (!csfailed) {
...
```
Or you meant something different?

metan-ucw: You left in the `|| tv->tv_sec < -TV_SEC_MAX_VAL` that shouldn't be needed because we without that we would end up in the if (tv->tv_sec < 0) branch. Otherwise it looks good.

metan-ucw: If we are not doing to do the multiplication in the case that `tv_sec < 0` then there is no point in checking for underflow and no point in checking some negative numbers differently. At least that is my reasoning why there is no need to treat a subset of negative numbers differently.

nmeyerhans:
> Please, when you're finish with your review, add your `Reviewed-by:` or `Acked-by:` tag.

Acked in c03bb2744966266ff1ef255c9220e7759243e803 in my fork

Zephkek (changes requested):
There's an integer overflow vulnerability when large values are provided for both packet size (-s) and preload (-l) parameters. When multiplying `alloc * rts->preload`, the result can exceed INT_MAX without any bounds checking.

[rcvbuf = hold = alloc * rts->preload;](https://github.com/pevik/iputils/blob/3c7b468a3ba6179e172c4dbfd402d2c649301800/ping/ping_common.c#L451-L452)

This results in incorrect socket buffer allocation and ~99.99% packet loss. Looking at setup() this doesn't have overflow protection like the interval/preload.

Recommend adding bounds checking before multiplication:
```c
if (alloc > INT_MAX / rts->preload) {
    error(0, 0, _("buffer size overflow, reduce packet size or preload"));
    hold = INT_MAX; // Or another reasonable maximum
} else {
    hold = alloc * rts->preload;
}
```
**Reproduction:**
Using large values for both packet size and preload:
```
$ sudo ./ping/ping -l 65536 -s 30000 localhost
../ping/ping_common.c:451:24: runtime error: signed integer overflow: 45348 * 65536 cannot be represented in type 'int'
```
image

Zephkek:
Looking further at ping_common I discovered there's a similar overflow issue with the -W option. Interestingly, this one only happens when you set -c 1 (sending just one packet). If you set -W to values > 2147, it causes another integer overflow when converting the time:

```
$ sudo ./ping/ping -c 1 -W 2148 8.8.8.8
../ping/ping_common.c:269:37: runtime error: signed integer overflow: 2148000 * 1000 cannot be represented in type 'int'
```

looks like the bug is in `__schedule_exit()` where it does `waittime = global_rts->lingertime * 1000`. This function is called during ping's exit sequence, but only when -c 1 is used because it immediately enters the exit path after the first packet. With -c 2 or more, ping never hits this code path in the same way. The -W value gets multiplied by 1000 once during parsing to convert to ms, then multiplied by 1000 again here to convert to μs. Could be fixed by using a long long or just adding another bounds check perhaps?

ping/ping_common.c
pevik: OK, makes sense, therefore modified. Ready to be reviewed, hopefully final version.

I just kept `(tv->tv_sec < 0)` in `else if` to make sure `triptime = tv->tv_sec * 1000000 + tv->tv_usec;` assignment in final `else` is really done only if `tv->tv_sec` is valid.

I also unified `tv->tv_usec` error messages.

```c
if (tv->tv_usec >= 1000000) {
	error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
	tv->tv_usec = 999999;
}

if (tv->tv_usec < 0) {
	error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
	tv->tv_usec = 0;
}

if (tv->tv_sec > TV_SEC_MAX_VAL) {
	error(0, 0, _("Warning: invalid tv_sec %ld s"), tv->tv_sec);
	triptime = 0;
} else if (tv->tv_sec < 0) {
	error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), tv->tv_sec);
	triptime = 0;
	if (!rts->opt_latency) {
		gettimeofday(tv, NULL);
		rts->opt_latency = 1;
		goto restamp;
	}
} else {
	triptime = tv->tv_sec * 1000000 + tv->tv_usec;
}
```

pevik:
> > Please, when you're finish with your review, add your `Reviewed-by:` or `Acked-by:` tag.
> 
> Acked in [c03bb27](https://github.com/iputils/iputils/commit/c03bb2744966266ff1ef255c9220e7759243e803) in my fork

@nmeyerhans Thank you! Can you please quickly check final changes? Diff against the version you acked below. May I add your `Reviewed-by:` now?
Also feel free to propose better error messages for new strings (NOTE: later after the release I would change `%ldus` => `%ld us`, that's why it's now the difference). 
```diff
+++ ping/ping_common.c
@@ -756,30 +756,28 @@ restamp:
 		tvsub(tv, &tmp_tv);
 
 		if (tv->tv_usec >= 1000000) {
-			error(0, 0, _("Warning: underflow tv_usec %ld us"), tv->tv_usec);
+			error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
 			tv->tv_usec = 999999;
 		}
 
 		if (tv->tv_usec < 0) {
-			error(0, 0, _("Warning: overflow tv_usec %ld us"), tv->tv_usec);
+			error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
 			tv->tv_usec = 0;
 		}
 
-		if (tv->tv_sec > TV_SEC_MAX_VAL || tv->tv_sec < -TV_SEC_MAX_VAL) {
+		if (tv->tv_sec > TV_SEC_MAX_VAL) {
 			error(0, 0, _("Warning: invalid tv_sec %ld s"), tv->tv_sec);
 			triptime = 0;
-		} else {
-			triptime = tv->tv_sec * 1000000 + tv->tv_usec;
-		}
-
-		if (triptime < 0) {
-			error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), triptime);
+		} else if (tv->tv_sec < 0) {
+			error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), tv->tv_sec);
 			triptime = 0;
 			if (!rts->opt_latency) {
 				gettimeofday(tv, NULL);
 				rts->opt_latency = 1;
 				goto restamp;
 			}
+		} else {
+			triptime = tv->tv_sec * 1000000 + tv->tv_usec;
 		}
 
 		if (!csfailed) {

**Review by Zephkek — APPROVED** [review]
Other than the formatting for tv_sec looks good to me.

Reviewed-by: Mohamed Maatalllah <hotelsmaatallahrecemail@gmail.com>

**Comment by pevik** [comment]
> Looking further at ping_common I discovered there's a similar overflow issue with the -W option. Interestingly, this one only happens when you set -c 1 (sending just one packet). If you set -W to values > 2147, it causes another integer overflow when converting the time:
> 
> ```
> $ sudo ./ping/ping -c 1 -W 2148 8.8.8.8
> ../ping/ping_common.c:269:37: runtime error: signed integer overflow: 2148000 * 1000 cannot be represented in type 'int'
> ```
> 
> looks like the bug is in `__schedule_exit()` where it does `waittime = global_rts->lingertime * 1000`. This function is called during ping's exit sequence, but only when -c 1 is used because it immediately enters the exit path after the first packet. With -c 2 or more, ping never hits this code path in the same way from what it looks like. The -W value gets multiplied by 1000 once during parsing to convert to ms, then multiplied by 1000 again here to convert to μs. Could be fixed by using a long long or just adding another bounds check perhaps?

Good catch, this will need to be handled as well (I might do it in a separate PR).

**Comment by pevik** [comment]
> ```
> $ sudo ./ping/ping -l 65536 -s 30000 localhost
> ../ping/ping_common.c:451:24: runtime error: signed integer overflow: 45348 * 65536 cannot be represented in type 'int'
> ```
And [this one](https://github.com/iputils/iputils/pull/585#pullrequestreview-2820034501) as well. Thanks for your reports.

**Comment by pevik** [comment]
> Other than the formatting for tv_sec this part looks good to me.

I miss this, feel free to point what you mean.

> Reviewed-by: Mohamed Maatalllah [hotelsmaatallahrecemail@gmail.com](mailto:hotelsmaatallahrecemail@gmail.com)

Thanks, added (Maatalllah => Maatallah)

**Comment by Zephkek** [comment]
As for the formatting im talking about the last error message where tv_sec is logged with format %ldus 
since its seconds thought should be %lds

**Comment by pevik** [comment]
> As for the formatting im talking about the last error message where tv_sec is logged with format %ldus
> since its seconds thought should be %lds

OK, updated. I'm surprised, that previous `-7256972569576721232us` got changed into `-4243165695442945 s`: i.e. only 1000x smaller (I would expect it 1000000x).

File: ping/ping_common.c [hunk 3]
```diff
@@ -754,16 +754,32 @@ int gather_statistics(struct ping_rts *rts, uint8_t *icmph, int icmplen,
...
 				rts->opt_latency = 1;
 				goto restamp;
 			}
+		} else {
+			triptime = tv->tv_sec * 1000000 + tv->tv_usec;
 		}
+
 		if (!csfailed) {
 			rts->tsum += triptime;
 			rts->tsum2 += (double)((long long)triptime * (long long)triptime);
```
metan-ucw, line 785: So we still update the statistics even if `triptime == 0` I guess that this is okay for a quick fix but should be fixed later.

metan-ucw:
The commit description talks about limiting the tv_sec to `<-LONG_MAX/1000000, LONG_MAX/1000000>`  but now we limit it to `<0, LONG_MAX/1000000>` , other than that the patch looks good.

metan-ucw (approved):

ping/ping_common.c
pevik: Is it really wrong? `2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 101ms` looks correct to me (+2 duplicates).

```
$ ping 127.0.0.1 -i0.1 -c2 # run with python script doing crafting on master
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.065 ms
../ping/ping_common.c:757:25: runtime error: signed integer overflow: -1157944657838082 * 1000000 cannot be represented in type 'long int'
./ping/ping: Warning: time of day goes back (-2310396749805860133us), taking countermeasures
./ping/ping: Warning: time of day goes back (-2310396749805859590us), taking countermeasures
24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
./ping/ping: Warning: time of day goes back (-4708252664467859731us), taking countermeasures
24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.119 ms

2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 101ms

$ ping 127.0.0.1 -i0.1 -c2 # run with python script doing crafting on this PR
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.064 ms
./ping/ping: Warning: invalid tv_usec -6510615555425847804 us
./ping/ping: Warning: time of day goes back (-53614076755970 s), taking countermeasures
./ping/ping: Warning: invalid tv_usec -6510615555425847641 us
./ping/ping: Warning: time of day goes back (-53614076755970 s), taking countermeasures
24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
./ping/ping: Warning: invalid tv_usec -6510615555425803771 us
./ping/ping: Warning: time of day goes back (-347681897578498 s), taking countermeasures
24 bytes from 127.0.0.1: icmp_seq=1 ttl=64 (truncated)
64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.161 ms

2 packets transmitted, 2 received, +2 duplicates, 0% packet loss, time 101ms
rtt min/avg/max/mdev = 0.000/0.056/0.161/0.065 ms
```

If it's really a bug, it would not be introduced by this. Trying to just `return 1` on wrong `tv_sec` leads to obviously wrong:

```
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.061 ms
./ping/ping: Warning: invalid tv_usec -6510615555425219887 us
./ping/ping: Warning: time of day goes back (-2799219165364226 s), taking countermeasures

1 packets transmitted, 2 received, -100% packet loss, time 0ms
rtt min/avg/max/mdev = 0.061/0.030/0.061/0.030 ms
```

Or you mean skip on `triptime == 0`? (I don't think it's correct)
```
-               if (!csfailed) {
+               if (!csfailed && !triptime) {
```

nmeyerhans:
updated signoff in b14620228d707884e6f48454b62f70dfd743c9e7

nmeyerhans (approved):

Zephkek (changes requested):
Found a tricky integer overflow in ping's RTT calculation. When receiving crafted ICMP replies with large timestamps (~268s), the RTT accumulates until it exceeds INT_MAX and wraps to negative. This becomes a problem only in adaptive mode because update_interval() uses this negative RTT value to calculate interval, resulting in negative intervals which causes the main loop to run without delay.

The issue is specifically here 
```c
rts->rtt += triptime - rts->rtt / 8;
```

to trigger this overflow, i made this script to craft the rtt values: 
https://gist.github.com/Zephkek/e8697cdbcc190dde673bcba0b18f5710

then ran ping in adaptive mode:
```
sudo ./ping/ping -A 127.0.0.1
```
I also added a few debug prints because ASan & UBSan issues..:
```
DEBUG AFTER RTT UPDATE: new rtt=2147299198, addition=268864057, division=268412399
64 bytes from 127.0.0.1: icmp_seq=56 ttl=64 time=268864 ms (DUP!)
DEBUG BEFORE RTT UPDATE: triptime=268933983, current rtt=2147299198
DEBUG AFTER RTT UPDATE: new rtt=-2147146514, addition=268933983, division=-268393314
64 bytes from 127.0.0.1: icmp_seq=57 ttl=64 time=268934 ms (DUP!)
DEBUG BEFORE RTT UPDATE: triptime=268994142, current rtt=-2147146514
DEBUG AFTER RTT UPDATE: new rtt=-1609759058, addition=268994142, division=-201219882
64 bytes from 127.0.0.1: icmp_seq=58 ttl=64 time=268994 ms (DUP!)
DEBUG BEFORE RTT UPDATE: triptime=268064302, current rtt=-1609759058
DEBUG AFTER RTT UPDATE: new rtt=-1140474874, addition=268064302, division=-142559359
64 bytes from 127.0.0.1: icmp_seq=59 ttl=64 time=268064 ms (DUP!)
DEBUG BEFORE RTT UPDATE: triptime=268124126, current rtt=-1140474874
DEBUG AFTER RTT UPDATE: new rtt=-729791389, addition=268124126, division=-91223923
64 bytes from 127.0.0.1: icmp_seq=60 ttl=64 time=268124 ms (DUP!)
DEBUG BEFORE RTT UPDATE: triptime=268194229, current rtt=-729791389
DEBUG AFTER RTT UPDATE: new rtt=-370373237, addition=268194229, division=-46296654
64 bytes from 127.0.0.1: icmp_seq=61 ttl=64 time=268194 ms (DUP!)
```
This in turn will cause interval to be negative, thus next to be negative thus stuck in this loop:
```c
		 do {
			next = pinger(rts, fset, sock);
			next = schedule_exit(rts, next);
		} while (next <= 0);		
```

Weirdly, neither ASAn nor UBSan catch this...

https://github.com/user-attachments/assets/2666acca-c81e-405d-adfc-d6e070f09e28

pevik:
Merged this as 070cfacd7348386173231fb16fad4983d4e6ae40, thanks all to review.
I'll try to fix the other 3 issues next week. 

@Zephkek It'd be better to add reports into separate issues or at https://github.com/iputils/iputils/issues/584 (which should be reopened then), but it's ok to keep it here (fix is more important than the formality).

ping/ping_common.c
pevik: > Also idea for a future, we should switch to `CLOCK_MONOTONIC` timer that is not going to go backwards unlike the wall clock.

Created an issue for it: https://github.com/iputils/iputils/issues/587.

yvs2014 (commented):
<details>
 <summary>changes around restamp</summary>
 
```
restamp:
                tvsub(tv, &tmp_tv);
-                 triptime = tv->tv_sec * 1000000 + tv->tv_usec;
-                 if (triptime < 0) {
-                         error(0, 0, _("Warning: time of day goes back (%ldus), taking countermeasures"), triptime);
-                         triptime = 0;
-                         if (!rts->opt_latency) {
-                                 gettimeofday(tv, NULL);
-                                 rts->opt_latency = 1;
-                                 goto restamp;
-                         }
-                 }
 
+                 if (tv->tv_usec >= 1000000) {
+                         error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
+                         tv->tv_usec = 999999;
+                 }
+ 
+                 if (tv->tv_usec < 0) {
+                         error(0, 0, _("Warning: invalid tv_usec %ld us"), tv->tv_usec);
+                         tv->tv_usec = 0;
+                 }
+ 
+                 if (tv->tv_sec > TV_SEC_MAX_VAL) {
+                         error(0, 0, _("Warning: invalid tv_sec %ld s"), tv->tv_sec);
+                         triptime = 0;
+                 } else if (tv->tv_sec < 0) {
+                         error(0, 0, _("Warning: time of day goes back (%ld s), taking countermeasures"), tv->tv_sec);
+                         triptime = 0;
+                         if (!rts->opt_latency) {
+                                 gettimeofday(tv, NULL);
+                                 rts->opt_latency = 1;
+                                 goto restamp;
+                         }
+                 } else {
+                         triptime = tv->tv_sec * 1000000 + tv->tv_usec;
+                 }
```

</details>
If I got it right: the purpose of (two-pass max) restamp section is to set triptime value >= 0 mandatory, plus to set latency option (to not call gettimeofday() further) if something is incorrect with triptime at first pass.
<br><br>
In both cases we have `triptime` value >= 0, but it's a bit different with some timevals:
let's say tv(.sec=12345 .usec=0) and tmp_tv(.sec=12346, .usec=-1500000), then tvsub() returns tv(.sec=0, .usec=-500000), and so that, there's triptime=0 and no "time of day goes back" message and latency option is not set with this example. Or I got it incorrectly?

Zephkek:
My fault for not spotting this bypass path during the review. I only recently had time to dig into the code regarding the zero-timestamp interaction over the weekend - it seems the check on the combined triptime magnitude before squaring is the missing piece.
Details on the bypass are in the new issue: [589](https://github.com/iputils/iputils/issues/589)

PR #590, opened 2025-05-13
SECURITY: communicate how security issues should be handled

okias:
@pevik

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,36 @@
+# Security Policy
+
+## Reporting a vulnerability
+
+Please put detailed report to
+https://github.com/iputils/iputils/security/advisories
+or send a detailed mail to pvorel@suse.cz and david@ixit.cz
+report vulnerabilities in iputils.
+
+Even when unsure whether the bug in question is an exploitable
+vulnerability, it is recommended to send the report to
+https://github.com/iputils/iputils/security/advisories or/and
+pvorel@suse.cz and david@ixit.cz (and obviously not to discuss the
+issue anywhere else).
+
+Vulnerabilities are expected to be discussed _only_ there, and not in public,
+until the official announcement.
+
+Examples for details to include:
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Reporting a vulnerability
+
+Please send a detailed mail to pvorel@suse.cz and david@ixit.cz
+report vulnerabilities in iputils.
+
+Even when unsure whether the bug in question is an exploitable
+vulnerability, it is recommended to send the report to
+pvorel@suse.cz and david@ixit.cz (and obviously not to discuss the
+issue anywhere else).
```
pevik, line 11: Thank you for doing this! 

How about trying to set security advisories: https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository. That way we would have both privacy and github tooling, e.g. [**button as is on other projects**](https://github.com/gphoto/libgphoto2/security/advisories) which opens a [**report form**](https://github.com/gphoto/libgphoto2/security/advisories/new)  (similar to issues form but private). I suppose all project members could read it.

Or do you think it would just invite AI generators to flood us with us with invalid reports?

okias: I'm open to whatever makes your (our) life easier, but I would prefer don't lock-in into GitHub ecosystem (so there would be easy possibility to migrate to gitlab/codeberge/forgejo etc.)

pevik: I would not worry too much about lock-in - if we decide to move, the old ones will be archived and we just start to use a different channel. Let me think twice before deciding this (I'll let you know during today).

pevik (approved):
Thank you!
```
```

pevik:
@okias I dared to replace `@` with ` at ` and merged. I also deleted the remote branch. Thank you!

PR #591, opened 2025-05-15
ping: Fix ipv4 ttl value when using SOCK_DGRAM on big endian systems

radolin:
When testing on a big endian (IBM s390) system I recently noticed that when pinging ipv4 addresses and using the SOCK_DGRAM interface the value returned for the ttl is always 0, for example:

`s390init03:~ # ping 127.0.0.1
64 bytes from 127.0.0.1: icmp_seq=1 ttl=0 time=0.054 ms
64 bytes from 127.0.0.1: icmp_seq=2 ttl=0 time=0.056 ms
64 bytes from 127.0.0.1: icmp_seq=3 ttl=0 time=0.114 ms
^C
3 packets transmitted, 3 received, 0% packet loss, time 2097ms
rtt min/avg/max/mdev = 0.054/0.074/0.114/0.027 ms`

I have traced it back to a commit fixing GCC warnings, which cast the value returned by CMSG_DATA to a uint8_t, hence losing the actual useful data on big endian systems.

### DIFF
--- a/ping/ping.c
+++ b/ping/ping.c
@@ -1642,7 +1642,7 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 	int csfailed;
 	struct cmsghdr *cmsgh;
 	int reply_ttl;
-	uint8_t *opts, *tmp_ttl;
+	uint8_t *opts;
 	int olen;
 	int wrong_source = 0;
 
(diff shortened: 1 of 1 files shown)


pevik:
Hm, looking at 7e7ffffacc2af71ff87a252783d0301f39f9a87b, maybe `ttl = *(int *) CMSG_DATA(cmsg);` worked but produced warnings. `memcpy()` is probably the best.

pevik:
7e7ffffacc2af71ff87a252783d0301f39f9a87b was released in s20161105, I wonder that nobody noticed error on s390x since then.

metan-ucw:
We have to use the `mempcy()` here because the data field cannot be assumed to be suitably aligned (see CMSG_DATA() in man cmsg).

pevik:
Fixed in c9c82b4576f0b616793ffbdc815c02e2e4da1f5c, thanks!
(I forget to add `Closes: https://github.com/iputils/iputils/pull/591` to the commit message.)
I was not sure if even the original 87dbb3a5db657d5eae6934707beaf0507980a1c3 (from s20150815) was problematic, but really the casting in 7e7ffffacc2af71ff87a252783d0301f39f9a87b (from s20161105) caused the problem.

PR #599, opened 2025-06-17
Unaligned memory access in call to gettimeofday, function build_echo (SIGBUS crash)

### LINKED ISSUE #598: Unaligned memory access in call to gettimeofday, function build_echo (SIGBUS crash)
**Description**
There's a memory alignment issue in ping6_common.c, build_echo() which causes ping to IPv6 addresses to crash with SIGBUS error on SPARC architecture.

Problematic line:
**gettimeofday((struct timeval *)&_icmph[8], NULL);**

There’s no guarantee that _icmph + 8 is 8-byte aligned.
On SPARC and other strict-alignment architectures, this results in a SIGBUS (BUS_ADRALN) when gettimeofday() writes to that unaligned address.

Note that x86 _allows_ unaligned access, so it does not crash, but unaligned access _is_ slower than aligned access, so worth fixing IMO.

**Reproduce (on SPARC):** 
```
$ ping ::1

_PING ::1 (::1) 56 data bytes
Bus error

(gdb) 
...
Program received signal SIGBUS, Bus error.
0xfff00001009d1824 in gettimeofday () from /usr/lib64/libc.so.6
```

**Proposed fix**
See PR #600 

The PR fix is confirmed, and working on SPARC.

**System information**
Distro: Gentoo Linux
Kernel: Linux 6.15.0
Hardware: Sun Fire V240

shalseth:
There's a memory alignment issue in ping6_common.c, build_echo() which causes ping to IPv6 addresses to crash with SIGBUS error on SPARC architecture.

Avoid unaligned access by using a local struct timeval and copying it with memcpy.

For details, see issue: https://github.com/iputils/iputils/issues/598

Fixes #598

### DIFF
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -601,8 +601,11 @@ int build_echo(struct ping_rts *rts, uint8_t *_icmph,
 	icmph->icmp6_seq = htons(rts->ntransmitted + 1);
 	icmph->icmp6_id = rts->ident;
 
-	if (rts->timing)
-		gettimeofday((struct timeval *)&_icmph[8], NULL);
+        if (rts->timing) {
+	        struct timeval tv;
+	        gettimeofday(&tv, NULL);
+	        memcpy(&_icmph[8], &tv, sizeof(tv));
+        }
 
 	cc = rts->datalen + 8;			/* skips ICMP portion */
 


shalseth:
Creating new pull request with signed commit

PR #600, opened 2025-06-17
Unaligned memory access in call to gettimeofday, function build_echo (SIGBUS crash)

### LINKED ISSUE #598: Unaligned memory access in call to gettimeofday, function build_echo (SIGBUS crash)

shalseth:
There's a memory alignment issue in ping6_common.c, build_echo() which causes ping to IPv6 addresses to crash with SIGBUS error on SPARC architecture

See #598 for details

Resolves #598

### DIFF
--- a/ping/ping6_common.c
+++ b/ping/ping6_common.c
@@ -601,8 +601,11 @@ int build_echo(struct ping_rts *rts, uint8_t *_icmph,
 	icmph->icmp6_seq = htons(rts->ntransmitted + 1);
 	icmph->icmp6_id = rts->ident;
 
-	if (rts->timing)
-		gettimeofday((struct timeval *)&_icmph[8], NULL);
+        if (rts->timing) {
+	        struct timeval tv;
+	        gettimeofday(&tv, NULL);
+	        memcpy(&_icmph[8], &tv, sizeof(tv));
+        }
 
 	cc = rts->datalen + 8;			/* skips ICMP portion */
 


shalseth:
Sure thing, does it look good now? Just let me know if you want me to make any other changes.

pevik:
> Sure thing, does it look good now? Just let me know if you want me to make any other changes.

Code is OK. Ideally you would put all info which is in this PR description also to the git commit message. One day github may disappear (or we move elsewhere, i.e. gitlab.com), therefore I value `git log` history.

metan-ucw:
I'm not sure if this is a complete fix though. The main problem here is that we define the buffer for the packet as a part of the `struct ping_rts` and it's defined as a `char` hence it's alignment may be at a byte granularity, which may break much more than just the timestamp. We cannot really even cast that safely to an integer either. Complete solution is probably to add `aligned(8)` attribute to the `outpack[]` array in the `struct ping_rts` definition. If I'm looking right at the code that should fix this issue as well.

pevik:
@metan-ucw Would you please send an alternative fix as new PR?

pevik:
I suppose he will push the fix with your `Reported-by: ...` and `Tested-by: ...` tag, but not sure.

Issue #605, opened 2025-09-01
ping broken output due JSON support changes


pevik:
e6b25a6225a2b71e408be87d5dde6ec25f3f0aa3 from https://github.com/iputils/iputils/pull/571 brought a regression:

`ping -c1 ::1` previously show:

```diff
-PING ::1 (::1) 56 data bytes
+PING ::1 (0.0.0.0) 56(84) bytes of data.
```

Obviously `printf(_("PING %s (%s) "), rts->hostname, target);` in [`ping_print_packet()`](https://github.com/iputils/iputils/commit/e6b25a6225a2b71e408be87d5dde6ec25f3f0aa3#r164952623) is wrong (must be `::1`, not `0.0.0.0`).

In the code it should not have been `target`, but `inet_ntoa(rts->whereto.sin_addr)` for IPv4 and `pr_raw_addr(rts, &rts->whereto6, sizeof rts->whereto6)` for IPv6.

Also both outputs are slightly different `"%d(%d) bytes of data.\n"` (IPv4) vs. `"%d data bytes\n"` (IPv6). I wonder if we can finally dare to unify them (which one to use)?

There might be more errors like that, but I have just a few tests in my fork. FYI I'm testing via [`testing` branch in my fork](https://github.com/pevik/iputils/commits/refs/heads/testing/).

PR #606, opened 2025-09-02
[WIP] ping: Fix regression from JSON support, improve JSON

pevik:
First draft of fix for https://github.com/iputils/iputils/issues/605 (broken non-JSON output for IPv4).

The fix itself is ready (first commit). If we aim to get a quick fix it could be merged separately.

```diff
-PING ::1 (0.0.0.0) 56(84) bytes of data. # wrong
+PING ::1 (::1) 56 data bytes # fixed
```

Then there is an attempt to add to JSON data length values (numbers `56` and `84`) from ping header in the second commit:
```
$ ping 127.0.0.1 # IPv4
PING 127.0.0.1 (127.0.0.1) 56 (84) data bytes
...
$ ping ::1 # IPv6
PING ::1 (::1) 56 data bytes
...
```
RFC: The last commit is touching non-JSON ping output - changing IPv4 `"%d(%d) bytes of data."` string to `"%d (%d) data bytes"` (to unify with  `"%d data bytes"` IPv6 string. I'm not sure myself if it's a good idea. It's dangerous to touch output because there might be some broken scripts. And it will take years when people start using JSON output (some newer if aim for busybox ping implementation compatibility). And if used, maybe put a description for second number?

TODO: there is missing source and target JSON (the part which has been fixed in the first commit).

### DIFF
--- a/ping/ping_common.c
+++ b/ping/ping_common.c
@@ -834,6 +834,17 @@ restamp:
 		uint8_t *cp, *dp;
 
 		print_timestamp(rts);
+
+		if (rts->opt_json) {
+			// TODO: improve key name
+			construct_json(rts, PING_JSON_INT, "data_bytes", rts->datalen);
+			/*
+			 * FIXME: Add this require either add ping_func_set_st *fset or move into ping_print_packet().
+			if (fset->ping_data_bytes)
+				construct_json(rts, PING_JSON_INT, "data_bytes_full", fset->ping_data_bytes(rts));
+			*/
+		}
+
 		ping_print_int(rts, _("%d bytes "), "bytes", cc);
 		// TODO: split host IP address and name into separate attributes in JSON?
 		ping_print_str(rts, _("from %s:"), "host", from);


metan-ucw (commented):
Second  commit:

So for ipv4 we have header in the received data bytes right? So it may make sense to split the JSON report into `data_size` and `header_size` and `packet_size` where `packet_size = data_size + header_size`?

metan-ucw (approved):
First commit:

Looks good to me.

metan-ucw (commented):
Third commit:

I wouldn't touch the human readable output like this at all, this is calling for a trouble.

pevik:
> Third commit:
> 
> I wouldn't touch the human readable output like this at all, this is calling for a trouble.

Fair enough, I removed the commit.

pevik:
> Second commit:
> 
> So for ipv4 we have header in the received data bytes right? So it may make sense to split the JSON report into `data_size` and `header_size` and `packet_size` where `packet_size = data_size + header_size`?

Probably. The first line is about ICMP echo request size + payload size. I need to look up ICMP vs. ICMPv6 differences (also header and payload IPv4 vs. IPv6 differences may apply), because there is a reason why ping IPv4 shows 2 numbers and IPv6 only one. I'll get to this, but maybe first I'd like to get merged partly improved tests (so I don't need to keep rebasing it and it helps to have code which is compatible for older releases before we do more changes to the code).

PR #607, opened 2025-09-03
Improved testing

pevik:
Improve testing of tools' output.

Partly rewritten python test code to perl based Test::Command framework.  That allows to compare tools output, not just exit code and describes better the output. Also it can be run standalone (without meson).
    
Covered mostly ping (but still needs to be improved).
TODO: covert all python tests.
    
Optional parameter allows to pass binary to be tested (path to binary would be better, but it'd have to be two directories: builddir/:builddir/ping/).

ping-01-basics.t mocks getaddrinfo() code in ping.

Other changes:
* "error: Unify output of error() implementation with glibc" required to fix test output
* "meson-build-dist-man.sh: Skip tests on building doc"

### DIFF
--- a/test/ping/ping-02-errors.t
+++ b/test/ping/ping-02-errors.t
@@ -1,0 +1,73 @@
+#!/usr/bin/perl -w
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2025 Petr Vorel <pvorel@suse.cz>
+
+use Test::Command tests => 6;
+use Test::More;
+
+my $lib = File::Basename::dirname(Cwd::abs_path($0)) . '/../lib.pl';
+require "$lib";
+
+my $ping = get_cmd($ARGV[0] // 'ping');
+
+# no arg
+{
+	my $cmd = Test::Command->new(cmd => "$ping");
+	$cmd->exit_is_num(2);
+	subtest 'output' => sub {
+		$cmd->stderr_is_eq("$ping: usage error: Destination address required\n");
+	}
+}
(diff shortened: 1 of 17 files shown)


test/ping/ping-02-errors.t
```diff
@@ -0,0 +1,73 @@
...
+	my $cmd = Test::Command->new(cmd => "$ping", env => { LC_ALL => 'C', LANG   => 'C' });
+	$cmd->exit_is_num(2);
+	subtest 'output' => sub {
+		$cmd->stderr_is_eq("$ping: usage error: Destination address required\n");
+	}
+}
+
+# -c1 -i0.001 127.0.0.1
+{
+	my $cmd = Test::Command->new(cmd => "$ping -c1 -i0.001 127.0.0.1", env => { LC_ALL => 'C', LANG   => 'C' });
```
metan-ucw, line 24: Shouldn't the env with `LC_ALL='C'` be passed in all tests? The `ping-01-basic.t` does not seem to do that. Or we can set these once before all the tests so that we do not need to repeat them in each test that is matching the output.

metan-ucw:
The coverage is very basic but it looks like the testing is moving into a right direction.

metan-ucw (changes requested):

test/ping/ping-02-errors.t
pevik: >  I'll put setup into [`test/lib.pl`](https://github.com/pevik/iputils/blob/testing/test/lib.pl)...

PR #612, opened 2025-11-27
tracepath: Add ICMP extensions support

daniellerts:
RFC 4884 defines a general framework that allows selected ICMP and ICMPv6 error messages to carry additional extension data.

It does so by allowing the sender to encode the length of the "original datagram" field in the ICMP header, and defining an ICMP extension structure that can encapsulate one or more ICMP extension objects, enabling arbitrary metadata to accompany the error message.

RFC 5837 builds on this framework and introduces the Interface Information Object (IIO).

This extension object provides interface level visibility into the path a packet takes by allowing routers and hosts to include:
- The interface that received the packet which triggered the ICMP error.
- The interface out of which the packet would have been forwarded (if
  forwardable).
- The nexthop IP address that would have been used for forwarding.

These extensions help diagnose forwarding decisions, MTU issues, and other routing-related problems. When ICMP extensions are present, they are returned via the socket’s error queue appended after the UDP probe payload.

Add support for ICMP extensions, by enabling receipt of RFC 4884 extended ICMP error messages when the user requests extension output via a flag, and implement complete parsing and display of RFC 5837 interface information.

Patchset overview:
Patches #1-#2: Preparations
Patch #3: Add ICMP extensions support

### DIFF
--- a/tracepath.c
+++ b/tracepath.c
@@ -18,6 +18,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <resolv.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -157,6 +158,7 @@ static int recverr(struct run_state *const ctl)
 		.msg_controllen = sizeof(cbuf),
 		0
 	};
+	bool restart;
 
  restart:
 	memset(&rcvbuf, -1, sizeof(rcvbuf));
@@ -298,23 +300,27 @@ static int recverr(struct run_state *const ctl)
 	else
 		rethops = 256 - rethops;
 
+	restart = false;
+
 	switch (e->ee_errno) {
 	case ETIMEDOUT:
 		printf("\n");
+		restart = true;
 		break;
 	case EMSGSIZE:
 		printf(_("pmtu %d\n"), e->ee_info);
 		ctl->mtu = e->ee_info;
 		progress = ctl->mtu;
+		restart = true;
 		break;
 	case ECONNREFUSED:
 		printf(_("reached\n"));
 		ctl->hops_to = sndhops < 0 ? ctl->ttl : sndhops;
 		ctl->hops_from = rethops;
-		return 0;
+		break;
 	case EPROTO:
 		printf("!P\n");
-		return 0;
+		break;
 	case EHOSTUNREACH:
 		if ((e->ee_origin == SO_EE_ORIGIN_ICMP &&
 		     e->ee_type == ICMP_TIME_EXCEEDED &&
@@ -328,22 +334,26 @@ static int recverr(struct run_state *const ctl)
 					printf(_("asymm %2d "), rethops);
 			}
 			printf("\n");
+			restart = true;
 			break;
 		}
 		printf("!H\n");
-		return 0;
+		break;
 	case ENETUNREACH:
 		printf("!N\n");
-		return 0;
+		break;
 	case EACCES:
 		printf("!A\n");
-		return 0;
+		break;
 	default:
 		printf("\n");
 		error(0, e->ee_errno, _("NET ERROR"));
-		return 0;
+		break;
 	}
-	goto restart;
+
+	if (restart)
+		goto restart;
+	return 0;
 }
 
 static int probe_ttl(struct run_state *const ctl)
--- a/tracepath.c
+++ b/tracepath.c
@@ -36,6 +36,8 @@
 #include <linux/errqueue.h>
 #include <linux/icmp.h>
 #include <linux/icmpv6.h>
+#include <linux/in.h>
+#include <linux/in6.h>
 #include <linux/types.h>
 
 #include "iputils_common.h"
@@ -102,8 +104,36 @@ struct run_state {
 	unsigned int
 		no_resolve:1,
 		show_both:1,
-		mapped:1;
+		mapped:1,
+		show_extensions:1;
 };
+
+#ifdef SO_EE_RFC4884_FLAG_INVALID
+enum {
+	ICMP_EXT_OBJ_CLASS_IIO = 2,
+};
+
+/*
+ * RFC 5837: Interface Information Object
+ */
+#define ICMP_EXT_CTYPE_IIO_ROLE(ROLE)	(((ROLE) >> 6) & 0x03)
+#define ICMP_EXT_CTYPE_IIO_MTU		(1U << (0))
+#define ICMP_EXT_CTYPE_IIO_NAME		(1U << (1))
+#define ICMP_EXT_CTYPE_IIO_IPADDR	(1U << (2))
+#define ICMP_EXT_CTYPE_IIO_IFINDEX	(1U << (3))
+
+#ifndef ICMP_AFI_IP
+#define ICMP_AFI_IP	1
+#endif
+#ifndef ICMP_AFI_IP6
+#define ICMP_AFI_IP6	2
+#endif
+
+struct icmp_ext_iio_name_subobj {
+	uint8_t len;
+	char name[];
+};
+#endif
 
 /*
  * All includes, definitions, struct declarations, and global variables are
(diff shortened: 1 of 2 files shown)


pevik:
@daniellerts I finally find some time to review. Could you rebase and force push (`git push -f`)?

tracepath.c
```diff
@@ -328,22 +509,38 @@ static int recverr(struct run_state *const ctl)
...
 		printf("\n");
 		error(0, e->ee_errno, _("NET ERROR"));
-		return 0;
+		break;
+	}
+
+	if (ctl->show_extensions) {
+#ifdef SO_EE_RFC4884_FLAG_INVALID
+		if (!(e->ee_rfc4884.flags & SO_EE_RFC4884_FLAG_INVALID)) {
+			if (e->ee_rfc4884.len)
```
pevik, line 539: On my network I always get `e->ee_rfc4884.len == 0` therefore I cannot test the functionality. Any hint what could be the reason?

daniellerts: @pevik Please look at the sysctl: net.ipv4.icmp_errors_extension_mask.
Its default is 0x00, meaning the Linux kernel does not include ICMP extensions in its Time Exceeded replies.
Only after setting it to 0x01 the kernel starts attaching extension objects to outgoing ICMP errors.

daniellerts: @pevik also, i pushed with rebase (no conflicts or any change), and now CI failed on openSUSE, which i dont see how it is related to my code. Can you check or are you aware or can you help me understand please?
Thanks!

pevik: @daniellerts I fixed CI issue, can you please once more rebase?

daniellerts: @pevik done, thank you!

pevik: > @pevik Please look at the sysctl: net.ipv4.icmp_errors_extension_mask. Its default is 0x00, meaning the Linux kernel does not include ICMP extensions in its Time Exceeded replies. Only after setting it to 0x01 the kernel starts attaching extension objects to outgoing ICMP errors.

Hm, it still did not help, `len` is still 0, therefore nothing printed:

```
$ sysctl net.ipv4.icmp_errors_extension_mask
net.ipv4.icmp_errors_extension_mask = 0

# sysctl -w net.ipv4.icmp_errors_extension_mask=1
$ sysctl net.ipv4.icmp_errors_extension_mask
net.ipv4.icmp_errors_extension_mask = 1

$ ./builddir/tracepath -e -n 198.51.100.4
 1?: [LOCALHOST]                      pmtu 1500
 1:  10.0.1.1                                           9.532ms 
 1:  10.0.1.1                                           3.887ms 
...
```

pevik:
> @pevik Please look at the sysctl: net.ipv4.icmp_errors_extension_mask.
> Its default is 0x00, meaning the Linux kernel does not include ICMP extensions in its Time Exceeded replies.
> Only after setting it to 0x01 the kernel starts attaching extension objects to outgoing ICMP errors.

@metan-ucw FYI you need `sysctl -w net.ipv4.icmp_errors_extension_mask=1` if you want to reproduce the output.

tracepath.c
pevik: Hm, fixed on a VM with nearly the same kernel, wierd. Anyway, working, I'm sorry for the noise.

```
# ./builddir/tracepath -e -n 198.51.100.4
 1?: [LOCALHOST]                      pmtu 1500
 1:  10.0.1.1                                         0.332ms 
     <type=2: role=inc,ifindex=10,ip=10.0.1.1,name=virbr0,mtu=1500>
 1:  10.0.1.1                                         0.122ms 
     <type=2: role=inc,ifindex=10,ip=10.0.1.1,name=virbr0,mtu=1500>
 2:  no reply
...
```

tracepath.c
```diff
@@ -72,8 +74,13 @@ struct hhistory {
...
 struct probehdr {
-	uint32_t ttl;
-	struct timespec ts;
+	union {
+		struct {
+			uint32_t ttl;
+			struct timespec ts;
+		} probe_data;
+		char buf[MAX_ICMP_MSG_SIZE];
+	} un;
```
metan-ucw: Maybe we can make the structure and union anonymous here (should be supported in gcc since 2011) with that there would be no need to change the actual code below.

pevik: 15 years old gcc feature is old enough to be used.

tracepath.c
```diff
@@ -135,6 +160,147 @@ static void print_host(struct run_state const *const ctl, char const *const a,
...
+			if (offset + sizeof(struct in6_addr) > obj_len)
+				goto err;
+
+			inet_ntop(AF_INET6, pext + offset, ip6addr,
+				  INET6_ADDRSTRLEN);
+			printf(",ip=%s", ip6addr);
+			offset += sizeof(struct in6_addr);
+			break;
+			}
+#endif
```
metan-ucw: This code is ugly and as far as I can tell subtly broken. We do not advance the offset if the ICMP_AFI_IP* constants are not defined with breaks the parsing.

I think that it makes much more sense to use the icmp_ext_echo_iio structure instead. I guess that we just need to make sure the data are at least sizeof(struct icmp_ext_echo_iio) then we can cast the data to this structure and access the members directly rather than this tricky play with offsets.

daniellerts: icmp_ext_echo_iio.ident is a union. Its three members (name, ifindex, addr) all start at the same offset (4 bytes into the struct). This is correct for RFC 8335, where class_type is a type selector: only one identifier is ever present per object. However, RFC 5837 is fundamentally different: class_type is a bitmask, and multiple sub-objects can be present simultaneously.

I can fix the fact default label doesnt promote the offset, and instead of declaring variables and printing inside each case, just set af and addr_size per case, then do a single inet_ntop call after the switch. This eliminates the duplicated code and the nested {} blocks.

will it be better?

metan-ucw: There is actually a `struct icmp_ext_echo_iio` in `linux/icmp.h` and the ident is properly defined as an union there. I think that the cleanest option is to check that the data are long enough >= sizeof(struct icmp_ext_echo_iio) and then cast the pointer to the structure. After that we can just access the members instead of adding offsets.

daniellerts: `icmp_ext_echo_iio.ident` is a union precisely because RFC 8335 only ever has one identifier. RFC 5837 can have all four sub-objects simultaneously, so a union cannot represent it:
"A single ICMP message can contain as few as zero and as many as four instances of the Interface Information Object" (RFC 5837 section 4).

pevik: So you suggesting casting to `struct icmp_ext_echo_iio` would not work when there are more than one sub-object, right?

daniellerts: Yes.

metan-ucw: Then we are back at my original comment, if this is a stream of bytes where we need to advance based on the class_type the code is broken with the `#ifdef ICMP_AFI_IP` because if we compile the code on a system where these are not defined the parser would be broken. And the code that prints ",ip=unknown" should probably `goto err;` because the offset we should advance with is undefined at that point. Or we can use the `addrlen` byte you skip in the code. Oh and we need to assert that the addrlen matches sizeof(ipv4_addr) and the same for ipv6.

Also I would still use more structures in the code, i.e.:

```
#define OFFSET_CHECK(offset, len, add) do {\
             if (offset+add > len)
                      goto err;
             offset+=add;
} while (0)

if (objh->class_type & ICMP_EXT_CTYPE_IIO_IPADDR) {
       struct icmp_ext_echo_ctype3_hdr *ctype3_hdr = pext + offset;

       OFFSET_ADVANCE(offset, obj_len, sizeof(*ctype3_hdr));

       switch (ctype3_hdr.afi) {
       case ICMP_AFI_IP:
               ...
       break;
       case ICMP_AFI_IP6:
              ...
       break;
       default:
                  printf(",ip=unknown");
                  OFFSET_ADVANCE(offset, obj_len, ctype3_hdr->addrlen);
       ...
```

metan-ucw: Hmm, looking at the kernel code in `net/ipv4/icmp.c` and `net/ipv6/icmp.c` the ` struct icmp6_ext_iio_addr4_subobj` and `struct icmp6_ext_iio_addr6_subobj` have 2 reserved bytes, no size byte there, so I guess the best we can do is:

- add ICMP_AFI_IP and ICMP_AFI_IP6 fallback definitions so that they are always defined
- do goto err on unknown `afi`
- add our own `struct icmp_ext_iio_addr_headr {uint16_t afi, uint16_t reserved}` and use that for parser

pevik: Sounds reasonable, thanks!

> `struct icmp6_ext_iio_addr4_subobj`

`struct icmp_ext_iio_addr4_subobj`

daniellerts: I tried to complete all of your requests. Please let me know if that is ok now.
Thank you!

tracepath.c
```diff
@@ -102,9 +104,39 @@ struct run_state {
...
+#ifndef ICMP_AFI_IP6
+#define ICMP_AFI_IP6	2
+#endif
+
+#define ICMP_EXT_IIO_NAME_SIZE		63
+
+struct icmp_ext_iio_name_subobj {
+	uint8_t len;
+	char name[ICMP_EXT_IIO_NAME_SIZE];
+};
```
metan-ucw, line 135: Shouldn't this be:

```
struct icmp_ext_iio_name_subobj {
	uint8_t len;
	char name[];
};
```

Adding explicit size to something that is a flex array is confusing.

tracepath.c
```diff
@@ -136,6 +168,136 @@ static void print_host(struct run_state const *const ctl, char const *const a,
...
+
+	offset += sizeof(*objh);
+
+	if (objh->class_type & ICMP_EXT_CTYPE_IIO_IFINDEX) {
+		uint32_t ifindex;
+
+		if (offset + sizeof(ifindex) > obj_len)
+			goto err;
+
+		ifindex = ntohl(*((uint32_t *)(pext + offset)));
```
metan-ucw: Can we at least add some static inline helpers so that the code gets more readable:

```
static inline int get_u32(uint32_t *res, void *addr, unsigned int *off, unsigned int len)
{
        if (*off+4>len)
            return 1;
       *res = ntohl(*(uint32_t*)(addr + *off)));
       *off+=4;
       return 0;
}
```

Then we could simplify the code to:

```
        uint32_t ifindex;
        if (get_u32(&ifindex, pext, &offset, obj_len))
               goto err;
        printf(",ifindex=%u", ifindex);
```

And the same for u16.

metan-ucw (changes requested):
The code looks correct to me now, but there are two minor issues I would like to get addressed.

metan-ucw:
Looks good to me.

tracepath.c
pevik: I'm sorry for a delay. 

@metan-ucw Now I see @daniellerts effort for non-anonymous union. It's not about gcc version, but about the standard we chosen. We still use C99 (`-std=c99`):

```
../tracepath.c:84:18: warning: ISO C99 doesn?t support unnamed structs/unions [-Wpedantic]
   84 |                 };
      |                  ^
../tracepath.c:86:10: warning: ISO C99 doesn?t support unnamed structs/unions [-Wpedantic]
   86 |         };
      |          ^
../tracepath.c:79:8: warning: struct has no named members [-Wpedantic]
   79 | struct probehdr {
      |        ^~~~~~~~
```

We should probably accept C11 (`-std=c11`).


https://github.com/pevik/iputils/commit/bb8e429b30f091b4e1793fb29bfa4103913a96d8
https://github.com/iputils/iputils/compare/master...pevik:iputils:refs/heads/tracepath_icmp_ext/v1.rebased.c11

```diff
+++ meson.build
@@ -3,7 +3,7 @@
 
 project('iputils', 'c',
        default_options : [
-               'c_std=c99',
+               'c_std=c11',
                'warning_level=3',
                'localstatedir=var',
        ],
```

WDYT? If you agree, can you add your `Reviewed-by:` please?

pevik: FYI we could even have `gnu11`, but IMHO not needed.

tracepath.c
pevik: I should have noted which VM or at least kernel I was using for testing. Because now I'm not able to reproduce (trying on several VM and real machines with different kernels). @daniellerts Any hint what else besides `sysctl -w net.ipv4.icmp_errors_extension_mask=1`?

pevik: Hm, this is strange, it looks to me completely unrelated to `net.ipv4.icmp_errors_extension_mask` setup. Currently I disabled it (and later even reboot) so that the setup is `net.ipv4.icmp_errors_extension_mask = 0` and extensions are shown. I'm confused.

Testing on Tumbleweed with 7.2.0-rc6-3.g1b28f10-default.

pevik: Ah, I got it. I really need to run it in VM, because extensions I see are only in virtual network created by libvirt setup:
```
$ ip link show virbr0 
4: virbr0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc htb state UP mode DEFAULT group default qlen 1000
    link/ether XX:XX:XX:XX:XX:XX brd ff:ff:ff:ff:ff:ff
```
I need to figure out how to create this network + run a host in an easy way  in our testsuite to be able to test this feature.

pevik:
```
Tested-by: Petr Vorel <pvorel@suse.cz>
```
LGTM, thanks

pevik:
> > @pevik Please look at the sysctl: net.ipv4.icmp_errors_extension_mask.
> > Its default is 0x00, meaning the Linux kernel does not include ICMP extensions in its Time Exceeded replies.
> > Only after setting it to 0x01 the kernel starts attaching extension objects to outgoing ICMP errors.
> 
> @metan-ucw FYI you need `sysctl -w net.ipv4.icmp_errors_extension_mask=1` if you want to reproduce the output.

<s>Obviously `sysctl -w net.ipv4.icmp_errors_extension_mask=1` is not needed. Likely it's more about specific routing condition (e.g. net created by libvirt).</s>
OK, the router needs to have `sysctl -w net.ipv4.icmp_errors_extension_mask=1` enabled, not the host. I'll update that in the commit message before merge.

pevik:
@metan-ucw @daniellerts Can I use your ack (your `Reviewed-by`) also for this change which I plan to merge together with the change?
https://github.com/pevik/iputils/commit/bb8e429b30f091b4e1793fb29bfa4103913a96d8

PR #616, opened 2026-03-03
Fix fd leak

cuiweixie:
should close fd after not using it.


metan-ucw:
The function is called just before the main() exits, hence there is no leak there. It may silence some static analyzers though.

PR #620, opened 2026-05-18
fix olen miscalculation when reading IP options from cmsg

loaniloroze:


ping/ping.c
```diff
@@ -1683,7 +1683,7 @@ int ping4_parse_reply(struct ping_rts *rts, struct socket_st *sock,
 					continue;
 				memcpy(&reply_ttl, CMSG_DATA(cmsgh), sizeof(reply_ttl));
 			} else if (cmsgh->cmsg_type == IP_RETOPTS) {
-				opts = (uint8_t *)CMSG_DATA(cmsgh);
+				opts = (uint8_t *)CMSG_DATA(cmsgh) - CMSG_LEN(0);
 				olen = cmsgh->cmsg_len;
```
metan-ucw, line 1687: This does not make any sense, it may have make sense to do `olen = cmsg->cmsg_len - CMSG_LEN(0)` since we are probably interesting only in the payload lenght. But subscribing the length from the `CMSG_DATA(cmsgh)` is wrong since the `CMSG_DATA` already advances the pointer after the header.

PR #621, opened 2026-05-24
ping: provide correct time in final statistics

dasha-uwu:
Before:
```
$ ping -w 1 127.0.0.1
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.060 ms

rtt min/avg/max/mdev = 0.060/0.060/0.060/0.000 ms
```
After:
```
$ ping -w 1 127.0.0.1
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.022 ms

1 packets transmitted, 1 received, 0% packet loss, time 1000ms
rtt min/avg/max/mdev = 0.022/0.022/0.022/0.000 ms
```


metan-ucw:
A bit better description wouldn't harm, looks like the `rts->cur_time` is the last time we send a probe, so this all depends on how the overall time is defined. If it's defined as a period in which we were sending probes, the original code is correct. If the overall time is the time the ping program was running this change is correct. @pevik you decide.

pevik:
> A bit better description wouldn't harm,

+1. @dasha-uwu describing the goal or a motivation helps to understand your intention :).

> looks like the `rts->cur_time` is the last time we send a probe, so this all depends on how the overall time is defined. If it's defined as a period in which we were sending probes, the original code is correct. If the overall time is the time the ping program was running this change is correct. @pevik you decide.

TL;DR I'm not sure myself. Due  backwards compatibility I tend to keep the current state unless there is a good reason to change (more usable to change the behavior).

Unfortunately the man page does not describe it (we should document it). I suppose it was meant as it is now i.e.  _period in which we were sending probes_  (i.e. no bug), because it has been from the start 33370345c7d8c217b51c13b0e2864b64b53d5f96 (in 2006). I searched the history. It was added in http://ftp.icm.edu.pl/packages/linux-iproute/ip-routing/iputils-ss010805.tar.gz, but RELNOTES file in the tarball does not mention this change. Also busybox ping uses this implementation. Inetutils ping does not print time at all.

I suppose the current functionality _how long were sending probes_ useful for somebody. Although probably people expect ti might be _how long the ping program was running_ one can get this value by using `time` shell buildin/tool.

PR #630, opened 2026-09-11
build: make `./configure -Dprefix=…` work

jengelh:
Traditionally, `./configure` scripts were almost always scripts generated by GNU autotools. Even if `configure` these days is not necessarily linked to some form of autotools execution, the user expectation is, I would argue, that all arguments are somehow processed/passed on. This is currently not the case for iputils.
    
With this change, `./configure -Dprefix=/whatever` becomes usable (mimicing the `./configure --prefix=…` one would have used with autoconf).


pevik:
Thanks! Also fixing something like this was my plan. Hopefully I'll find a time to review soon.

pevik:
> Thanks! Also fixing something like this was my plan. Hopefully I'll find a time to review soon.

Well, I meant something slightly different: posted in https://github.com/iputils/iputils/pull/631#issuecomment-5648394257 but that comment is actually more relevant in this ticket, thus repeat: What I'd prefer is to reconfigure project when I pass whatever option to the script and have the output we have at the end. That is the goal I wanted.

PR #631, opened 2026-09-11
build: make `./configure --help` work

jengelh:
Traditionally, `./configure` scripts were almost always scripts generated by GNU autotools. Even if `configure` these days is not necessarily linked to some form of autotools execution, the user expectation is, I would argue, that a singular `--help` option should be understood, if only as a user courtesy. This is currently not the case for iputils, so add it.


yvs2014:
> Traditionally, ./configure scripts were almost always scripts generated by GNU autotools.

Yeah, it's a bit confusing, maybe it needs to rename.

p.s. to get good looking output there's a built-in way to display options, just
`meson configure`
without any other arguments/options

pevik:
> > Traditionally, ./configure scripts were almost always scripts generated by GNU autotools.
> 
> Yeah, it's a bit confusing, maybe it needs to rename.

I'm not sure what the script name should be. I would be ok to keep it as is. Or, maybe we should look at mesa tools what they use.

> p.s. to get good looking output there's a built-in way to display options, just `meson configure` without any other arguments/options

To be honest I consider `meson configure` too verbose. What I'd prefer is to reconfigure project when I pass whatever option to the script and have the output we have at the end. That is the goal I wanted.

yvs2014:
> I'm not sure what the script name should be. I would be ok to keep it as is.
>
  With 'configure' name and having nothing in common with configure* from autotools, IDK but it certainly confuses

PR #633, opened 2026-09-12
Make xsltproc optional

jengelh:
xsltproc is an optional dependency, but it still requires an explicit opt-out switch.

Take a more relaxed approach and make the use of xsltproc opportunistic rather than opt-out/in, modeled after how many existing projects' configure scripts behave.


pevik:
Thank you, I wanted to implement this myself, but haven't find a time. I'll try to review soon.
Could you please fix meson error:
```
meson.build:4:0: ERROR: Value "true" (of type "string") for option "BUILD_HTML_MANS" is not one of the choices. Possible choices are (as string): "enabled", "disabled", "auto".
```

